PR #10, opened 2020-01-13
Add example of mixed-source app and fix module name bug

ob:
Add a [mixed source example](https://github.com/ob/bazel-tests/tree/master/mixed) and get it closer to compiling.

Without this PR, I get: 

```
> bazel build //examples/...
INFO: Analyzed 15 targets (2 packages loaded, 53 targets configured).
INFO: Found 15 targets...
ERROR: /Users/obonilla/o/rules_ios/examples/mixed-test/BUILD.bazel:3:1: Compiling Swift module mixed-test failed (Exit 1)
<unknown>:0: error: module name "mixed-test" is not a valid identifier
INFO: Elapsed time: 0.180s, Critical Path: 0.05s
INFO: 4 processes: 4 darwin-sandbox.
```

After normalizing the module name, I get:

```
> bazel build //examples/...
DEBUG: /Users/obonilla/o/rules_ios/rules/library.bzl:403:9: module name = mixed_test
INFO: Analyzed 15 targets (2 packages loaded, 53 targets configured).
INFO: Found 15 targets...
ERROR: /Users/obonilla/o/rules_ios/examples/mixed-test/BUILD.bazel:3:1: C++ compilation of rule '//examples/mixed-test:mixed-test_objc' failed (Exit 1) wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang -arch x86_64 '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG -include ... (remaining 50 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
examples/mixed-test/ObjcFile.m:12:9: fatal error: 'mixed_test-Swift.h' file not found
#import "mixed_test-Swift.h"
        ^~~~~~~~~~~~~~~~~~~~
1 error generated.
INFO: Elapsed time: 4.883s, Critical Path: 4.76s
INFO: 9 processes: 6 darwin-sandbox, 3 worker.
```

This is because the name of the module is actually `mixed_test_swift-Swift.h`. Unfortunately,  `rules_swift`'s `swift_library` rule doesn't allow changing the name of the header generated (see [this PR](https://github.com/bazelbuild/rules_swift/pull/350) for a possible solution there).

Putting this up here while we decide whether to push on that `swift_library` PR and bump our dependency (preferred) or work around it.


thii:
As a workaround, you should be able to use `@import mixed_test.Swift;`.

rules/library.bzl
```diff
@@ -243,7 +246,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         else:
             fail("Unable to compile %s in apple_framework %s" % (f, name))
 
-    module_name = kwargs.pop("module_name", name)
+    module_name = _normalize_module_name(kwargs.pop("module_name", name))
```
segiddins: I think we should only normalize in the fallback case, i.e.

`kwargs.pop("module_name", _normalize_module_name(name))`

examples/mixed-test/SwiftObj.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  mixed_test.swift
+//  mixed-test
+//
+//  Created by Oscar Bonilla on 7/23/18.
+//  Copyright © 2018 Oscar Bonilla. All rights reserved.
+//
+
+public class SwiftObj : NSObject {
```
segiddins: needs an `import Foundation`

ob: Shouldn't that come from the Objective-C headers?

rules/library.bzl
ob: Hmm... good point. I'll change it.

examples/mixed-test/SwiftObj.swift
segiddins: I don't think Swift will pull in transitive modules like that?

PR #16, opened 2020-02-07
Allow customizing which resource extensions are filtered

segiddins:
This way, our macros can filter out other directories that are causing clashes at the moment (`.lproj` dirs, I'm looking at you...)

### DIFF
--- a/rules/library/resources.bzl
+++ b/rules/library/resources.bzl
@@ -1,11 +1,11 @@
-def _is_apple_resource_file(file):
-    if file.extension in ("xcdatamodeld", "xcmappingmodel"):
+def _is_apple_resource_file(file, extensions_to_filter):
+    if file.extension in extensions_to_filter:
         return False
 
     return True
 
 def _resources_filegroup_impl(ctx):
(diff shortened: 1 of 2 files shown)


rules/library/resources.bzl
```diff
@@ -16,16 +16,22 @@ resources_filegroup = rule(
...
     },
     doc = """Wraps a set of srcs for use as `data` in an `objc_library` or `swift_library`,
 or `resources` in an `apple_resource_bundle`.
 """,
 )
 
-def wrap_resources_in_filegroup(name, srcs, **kwargs):
+def wrap_resources_in_filegroup(name, srcs, extensions_to_filter = [], **kwargs):
+    extensions_to_filter = list(extensions_to_filter)
+    for x in ("xcdatamodeld", "xcmappingmodel"):
```
justinseanmartin: Its ok to not provide a way to opt-out of these two being appended, right? Seems fine to me, but just checking.

segiddins: They need to be filtered, otherwise there’s a guaranteed failure in the Apple bundling rules

PR #17, opened 2020-02-13
Add iOS 13 test hosts

segiddins:
This way, `test_host = True` on an `ios_unit_test(minimum_os_version = "13.0")` will work


rules/test_host_app/BUILD.bazel
```diff
@@ -23,5 +23,7 @@ exports_files([
         "10.0",
         "11.0",
         "12.0",
+        "13.0",
```
justinseanmartin: Is this something that we should be able to define externally so we don't have to make library bumps each time a new xcode comes out?

segiddins: There's no way to do that

PR #18, opened 2020-02-14
Allow headers in test bundles /apps to be imported the same as in Xcode

segiddins:
With these changes, we can successfully run almost all of our apps & tests!

- Add a test for a test target importing headers from its test host

- Tag internally expanded rules as manual

- Don’t try to gather direct_modulemaps from a swift target
    
    apple_common.Objc.direct_module_maps is broken coming from swift_library
    (it contains one level of transitive module maps), so ignore SwiftInfo from swift_library,
    since it doesn't have a module_map field anyway

- Stop generating modulemaps for tests & apps
    
    This matches CocoaPods’ behavior
    
    - The namespace for header names is the Test/Framework name, rather than the module name
    - Test/Framework don’t use -import-underlying-modulemap, allowing bridging headers to be used

- Specify custom swift_librart generated_header_name

- Update to Bazel 2.1 and HEAD of apple rules

- Infer LaunchScreen.storyboard from resources filegroup

- Don’t default packaging rule attrs to none unecessarily
    
    We can have them not even be in the kwargs dict


segiddins:
@ob @chenxiao0228 btw this both adds a new test (:yay:) and uses `generates_header_name` so imports of `-Swift.h` headers will work!

PR #23, opened 2020-03-12
Add support for deriving copts from apple_library xcconfig param

segiddins:
This will allow `cocoapods-bazel` to export `pod_target_xcconfigs` straight to `BUILD` files, and the macros will translate those xcconfigs into the appropriate `copts`, which the library then consumes.

I also figured out a (very hacky) way to be able to test that behavior, so there are tests now that the xcspec data is interpreted correctly.

The way this works is:

1. there's a python script that dumps out a bunch of xcspecs into a bzl file
2. there's a starlark macro that takes in a dict of xcconfigs
3. for each xcconfig, loop through the xcspecs and see if it has a matching option
4. if so, interpret the option, extracting command line args as copts. Error if it's a case we haven't handled yet.
5. Accumulate those copts, return them in a struct to the library macro
6. Append those copts to the user-supplied copts


tests/xcconfig/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+# apple_framework(
+#     name = "invalid_boolean_key_xcconfig",
+#     srcs = SRCS,
+#     xcconfig = {
+#         "GCC_TREAT_WARNINGS_AS_ERRORS": "FILE_NOT_FOUND"
+#     },
+# )
+
+genquery(
+    name = "boolean_key_xcconfig_libraries",
```
amberdixon: How is the output file generated by this genquery being used?

tests/xcconfig/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+
+EXPECTED_FAILURE = ["manual"]
+
+apple_framework(
+    name = "empty_xcconfig",
+    srcs = SRCS,
+    xcconfig = {},
+)
+
+apple_framework(
```
amberdixon: Are you just invoking these rule implementations from these tests to ensure that the rule implementation doesn't break when these different xcconfig values are passed in?

data_generators/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
...
+    srcs = ["xcspec_extractor.py"],
+)
+
+genrule(
+    name = "extract_xcspecs",
+    outs = ["extract_xcspecs.sh"],
+    cmd = """set -eu
+cat <<EOS > $(OUTS)
+set -eu
+readonly output=\\$$($(location @bazel_tools//tools/osx:xcode-locator) 2>/dev/null | grep : | sed -Ee s/.*'"(.+)": "(.+)".*'/'\\1.0|\\2'/g | sort --version-sort -k 1 -t '|' | tail -n 1)
```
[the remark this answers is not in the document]
segiddins: If it fails, the output will be printed. No need to check the status code, but probably want to add a -o pipefail.

docs/BUILD.bazel
```diff
@@ -22,6 +22,19 @@ bzl_library(
         "@build_bazel_apple_support//lib:apple_support.bzl",
         "@build_bazel_rules_apple//apple:ios.bzl",
         "@build_bazel_rules_swift//swift:swift.bzl",
+        "@rules_cc//cc:defs.bzl",
```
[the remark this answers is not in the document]
segiddins: It’s imported somewhere along the line

tests/xcconfig/BUILD.bazel
segiddins: It’s not, I can remove it, but it’s a way of seeing what copts are being passed through to the underlying rules

tests/xcconfig/BUILD.bazel
segiddins: Correct

data_generators/BUILD.bazel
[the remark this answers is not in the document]
justinseanmartin: Maybe worth a comment here of what output you're expecting from `xcode-locator` to make it easier to follow the rest of the logic here.

data_generators/xcspec_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+ids = (
+    "com.apple.compilers.llvm.clang.1_0",
+    "com.apple.compilers.model.coredata",
+    "com.apple.compilers.model.coredatamapping",
+    "com.apple.pbx.linkers.ld",
+    "com.apple.xcode.tools.ibtool.compiler",
+    "com.apple.xcode.tools.swift.compiler",
+)
+
+print("# Extracted from Xcode %s" % xcode_version)
```
justinseanmartin: IIUC - `data_generators` is a manual task to run to update the xcspec, correct? Worth adding a comment to the top of the file produced with how to go about regenerating it?

tests/xcconfig/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+    name = "boolean_key_xcconfig",
+    srcs = SRCS,
+    xcconfig = {
+        "GCC_TREAT_WARNINGS_AS_ERRORS": "YES",
+    },
+)
+
+# apple_framework(
```
justinseanmartin: Assume this is a case of "uncomment to make sure that this fails as expected"? Might be worth having a comment explaining the commented code.

data_generators/BUILD.bazel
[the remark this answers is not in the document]
ob: Agree with `-o pipefail` and also maybe it'd be worth writing the `cmd` as an actual shell script since it's not trivial?

data_generators/xcspec_extractor.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
ob: What if python3 is not installed? I think we can only count on python 2 by default no?

data_generators/xcspec_extractor.py
```diff
@@ -0,0 +1,67 @@
...
+
```
ob: A comment saying what the program does and simple usage info would be nice

rules/library/xcconfig.bzl
```diff
@@ -0,0 +1,159 @@
...
+_MOMC = "com.apple.compilers.model.coredata"
+_MAPC = "com.apple.compilers.model.coredatamapping"
+_IBTOOL = "com.apple.xcode.tools.ibtool.compiler"
+_OTHERWISE = "<<otherwise>>"
+
+def _unknown_enum_value(option, value, fatal = False):
+    if option["Type"] != "Enumeration":
+        return
+
+    printer = fail if fatal else print
```
ob: Cute

data_generators/xcspec_extractor.py
segiddins: Given this is only run manually, I'm not too too concerned -- we check in it's artifact, and the rules don't use the generator directly

ob: Sure, but it doesn't look like you're using anything from python 3 in particular... shouldn't be too hard to keep it python2/python3 compatible? I agree that i's not a biggie though...

data_generators/xcspec_extractor.py
justinseanmartin: Bump

tests/xcconfig/BUILD.bazel
justinseanmartin: Bump

tests/xcconfig/tests.bzl
```diff
@@ -0,0 +1,177 @@
...
+
+    r(
+        name = name,
+        target_under_test = name + "_target_under_test",
+        expected = expected,
+        error = error,
+    )
+    return name
+
+def xcconfig_unit_test_suite():
```
justinseanmartin: Nice test suite!

tests/xcconfig/BUILD.bazel
segiddins: This diff is outdated, commented-out code has been removed

PR #24, opened 2020-03-13
Add a clean action to framework rule

segiddins:
This ensures stale files will get removed when running in non-sandboxed mode, fixing cases where stale files (mostly headers) could incorrectly break compilation


tests/app/ios/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("//rules:framework.bzl", "apple_framework")
+load("//rules:app.bzl", "ios_application")
+
+apple_framework(
+    name = "FW",
+    srcs = glob(["FW/*"]),
+    private_headers = glob(["FW/*_Private.*"]),
+)
+
+ios_application(
```
justinseanmartin: This is just verifying that we can build frameworks generally, but not necessarily that the clean action is working, correct? Just want to make sure I'm understanding correctly the intent.

segiddins: Correct -- I needed a testbed to make changes in, and there wasn't one with apps and frameworks, so I added it

rules/framework.bzl
```diff
@@ -166,20 +168,43 @@ def _apple_framework_packaging_impl(ctx):
...
+            framework_root,
+            "--inputs",
+            ctx.actions.args().use_param_file("%s", use_always = True).set_param_file_format("multiline")
+                .add_all(framework_files),
+            "--outputs",
+            framework_manifest.path,
+        ],
+        outputs = [framework_manifest],
+        mnemonic = "CleaningFramework",
+        execution_requirements = {
```
ob: Is there a way to only run this action in incremental mode?

rules/framework/framework_packaging.py
```diff
@@ -80,6 +80,40 @@ def _copy_modulemap(framework_root, modulemap_path):
...
+def _clean(framework_root, manifest_file, output_manifest_file):
+    """Remove stale files from the framework root.
+
+    Args:
+        framework_root: root folder of the framework
+        manifest_file: path to the input manifest file
+        output_manifest_file: path to the output manifest file
+    """
+
+    with open(manifest_file) as f:
```
ob: If this runs in full builds as well as incremental, I'd implement a _fail early_ here where if the `framework_root` doesn't exist this function just returns without having to read the manifest file.

rules/framework/framework_packaging.py
```diff
@@ -80,6 +80,40 @@ def _copy_modulemap(framework_root, modulemap_path):
...
+        dirs_to_keep = set((framework_root,))
+        for file in manifest_lines:
+            files_to_keep.add(file)
+            
+            dir = os.path.dirname(file)
+            while dir not in dirs_to_keep:
+                dirs_to_keep.add(dir)
+                dir = os.path.dirname(dir)
+
+    for root, dirs, files in os.walk(framework_root):
```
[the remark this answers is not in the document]
segiddins: purposefully not, since we're leaving some files behind

rules/framework.bzl
segiddins: Not that I could find :(

rules/framework.bzl
```diff
@@ -166,20 +168,43 @@ def _apple_framework_packaging_impl(ctx):
...
-    private_header_out = _framework_packaging(ctx, "private_header", private_header_in, private_header_out)
-    modulemap_out = _framework_packaging(ctx, "modulemap", [modulemap_in], modulemap_out)
-    swiftmodule_out = _framework_packaging(ctx, "swiftmodule", [swiftmodule_in], swiftmodule_out)
-    swiftdoc_out = _framework_packaging(ctx, "swiftdoc", [swiftdoc_in], swiftdoc_out)
+    binary_out = _framework_packaging(ctx, "binary", binary_in, binary_out, framework_manifest)
+    header_out = _framework_packaging(ctx, "header", header_in, header_out, framework_manifest)
+    private_header_out = _framework_packaging(ctx, "private_header", private_header_in, private_header_out, framework_manifest)
+    modulemap_out = _framework_packaging(ctx, "modulemap", [modulemap_in], modulemap_out, framework_manifest)
+    swiftmodule_out = _framework_packaging(ctx, "swiftmodule", [swiftmodule_in], swiftmodule_out, framework_manifest)
+    swiftdoc_out = _framework_packaging(ctx, "swiftdoc", [swiftdoc_in], swiftdoc_out, framework_manifest)
```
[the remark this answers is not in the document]
segiddins: No, but it needs to be an input to force the clean action to run _before_ the packaging actions -- otherwise, it would never actually get run

rules/framework/framework_packaging.py
```diff
@@ -80,6 +80,40 @@ def _copy_modulemap(framework_root, modulemap_path):
...
+        for d in dirs:
+            path = os.path.join(root, d)
+            if path not in dirs_to_keep:
+                os.rmdir(path)
+        for f in files:
+            path = os.path.join(root, f)
+            if path not in files_to_keep:
+                os.remove(path)
+
+    _cp(manifest_file, output_manifest_file)
```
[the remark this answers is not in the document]
segiddins: This allows the `clean` action to get cached (along with the fact that actions in bazel always need to have at least one output). Also, without having an output, it wouldn't be possible to make this action a pre-requisite to the other actions

rules/framework.bzl
```diff
@@ -166,20 +168,43 @@ def _apple_framework_packaging_impl(ctx):
...
     framework_files = _concat(binary_out, modulemap_out, header_out, private_header_out, swiftmodule_out, swiftdoc_out)
-
     framework_root = _find_framework_dir(framework_files)
 
+    ctx.actions.run(
+        executable = ctx.executable._framework_packaging,
+        arguments = [
```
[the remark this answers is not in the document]
segiddins: No -- actions won't be run unless all of their inputs exist

rules/framework/framework_packaging.py
ob: How about adding? 
```suggestion
    if not os.path.exists(framework_root):
        return
    with open(manifest_file) as f:
```?

PR #28, opened 2020-03-24
Split out tests into multiple different github actions

segiddins:
This should speed up PRs by parallelizing, as well as making it clearer which bit is failing


.github/workflows/tests.yml
```diff
@@ -14,16 +14,41 @@ jobs:
...
         if: failure()
         with:
           name: bazel-testlogs
           path: bazel-testlogs
+  check_docs:
+    name: Check Starlark
+    runs-on: macOS-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Select Xcode 11.2
```
amberdixon: small nit: is it necessary to select xcode versions before running buildifier?

.bazelrc
```diff
@@ -1,3 +1,3 @@
 # We can't create a bzl_library for rules-swift because of its visibility,
 # so circumvent by not using the sandbox
-build --strategy=Stardoc=standalone
+common --strategy=Stardoc=standalone
```
amberdixon: I see the strategy option is documented in the bazel manuel and includes a valid value of standalone. But why is Stardoc also included here?

amberdixon: I see the standalone value listed in the bazel manual for the strategy option. But why is Stardoc also mentioned here?

.github/workflows/tests.yml
segiddins: No clue

.bazelrc
segiddins: because we only want to run actions with `Stardoc` as the mnemonic in standalone mode (we don't want to run other actions in standalone mode)

PR #31, opened 2020-03-31
fix bug: swift and OC mix framework

mengxiangyue:
try to fix this issue https://github.com/ob/rules_ios/issues/22, but I am not familiar with bazel, plz review it seriously. And test passed with demo https://github.com/vincephamgrab/demo-mixedobjcswift .

### DIFF
--- a/rules/hmap.bzl
+++ b/rules/hmap.bzl
@@ -1,6 +1,6 @@
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo")
 
-def _make_headermap_input_file(namespace, hdrs, flatten_headers):
+def _make_headermap_input_file(namespace, hdrs, flatten_headers, bin_dir_path):
     """Create a string representing the mappings from headers to their
     namespaced include versions. The format is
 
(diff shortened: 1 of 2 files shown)


rules/library.bzl
```diff
@@ -370,6 +370,27 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     )
     internal_deps.append(public_hmap_name)
 
+    if swift_sources:
+        generated_swift_header_name = module_name + "-Swift.h"
```
madnik, line 374: The generated header doesn't guarantee to have the suffix "-Swift.h". This can be changed.  I think we should fetch this from providers of `swift_library` rule.

rules/hmap.bzl
```diff
@@ -22,10 +22,13 @@ def _make_headermap_input_file(namespace, hdrs, flatten_headers):
     """
     entries = []
     for hdr in hdrs:
+        path = hdr.path
+        if path.endswith("-Swift.h"):
```
madnik, line 26: Here again. We should not assume always generated header has suffix -`-Swift.h`.

rules/library.bzl
segiddins: This is us setting what the header name will be

segiddins:
Can you add a test app/framework that covers this fix? Thanks!

rules/library.bzl
jschear: It looks like another variable is created for this header name further down in the file -- would it be better to reuse the variable declared here?

https://github.com/ob/rules_ios/blob/a690f50025d9c6014f36a309bba669859dc2357f/rules/library.bzl#L500

segiddins: yeah i think we can re-use

mengxiangyue: I find many places also use `"-Swift.h"` directly.

ob:
This PR is superseded by #38 right? Can we close it?

ob:
I think this got fixed in #38. Please reopen if that is not the case.

PR #36, opened 2020-04-10
Add resources to ios_unit_test kwargs

wileykestner:
- Pass the `resources` argument from the `rules_ios` `ios_unit_test` macro into the `ios_unit_test` rule vended by `build_bazel/rules_apple`

- Ensure the `resources` argument from the `rules_ios` `ios_unit_test` macro does not get passed along to `swift_library` and `objc_library`


segiddins:
(Thanks for adding a test for this!!)

PR #37, opened 2020-04-11
Refactor Xcodeproject generation code. Add in xcode scheme test action env vars and command line args.

amberdixon:


rules/xcodeproj.bzl
```diff
@@ -1,7 +1,7 @@
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 
-def _get_transitive(deps, provider, field):
+def _get_attr_values_for_name(deps, provider, field):
```
amberdixon, line 4: WDYT of this method rename? I thought this name was easier to understand, but open to hearing other options :)

rules/xcodeproj.bzl
```diff
@@ -81,11 +91,12 @@ def _xcodeproj_aspect_impl(target, ctx):
             actual = getattr(ctx.rule.attr, "tests")[0]
         elif ctx.rule.kind in ("alias"):
```
amberdixon, line 92: On line 92, how does the `in` keyword work. The line is`ctx.rule.kind in ("alias"):`

Is `("alias")` treated as a list with one item?

tests/xcodeproj/test.swift
```diff
@@ -2,11 +2,7 @@ import XCTest
 
 class SwiftTests : XCTestCase {
   func testPasses() {
-      let envvarvalue = ProcessInfo.processInfo.environment["test_envvar_key"]
```
segiddins: why did you get rid of the tests for the args / env var?

tests/xcodeproj/build.sh
```diff
@@ -2,6 +2,4 @@ set -eux
 
 cd $(dirname $0)
 
-xcodebuild -project Single-Application-Project.xcodeproj -quiet
-export SIM_DEVICE_ID=`xcodebuild -project Single-Application-Project.xcodeproj -scheme "Single-Application-UnitTests" -showdestinations | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last"`
-xcodebuild -project Single-Application-Project.xcodeproj -scheme "Single-Application-UnitTests"  test-without-building -destination "id=$SIM_DEVICE_ID"
```
segiddins: why not run the test?

tests/xcodeproj/build.sh
```diff
@@ -2,4 +2,6 @@ set -eux
 
 cd $(dirname $0)
 
-xcodebuild -project Single-Application.xcodeproj -quiet
+xcodebuild -project Single-Application-Project.xcodeproj -quiet
+export SIM_DEVICE_ID=`xcodebuild -project Single-Application-Project.xcodeproj -scheme "Single-Application-UnitTests" -showdestinations | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last"`
```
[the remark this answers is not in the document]
amberdixon: I will put back this change in later when I need it, right now it is not necessary run these tests. I think it will be necessary when the tests ensure that runs can run, when they are supposed to run inside an apphost.

tests/xcodeproj/test.swift
```diff
@@ -0,0 +1,12 @@
+import XCTest
+
+class SwiftTests : XCTestCase {
```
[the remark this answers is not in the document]
amberdixon: Yeah, it looks like it invokes a test runner and complains the argument is invalid. So I had changed the argument to `-h`, which is the help option for the test runner :D

tests/xcodeproj/build.sh
amberdixon: Since we validate that the scheme in the xcodeproject is associated with the correct set of env vars and cli args (by running git diff on the project.pbxproj file), we already ensure that bazel has correctly associated schemes with the right set of env and args.

tests/xcodeproj/test.swift
amberdixon: See comment below. The functionality being added in this PR can be tested by doing `git diff` on the xcodeproject file and ensuring it hasn't changed.

rules/xcodeproj.bzl
```diff
@@ -96,105 +107,119 @@ _xcodeproj_aspect = aspect(
...
+    )
+
+    ctx.actions.write(xcodegen_jsonfile, xcodeproj_info.to_json())
 
     ctx.actions.run(
         executable = ctx.executable._xcodegen,
-        arguments = ["--quiet", "--no-env", "--spec", xcodegen_yaml.path, "--project", project.dirname],
-        inputs = depset([xcodegen_yaml], transitive = [target.srcs for target in targets]),
+        arguments = ["--quiet", "--no-env", "--spec", xcodegen_jsonfile.path, "--project", project.dirname],
+        inputs = depset([xcodegen_jsonfile], transitive = [target.srcs for target in targets]),
```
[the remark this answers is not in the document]
amberdixon: According to this documentation, `validate` is not a field, but `optional` is. I have changed this to say `optional: True`. 
https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#sources

tests/xcodeproj/test.swift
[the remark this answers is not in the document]
segiddins: ha! it's possible command line args aren't yet possible to pass to https://github.com/google/xctestrunner, which is the underlying test runner the apple rules hook up

PR #40, opened 2020-04-15
Actually prepend copts so they get passed to tools

segiddins:
Due to a prior refactor, `_prepend_copts` didn't actually prepend to
its parameters, but rather re-assigned.

This PR ensures that the list that gets passed through to the underlying
rules is modified.

This change is tested by adding a library that cannot compile without
the `copts` from its xcconfig applied. ObjC files using manual reference
counting cannot be compiled without the `-fobjc-weak` flag, which in
this case is only applied by the `CLANG_ENABLE_OBJC_WEAK` xcconfig.

### DIFF
--- a/tests/xcconfig/BUILD.bazel
+++ b/tests/xcconfig/BUILD.bazel
@@ -39,4 +39,12 @@ apple_framework(
     },
 )
 
+apple_framework(
+    name = "wont_build_without_xcconfig",
+    non_arc_srcs = ["weak_property.m"],
+    xcconfig = {
+        "CLANG_ENABLE_OBJC_WEAK": "YES",
+    },
+)
+
 xcconfig_unit_test_suite()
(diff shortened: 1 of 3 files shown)


tests/xcconfig/BUILD.bazel
```diff
@@ -39,4 +39,12 @@ apple_framework(
     },
 )
 
+apple_framework(
+    name = "wont_build_without_xcconfig",
+    non_arc_srcs = ["weak_property.m"],
```
justinseanmartin, line 44: Do we know for sure that this option is working properly? Could potentially explicitly add a call to `addRef` and `release` so that compiling the code double checks that.

segiddins: We do, note that this target fails to compile without the xcconfig applied

PR #41, opened 2020-04-15
[WIP] try setting test host app by updating installer

amberdixon:
TODO: set bundle loader to point to derived data test host app


segiddins:
@amberdixon is this something you're still working on?

PR #42, opened 2020-04-15
Set test host for test target by pointing scheme to test host application, stub test host app's Info.plist

amberdixon:
This approach mostly works, but incremental xcodebuild and testruns seem to fail if the DerivedData Build/Intermediates directly is not removed first.


amberdixon:
> This approach mostly works, but incremental xcodebuild and testruns seem to fail if the DerivedData Build/Intermediates directly is not removed first.

This problem has been resolved. The issue was that the new build system leaves a folder in the DerivedData directory which (for some reason) causes incremental build and test attempts during local development to fail with an error from Xcode at launch time which says "Cannot attach to pid." This error only happen in the Xcode IDE, not when the proejct is tested from the `xcodebuild` command.

rules/xcodeproj.bzl
```diff
@@ -61,11 +68,15 @@ def _xcodeproj_aspect_impl(target, ctx):
                 direct_srcs = [],
             ),
         )
-        target_info = _TargetInfo(direct_target = info, targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
+        direct_targets = [info]
+        if test_host_target:
+            direct_targets.extend(test_host_target[_TargetInfo].direct_targets)
+        target_info = _TargetInfo(direct_targets = direct_targets, targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
```
amberdixon, line 74: Even if the developer does not want to load transitive targets, we still have to load the test host application as a target. Otherwise, Xcode will complain (since the xcode test target is pointing to a test host application.)

rules/xcodeproj.bzl
```diff
@@ -265,6 +291,8 @@ $BAZEL_INSTALLER
             "$(clang_stub_short_path)": ctx.executable.clang_stub.short_path,
             "$(clang_stub_ld_path)": ctx.executable.ld_stub.short_path,
             "$(clang_stub_swiftc_path)": ctx.executable.swiftc_stub.short_path,
+            "$(infoplist_stub)": ctx.file._infoplist_stub.short_path,
+            "$(workspacesettings_xcsettings_short_path)": ctx.file._workspace_xcsettings.short_path,
```
amberdixon, line 295: It is necessary to provide an Info.plist file which sets the bundle ID to $PRODUCT_BUNDLE_IDENTIFIER. Otherwise, if you try to run a test target which refers to a test host app and the test host app doesn't have an Info.plist, Xcode will complain that the Product bundle ID could not be determined.

tests/xcodeproj/BUILD.bazel
```diff
@@ -71,3 +71,13 @@ xcodeproj(
...
     ],
 )
+
+xcodeproj(
+    name = "Test-Target-With-Test-Host-Project",
+    testonly = True,
+    bazel_path = "bazelisk",
+    include_transitive_targets = False,
+    deps = [
+        "//tests/test/ios/test-imports-app:TestImports-Unit-Tests",
```
amberdixon, line 81: Note that the TestImports-Unit-Tests bazel target is associated with a test host. This test target also references a symbol (named `EmptyClass`) in a framework that is linked with the test host application.

If the generated xcodeproject does *not* associated the xcode test target with the test host application, then the test run will fail at runtime with the error: ```2020-05-17 22:50:25.228654-0700 xctest[20992:5296104] The bundle “TestImports-Unit-Tests” couldn’t be loaded. Try reinstalling the bundle.
2020-05-17 22:50:25.228838-0700 xctest[20992:5296104] (dlopen(/tmp/dd/Test-Target-With-Test-Host-Project-clhcmbpljtiftpecapezszzjrnmi/Build/Products/Debug-iphonesimulator/TestImports-Unit-Tests.xctest/TestImports-Unit-Tests, 265): Symbol not found: _OBJC_CLASS_$__TtC15TestImports_App10EmptyClass
  Referenced from: /tmp/dd/Test-Target-With-Test-Host-Project-clhcmbpljtiftpecapezszzjrnmi/Build/Products/Debug-iphonesimulator/TestImports-Unit-Tests.xctest/TestImports-Unit-Tests
  Expected in: /Users/amberdixon/Downloads/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/Library/Xcode/Agents/xctest
 in /tmp/dd/Test-Target-With-Test-Host-Project-clhcmbpljtiftpecapezszzjrnmi/Build/Products/Debug-iphonesimulator/TestImports-Unit-Tests.xctest/TestImports-Unit-Tests)
Program ended with exit code: 83
```

**Comment by amberdixon** [comment]
I considered an alternative approach which I discussed with @segiddins . This approach would be to:
1. Not set the test host application in the xcode test target.
2. Place the bazel-built test host app and xctest bundles in the correct deriveddata directory, so that the xctest would be able to resolve symbols that are built into the app.

Even after shuffling these bundles around the derivedData directory, I was not able to get this approach to work. I looked closely at how both Xcode and the bazel testrunner launch tests that are in a test host app. When these tests run successfully, the test runner itself launches the test host application first. In Xcode, this is apparent by looking at the console logs for test launches. In xctestrunner, there is a flag called "app-under-test" that is set at launch time.

I concluded that moving bundles in the derived data directory was not going to work. The necessary prequisite is for the test runner to be aware that the test bundle needs to be launched within an app.

Tulsi and XCHammer have also associated test host application xcode targets with stubbed Info.plist files and pointed xcode test targets to those test host apps.

File: rules/BUILD.bazel [hunk 4]
```diff
@@ -2,3 +2,7 @@ exports_files(
     glob(["*.bzl"]),
     visibility = ["//docs:__pkg__"],
 )
+
+exports_files([
+    "WorkspaceSettings.xcsettings",
```
segiddins, line 7: this file doesn't seem to exist in this package?

rules/xcodeproj.bzl
```diff
@@ -40,9 +42,13 @@ def _xcodeproj_aspect_impl(target, ctx):
             test_env_vars = tuple(env_key_value_pairs.items())
             commandlines_args = getattr(ctx.rule.attr, "args", [])
             test_commandline_args = tuple(commandlines_args)
+            test_host_target = getattr(ctx.rule.attr, "test_host", None)
+            if test_host_target and test_host_target.label.package != "rules/test_host_app":
```
segiddins, line 46: nit: should this also check the repository?
(I'm guessing this is here so we don't add in the default test host, or something like that? a comment to that effect would be helpful)

rules/xcodeproj.bzl
```diff
@@ -40,9 +42,13 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+                test_host_appname = test_host_target.label.name
```
segiddins, line 47: is this the name of the app, or the name of the target that produces the app?

tools/xcodeproj-shims/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amberdixon, line 1: The new build system leaves a folder in the DerivedData directory which (for some reason) causes incremental build and test attempts during local development to fail with an error from Xcode at launch time. The error message says "Cannot attach to pid." This error seems to happen in the Xcode IDE, not when the project is tested from the xcodebuild command.

rules/xcodeproj.bzl
segiddins: Looks like we might want the `bundle_name` from the `ios_application` target, rather than just its labels name. We might need to dig in and get it via a provider from that target

tools/xcodeproj-shims/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,10 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>BuildSystemType</key>
```
segiddins, line 5: note, this may not work with future versions of xcode (if the legacy build system is removed)

tools/xcodeproj-shims/xcodeproj-installer.sh
```diff
@@ -26,4 +27,7 @@ chmod -R +w "${tmp_dest}"
 # always trim three ../ from path, since that's "bazel-out/darwin-fastbuild/bin"
 sed -i.bak -E -e 's|([ "])../../../|\1|g' "${tmp_dest}/project.pbxproj"
 rm "${tmp_dest}/project.pbxproj.bak"
-rsync --recursive --quiet --copy-links "${tmp_dest}" "${dest}"
\ No newline at end of file
+rsync --recursive --quiet --copy-links "${tmp_dest}" "${dest}"
+
+mkdir -p "$dest/project.xcworkspace/xcshareddata/"
+mv "$(workspacesettings_xcsettings_short_path)" "$dest/project.xcworkspace/xcshareddata/"
```
segiddins, line 33: this should be a `cp`, rather than an `mv`, I think

rules/BUILD.bazel
amberdixon: will fix!

tools/xcodeproj-shims/WorkspaceSettings.xcsettings
amberdixon: Yes, this is frustrating. FYI Tulsi uses the old build system too. 

Since we know which subdirectory in intermediates is causing problems for us, we could probably just remove it ourselves, even if we are forced to use the new build system.

rules/xcodeproj.bzl
amberdixon: I realized that this logic is probably more correct. WDYT?

```
            # The ios_unit_test rule sets the test_host attribute to the rules_ios repo's rules/test_host_app target by default, if test_host = True.
            # The generated xcodeproject should only reference test host applications that live in the same workspace as the test target.
            if test_host_target.label.workspace_root == target.label.workspace_root:
                test_host_appname = test_host_target.label.name
```

rules/xcodeproj.bzl
amberdixon: It is the name of the bazel target that produces the app.

In the xcodeproject, the xcode test target should point to the xcode target name of the test host application, which should match the name of the bazel target that produces the app.

Does that all sound right?

rules/xcodeproj.bzl
segiddins: Will app-hosted tests that have cross-repository deps continue to work if you do that?

rules/xcodeproj.bzl
segiddins: That does sound right, but it seems this variable is also used for the _app_ name (when setting the TESTHOST build setting), which can be different from the target name

tools/xcodeproj-shims/WorkspaceSettings.xcsettings
gyfelton: should this be called out in a readme file or an explicit comment some where? This can be unexpected for new onboard person?

segiddins:
lgtm once the test is fixed up (looks like the test script might need to simctl create a new device to use for tests to run on)

tools/xcodeproj-shims/xcodeproj-installer.sh
amberdixon: I'll add a comment!

tools/xcodeproj-shims/WorkspaceSettings.xcsettings
amberdixon: I'll add a comment!

amberdixon:
> lgtm once the test is fixed up (looks like the test script might need to simctl create a new device to use for tests to run on)

I'm not sure why it was working before ...

rules/xcodeproj.bzl
```diff
@@ -180,28 +200,44 @@ def _xcodeproj_impl(ctx):
...
+            "PRODUCT_NAME": target_info.name,
+            "BAZEL_BIN_SUBDIR": target_info.bazel_bin_subdir,
+            "MACH_O_TYPE": target_macho_type,
+        }
+
+        if target_info.product_type == "application":
+            target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
+            target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
+        if target_info.product_type == "bundle.unit-test":
+            target_settings["SUPPORTS_MACCATALYST"] = False
```
amberdixon, line 213: For some reason, when running tests on using legacy build system, I had to set this to false in xcode build settings.

rules/xcodeproj.bzl
```diff
@@ -25,13 +25,20 @@ def _xcodeproj_aspect_impl(target, ctx):
     deps += getattr(ctx.rule.attr, "deps", [])
     deps += getattr(ctx.rule.attr, "infoplists", [])
     deps.append(getattr(ctx.rule.attr, "entitlements", None))
+    deps.append(getattr(ctx.rule.attr, "test_host", None))
 
     # TODO: handle apple_resource_bundle targets
     test_env_vars = ()
     test_commandline_args = ()
+    bazel_build_target_name = ""
+    if target.label.workspace_name != "":
```
gyfelton, line 34: @amberdixon what if we can make code reading easier by having example here about what each var might represent in real life (as comments), so that understanding of code can be easier in future?

amberdixon:
> What's your expectation for me in terms of checking out this branch and run sth (tests? new code?)
> The walkthrough is great btw!

I wouldn't expect you to run all the tests locally, though you can if you want. It may make it easier for you to review if you check out the branch locally. I'd be fine with you approving as long as you feel like you understand the changes and don't see any bugs.

PR #43, opened 2020-04-16
Avoid generating a CompileAssetCatalog build phase

amberdixon:
For xcasset sources, we do not want xcode to run actool. I'm not sure how to stub out the CompileAssetCatalog command, so it's easier to just load xcasset files into the xcodeproject, but not associate them with any build phase.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -31,7 +31,6 @@ def _xcodeproj_aspect_impl(target, ctx):
     test_commandline_args = ()
     if AppleBundleInfo in target:
         bundle_info = target[AppleBundleInfo]
-        srcs = []
         bazel_name = target.label.name
         if ctx.rule.kind == "ios_unit_test":
             env_key_value_pairs = getattr(ctx.rule.attr, "env", {})
@@ -47,7 +46,8 @@ def _xcodeproj_aspect_impl(target, ctx):
             bundle_extension = bundle_info.bundle_extension,
             package = target.label.package,
             bazel_name = bazel_name,
-            srcs = depset(srcs, transitive = _get_attr_values_for_name(deps, _SrcsInfo, "srcs")),
+            srcs = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "srcs")),
+            asset_srcs = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "asset_srcs")),
             build_files = depset([ctx.build_file_path], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "build_files")),
             product_type = bundle_info.product_type[len("com.apple.product-type."):],
             test_env_vars = test_env_vars,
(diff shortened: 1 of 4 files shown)


tests/xcodeproj/Single-Application-Project.xcodeproj/project.pbxproj
```diff
@@ -17,6 +18,9 @@
 		48A8F4B41EF706F8BEEC7ACB /* Single-Application-UnitTests.xctest */ = {isa = PBXFileReference; includeInIndex = 0; lastKnownFileType = wrapper.cfbundle; path = "Single-Application-UnitTests.xctest"; sourceTree = BUILT_PRODUCTS_DIR; };
 		5FF81A9585DBB110EBD9CAC0 /* test.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = test.swift; path = test.swift; sourceTree = "<group>"; };
 		8DBB99CE021DEB0B97F196EA /* common.pch */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; name = common.pch; path = ../../rules/library/common.pch; sourceTree = "<group>"; };
+		AB7BB9505A64B479DA80B398 /* resource_bundle.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist; name = resource_bundle.plist; path = ../../rules/library/resource_bundle.plist; sourceTree = "<group>"; };
+		BDFE1D6BB539968405F979CA /* Single_Application_UnitTests-Swift.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; name = "Single_Application_UnitTests-Swift.h"; path = "Single_Application_UnitTests-Swift.h"; sourceTree = "<group>"; };
```
amberdixon, line 22: TODO: `Single_Application_UnitTests-Swift.h` should probably not be here, as it is a generated file. There is some bug somewhere where `is_source` returns true for this file, causing the xcodeproj rule to add it into the project.

rules/xcodeproj.bzl
```diff
@@ -151,12 +151,21 @@ def _xcodeproj_impl(ctx):
...
     for target_info in targets:
         target_macho_type = "staticlib" if target_info.product_type == "framework" else "$(inherited)"
+        xcode_sources = []
+        for s in target_info.srcs.to_list():
+            src_info = {
+                            "path": paths.join(src_dot_dots, s.short_path),
+                            "group": paths.dirname(s.short_path),
+                            "optional": True,
+            }
+            if ".xcassets" in s.path:
```
segiddins, line 161: do we actually maybe want to filter out everything that would go into the `PBXResourcesBuildPhase`, instead of only asset catalogs?

amberdixon: I wasn't sure how to figure out what all these possible things were, but I would have preferred to do that.

segiddins: Take a look at `_bucketize` in https://github.com/bazelbuild/rules_apple/blob/9c1e460ec08af6e36673d6c6f34c723b09ae9328/apple/internal/resources.bzl -- I think everything explicitly called out there (e.g. not `"unprocessed"`) could be ignore.

Another approach would be to propagate 2 different fields in the provider -- one for explicit `srcs` that come from the `srcs` attribute and another with basically everything else. Everything not in `srcs` would get added without a build phase. That's probably the more elegant way of handling this, rather than hardcoding based on extensions

tests/xcodeproj/Single-Application-Project.xcodeproj/project.pbxproj
amberdixon: Sam fixed it.

rules/xcodeproj.bzl
```diff
@@ -31,7 +31,6 @@ def _xcodeproj_aspect_impl(target, ctx):
     test_commandline_args = ()
     if AppleBundleInfo in target:
```
gyfelton, line 32: I am trying to understand this if stmt because the code change applies to all blocks of this if stmts correct?
there are 
1. `AppleBundleInfo in target`
2. `ctx.rule.kind == "apple_framework_packaging"`
3. `else`
I wonder each is for what cases?

amberdixon: In practice:
1. `if AppleBundleInfo in target` is generally applied to targets that are unit test bundles or application bundles.
2. When the target is packaged up framework (this target usually depends on the components of a framework, like libraries, resource bundles, plists.)
3. Everything else. If the target is not performing a bundling/packaging operation, it often contains some kind of source files and dependencies with their source files.

justinseanmartin:
Sorry for the late review, but LGTM

PR #45, opened 2020-04-16
Update bazel version to 3.0.0 and latest rules 

gyfelton:
1. Update bazel to 3.0.0 and fixes a bug where we pass mutable values (dict/list) to `depset` for env var and launch arguments for tests. Instead we pass tuples to ensure correct data is passed in a safer way.
2. Added tests on actual project side to ensure they will fail should the passing of the env var and launch arg not done correcetly.
3. Fixed test case failure due to latest bazel_swift rule update due to adding of modules maps from non-swift code, which is sth we don't anticipate.


rules/repositories.bzl
```diff
@@ -27,15 +27,15 @@ def rules_ios_dependencies():
     _maybe(
         git_repository,
         name = "build_bazel_rules_apple",
```
segiddins, line 29: I think `build_bazel_apple_support` might also need a bump

README.md
```diff
@@ -90,5 +85,3 @@ ios_application(
 )
 ```
 
-See the [examples](https://github.com/ob/rules_ios/tree/master/examples)
```
segiddins, line 93: why is this stuff removed from the README?

rules/repositories.bzl
```diff
@@ -27,15 +27,15 @@ def rules_ios_dependencies():
...
-        commit = "fe8cf69b058c86624ba1786e6024c2a1af1d48e9",
+        commit = "32ede960cb9029be15890622366eb5598ceebc74",
         shallow_since = "1581368998 -0800",
```
segiddins, line 31: `DEBUG: Rule 'build_bazel_rules_apple' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1587042139 -0700"`

rules/repositories.bzl
```diff
@@ -27,15 +27,15 @@ def rules_ios_dependencies():
...
         remote = "https://github.com/bazelbuild/rules_apple.git",
     )
 
     _maybe(
         git_repository,
         name = "build_bazel_rules_swift",
-        commit = "657eda57a097980848dbb3c880b94faeddaa3cd1",
```
segiddins, line 38: `DEBUG: Rule 'build_bazel_rules_swift' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1586818875 -0700"`

README.md
gyfelton: 404, no examples in master branc

gyfelton:
@segiddins does it actually check if the generated xcodeproj works? I saw there is a check to see if bazel generates a different xcodeproj file, but it does not actuall test the project. I am at least adding some tests that will fail if I did not do the passing of test env correctly

rules/repositories.bzl
gyfelton: nope we are still pointing at HEAD of master

README.md
```diff
@@ -14,12 +14,7 @@ in Bazel, and by the
 [`swift_library` rule](https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md#swift_library)
 available from [rules_swift](https://github.com/bazelbuild/rules_swift).
 
-These rules require Bazel 2.0.
-
-## Reference documentation
-
-[Click here](https://github.com/ob/rules_ios/tree/master/doc)
-for the documentation.
+Bazel version required by tThese rules is [here](https://github.com/ob/rules_ios/blob/master/.bazelversion)
```
segiddins, line 17: typo

README.md
```diff
@@ -14,12 +14,7 @@ in Bazel, and by the
 [`swift_library` rule](https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md#swift_library)
 available from [rules_swift](https://github.com/bazelbuild/rules_swift).
 
-These rules require Bazel 2.0.
-
-## Reference documentation
-
-[Click here](https://github.com/ob/rules_ios/tree/master/doc)
```
segiddins, line 21: should still link to the docs

rules/framework.bzl
```diff
@@ -155,6 +155,12 @@ def _apple_framework_packaging_impl(ctx):
 
             # collect modulemaps
             for modulemap in dep[apple_common.Objc].direct_module_maps:
+                # rule_swift changed how non swift generates module map in this commit
```
segiddins, line 158: thanks for the comment here!

tests/xcodeproj/Single-Application-Project.xcodeproj/project.pbxproj
```diff
@@ -17,6 +17,7 @@
 		48A8F4B41EF706F8BEEC7ACB /* Single-Application-UnitTests.xctest */ = {isa = PBXFileReference; includeInIndex = 0; lastKnownFileType = wrapper.cfbundle; path = "Single-Application-UnitTests.xctest"; sourceTree = BUILT_PRODUCTS_DIR; };
 		5FF81A9585DBB110EBD9CAC0 /* test.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = test.swift; path = test.swift; sourceTree = "<group>"; };
 		8DBB99CE021DEB0B97F196EA /* common.pch */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; name = common.pch; path = ../../rules/library/common.pch; sourceTree = "<group>"; };
+		BDFE1D6BB539968405F979CA /* Single_Application_UnitTests-Swift.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; name = "Single_Application_UnitTests-Swift.h"; path = "Single_Application_UnitTests-Swift.h"; sourceTree = "<group>"; };
```
segiddins, line 20: cc @amberdixon on this getting added

gyfelton: is this expected?

README.md
gyfelton: will point to `https://github.com/bazel-ios/rules_ios/tree/master/docs` same for the other deletion i made

gyfelton:
@segiddins While u continue to review, I am going to test this with our app repo and inform u on what works and what does not

tests/xcodeproj/Single-Application-Project.xcodeproj/project.pbxproj
amberdixon: I still don't know why this is getting added, but I'm seeing this in my PRs too. The underlying reason is that the Swift.h file is being treated as a source file (i.e. running `File.is_source` on it returns true), even though the file is actually being generated by one of the rules in rules_swift. Perhaps we can leave it here and make a note to remove it later, after figuring out why it's being added.

tests/xcodeproj/test.swift
```diff
@@ -4,5 +4,12 @@ class SwiftTests : XCTestCase {
   func testPasses() {
     XCTAssertTrue(true)
   }
+  func testTestEnvArgsMatches() {
+    XCTAssertEqual(ProcessInfo.processInfo.environment["test_envvar_key1"], "test_envvar_value1")
```
amberdixon, line 8: This is fine to add here, but FYI, we aren't actually running these tests in CI.

I will probably run these tests as part of CI in a separate PR.

rules/framework.bzl
amberdixon: I'm confused about the original implementation here, as well as the change. If we assign modulemap_in to the modulemap on every iteration, aren't we just assigning modulemap_in with the last value in the array `direct_module_maps` ?

gyfelton: i have the same question actually! @segiddins

tests/xcodeproj/test.swift
gyfelton: I am actually considering to try to add this into CI in this PR (otherwise my change has no effect)

tests/xcodeproj/Single-Application-Project.xcodeproj/project.pbxproj
gyfelton: Do we have expectation that no file change should happen to project file?

amberdixon: yes, because there is a check in CI for `git diff --nocheck` ... so if the xcodeproj rule generates a project that is different from the one checked in already, then the CI job will fail.

I'm not sure why the -Swift.h file is showing up in the bazel target's sources as an actual source (because it's really a generated file.) Because it appears to be an actual source, the xcodeproj rule adds this file to the generated xcodeproject.

gyfelton: Will deal with this in future: https://github.com/bazel-ios/rules_ios/issues/49

gyfelton:
@amberdixon added the CI job in the workflow, can take a look?

tests/xcodeproj/tests.sh
```diff
@@ -0,0 +1,5 @@
+set -eux
+
+cd $(dirname $0)
+
+xcodebuild -project Single-Application-Project.xcodeproj -quiet -scheme Single-Application-UnitTests test
```
amberdixon, line 5: thanks for putting these in! LGTM

gyfelton:
waiting for Sam's PR to merge so that I don't add the `-Swift.h` generated file

PR #47, opened 2020-04-17
Add support for specifying a dict of infoplist values

segiddins:
This allows specifying entries directly in the BUILD file,
rather than requiring users to write out a plist file themselves

This will allow cocoapods-bazel to emit `app_spec.info_plist`
specifications into BUILD files, keeping the behavior whereby no
files outside of BUILD files need to be written

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -14,6 +14,16 @@ _IOS_APPLICATION_KWARGS = [
 ]
 
 def write_info_plists_if_needed(name, plists):
+    """
+    Writes info plists for an app if needed.
+
+    Given a list of infoplists, will write out any plists that are passed as a
+    dict, and will add a default app Info.plist if no non-dict plists are passed.
+
+    Args:
+        name: The name of the app target these infoplists are for.
+        plists: A list of either labels or dicts.
+    """
     already_written_plists = []
     written_plists = []
     for idx, plist in enumerate(plists):
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -1,3 +1,4 @@
+load("@bazel_skylib//lib:types.bzl", "types")
 load("@build_bazel_rules_apple//apple:ios.bzl", rules_apple_ios_application = "ios_application")
 load("//rules:library.bzl", "apple_library", "write_file")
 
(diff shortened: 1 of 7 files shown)


rules/app.bzl
```diff
@@ -12,6 +13,36 @@ _IOS_APPLICATION_KWARGS = [
     "launch_storyboard",
 ]
 
+def write_info_plists_if_needed(name, plists):
```
justinseanmartin, line 16: I know this precedes this change, but when do we expect to see more than 1 plist supplied?

rules/app.bzl
```diff
@@ -12,6 +13,36 @@ _IOS_APPLICATION_KWARGS = [
...
+                name = plist_name,
+                destination = plist_name + ".plist",
+                content = struct(**plist).to_json(),
+            )
+            written_plists.append(plist_name)
+        else:
+            already_written_plists.append(plist)
+
+    if not already_written_plists:
+        already_written_plists.append("@build_bazel_rules_ios//rules/test_host_app:Info.plist")
```
justinseanmartin, line 42: Worth adding a test covering the default behavior? Or not worthwhile because no app hosted tests would be launchable?

rules/app.bzl
```diff
@@ -12,6 +13,36 @@ _IOS_APPLICATION_KWARGS = [
...
+def write_info_plists_if_needed(name, plists):
+    """
+    Writes info plists for an app if needed.
+
+    Given a list of infoplists, will write out any plists that are passed as a
+    dict, and will add a default app Info.plist if no non-dict plists are passed.
+
+    Args:
+        name: The name of the app target these infoplists are for.
+        plists: A list of either labels or dicts.
```
justinseanmartin, line 25: Can you explain what `labels` would look like and when it would be used? It would be the path to an infoplist file? It looks like you are exercising a single dict path in the test that was added. Worth adding a test that includes the other form? Or does that already have coverage?

rules/app.bzl
segiddins: `rules_apple` allows passing multiple plists and it coalesces them, so we're just supporting the same interface as the apple rules for consistency

rules/app.bzl
segiddins: they wouldn't be launchable without the default, so it's defacto covered by the existing test tests

rules/app.bzl
segiddins: The label would either be a path to a source plist file, or the label of a generated file. Bazel in general doesn't understand paths, instead using labels to refer to files when instantiating rules, and every source file implicitly gets a label that matches its path

rules/app.bzl
```diff
@@ -12,6 +13,36 @@ _IOS_APPLICATION_KWARGS = [
...
+                content = struct(**plist).to_json(),
+            )
+            written_plists.append(plist_name)
+        else:
+            already_written_plists.append(plist)
+
+    if not already_written_plists:
+        already_written_plists.append("@build_bazel_rules_ios//rules/test_host_app:Info.plist")
+
+    return already_written_plists + written_plists
```
gyfelton, line 44: I don't understand what this method returns as? A list of written plists? The doc does not specify it. This will also helps me understand the doc better cos I saw 3 things happen inside the method when doc says there are 2 things happened (from my view)

rules/app.bzl
justinseanmartin, line 42: Does this need to be added in the case that a dict is supplied?

rules/app.bzl
segiddins: It returns a list of labels to plists.

What thing do you think is missing from the doc?

rules/app.bzl
segiddins: Yes, to match cocoapods' behavior

PR #48, opened 2020-04-21
Refer to actual generated swift header when making swift hmap

segiddins:
Before, the file in hdrs referred to a non-existant source file

We can see the fix in effect by the xcodeproj build not failing, since the `-Swift.h` header is no longer getting added to the xcode project !

cc @wileykestner

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,3 @@
 # We can't create a bzl_library for rules-swift because of its visibility,
 # so circumvent by not using the sandbox
-common --strategy=Stardoc=standalone
+build --strategy=Stardoc=standalone
--- a/rules/hmap.bzl
+++ b/rules/hmap.bzl
@@ -42,9 +42,15 @@ def _make_headermap_impl(ctx):
 
     # Write a file for *this* headermap, this is a temporary file
     input_f = ctx.actions.declare_file(ctx.label.name + "_input.txt")
-    all_hdrs = []
-    for provider in ctx.attr.hdrs:
-        all_hdrs += provider.files.to_list()
+    all_hdrs = list(ctx.files.hdrs)
+    for provider in ctx.attr.direct_hdr_providers:
+        if apple_common.Objc in provider:
+            all_hdrs += provider[apple_common.Objc].direct_headers
+        elif CcInfo in hdr_provider:
+            all_hdrs += provider[CcInfo].compilation_context.direct_headers
+        else:
+            fail("direct_hdr_provider %s must contain either 'CcInfo' or 'objc' provider" % provider)
+
     out = _make_headermap_input_file(ctx.attr.namespace, all_hdrs, ctx.attr.flatten_headers)
     ctx.actions.write(
         content = out,
@@ -68,22 +74,6 @@ def _make_headermap_impl(ctx):
             fail("hdr_provider must contain either 'CcInfo' or 'objc' provider")
 
         for hdr in hdrs:
-            if SwiftInfo in hdr_provider and hdr.path.endswith("-Swift.h"):
-                namespace = ctx.attr.namespace
-                basename = hdr.basename
-
-                # Only propogate the Swift header from this module
-                # The name of the swift header may be -Swift.h or _Swift-Swift.h
-                # dur to bazelizer generated rule naming convention.
-                # Because rules_swift outputs the C header with name of the rule not
-                # the module.
-                normalized_basename = namespace + "-Swift.h"
-                if basename in [normalized_basename, namespace + "_Swift-Swift.h"]:
-                    inputs.append(hdr)
-                    mappings += [namespace + "/" + normalized_basename + "|" + hdr.path]
-                    if ctx.attr.flatten_headers:
-                        mappings += [normalized_basename + "|" + hdr.path]
-
             # only merge public header maps
             if hdr.path.endswith("public_hmap.hmap"):
                 # Add headermaps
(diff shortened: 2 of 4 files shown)


rules/hmap.bzl
```diff
@@ -42,9 +42,15 @@ def _make_headermap_impl(ctx):
...
+    for provider in ctx.attr.direct_hdr_providers:
+        if apple_common.Objc in provider:
+            all_hdrs += provider[apple_common.Objc].direct_headers
+        elif CcInfo in hdr_provider:
+            all_hdrs += provider[CcInfo].compilation_context.direct_headers
+        else:
+            fail("direct_hdr_provider %s must contain either 'CcInfo' or 'objc' provider" % provider)
+
     out = _make_headermap_input_file(ctx.attr.namespace, all_hdrs, ctx.attr.flatten_headers)
     ctx.actions.write(
```
amberdixon, line 55: The `mappings` variable defined a few lines below is no longer being used, I think. It seems like it remains empty now.

rules/library.bzl
```diff
@@ -456,22 +440,21 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-    objc_copts += headermap_copts + [
+    objc_copts += [
         "-fmodules",
         "-fmodule-name=%s" % module_name,
         "-gmodules",
     ]
 
-    swift_copts += [j for i in ([["-Xcc", copt] for copt in headermap_copts]) for j in i] + [
+    swift_copts += [
         "-Xcc",
```
amberdixon, line 458: it seems like you already added `-Xcc` on line 448.

.bazelrc
```diff
@@ -1,3 +1,3 @@
 # We can't create a bzl_library for rules-swift because of its visibility,
 # so circumvent by not using the sandbox
-common --strategy=Stardoc=standalone
+build --strategy=Stardoc=standalone
```
amberdixon, line 3: Why does this setting need to change?

segiddins: `--strategy` is apparently not valid for `query`

rules/library.bzl
segiddins: You need an `-Xcc` for every argument that you want passed to clang

rules/hmap.bzl
segiddins: good catch! I'll delete the code referencing it

rules/hmap.bzl
```diff
@@ -42,17 +42,22 @@ def _make_headermap_impl(ctx):
 
     # Write a file for *this* headermap, this is a temporary file
     input_f = ctx.actions.declare_file(ctx.label.name + "_input.txt")
-    all_hdrs = []
-    for provider in ctx.attr.hdrs:
-        all_hdrs += provider.files.to_list()
+    all_hdrs = list(ctx.files.hdrs)
+    for provider in ctx.attr.direct_hdr_providers:
+        if apple_common.Objc in provider:
```
ob, line 47: Ah, I assume swift rules provide the `-Swift.h` header in an `objc` provider now no?

segiddins: yeah, they're in `direct_headers`

rules/library.bzl
```diff
@@ -533,6 +515,20 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
         )
         lib_names += [swift_libname]
+
+        # Add generated swift header to header maps for double quote imports
+        swift_doublequote_hmap_name = name + "_swift_doublequote_hmap"
+        headermap(
+            name = swift_doublequote_hmap_name,
+            namespace = namespace,
+            hdrs = [],
+            direct_hdr_providers = [swift_libname],
```
wileykestner, line 525: Echoing @ob's question from above, I assume what we're doing here is passing an instantiated `swift_library` rule into the `headermap` rule.  In my beginner Bazel adopters mind, what's happening here isn't that we are passing some artifact, we are passing the information that the instantiated `swift_library` rules wants its consumers it to have: in this case, `swift_library` generates a `-Swift.h` header.  In Bazel terms, the information an instantiated rule wants its consumers to be able to access is called a `Provider`, and `Provider`s can be thought of as the "return value" of the instantiating a rule.

In this case, a `headermap` rule would like to include information about the instantiated `swift_library`s -Swift.h`header into a header map, so that ObjC classes in targets that consume this `apple_library`  can call methods on the Swift classes that the `swift_library` exposes in its header.

Is that more or less what's happening here?

 so here we pass the instantiated `swift_libname` (a provider) into `direct_hdr_providers` 

Inside the `headermap` rule's implementation, the `direct_hdr_providers` list  a `Provider` (@ob asked if was an ObjC provider - I don't have any intuition for what it is).  My assumption is that

ob: That is exactly what’s going on.

wileykestner: Awesome.  Thanks for confirming @ob!

.bazelrc
gyfelton: thx for fixing this! I just ran into it

segiddins:
> Was the problem we were trying to solve here that we had an extra -Swift.h file? If so, is it possible to write a test to show that we no longer generate an extra -Swift.h file?

It wasn't actually generated, but the hmap had a reference to a _source file_ with the name `ModuleName-Swift.h`. That file didn't exist, but we didn't catch it because nothing in the `hmap` rule actually uses the headers as _inputs_ to any _actions_, so bazel was happy to continue passing it along, even though it didn't exist.

This showed up in the Xcodeproj rule, since that rule gathers all transitive source (e.g. not generated) files, and the -Swift.h file was showing up there, even though it (should have been a) generated file.

PR #51, opened 2020-04-22
Fix swift's clang importer being able to see hmaps

segiddins:
* Remove public hmap propagation
  It can only confuse clang, since all public headers are now visible via frameworks

* Update shallow_since for rules_swift

* Fix typo in hmap impl

* Stop quoting the path to hmaps
  objc_library already have the quotes stripped, and the extra quotes break swift_library from actually seeing the headermaps

### DIFF
--- a/docs/hmap_doc.md
+++ b/docs/hmap_doc.md
@@ -5,7 +5,7 @@
 ## headermap
 
 <pre>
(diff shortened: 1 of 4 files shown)


segiddins:
> Fair to say the entry point is hdr_providers = deps where we are specifying all deps as hdr providers?

Yes, and that's why we're removing it -- we want to specify headers explicitly

PR #52, opened 2020-04-24
Add a more concise input format for header maps

ob:
The `hmaptool` header map builder takes input files with the format

```
  KEY|VALUE
```

which is pretty flexible and generic. However, in practice, the key can always be derived from the value because the value is a path and the key is the basename with a namespace optionally prepended.

This change removes the `KEY` portion adding instead a `--namespace` CLI argument. It also adds support for _param files_ and removes unused options (`--merge-hmaps` and `--add-mappings`. This allows the Starlark rules to use `ctx.actions.args()` for writing the input files.

**BREAKING CHANGE:** The `flatten_headers` field has been removed and `namespace` has been made optional. If `namespace` is present, then namespaced headers are added, if it's not, then the header map will have just `basename(header) -> header` mappings.


rules/hmap.bzl
```diff
@@ -113,7 +86,7 @@ headermap = rule(
             mandatory = False,
             doc = "Targets whose direct headers should be added to the list of hdrs",
         ),
-        "flatten_headers": attr.bool(
+        "namespace_headers": attr.bool(
```
segiddins, line 89: I wonder if we could remove this, and just namespace when a `namespace` is passed?

rules/hmap.bzl
```diff
@@ -51,16 +23,17 @@ def _make_headermap_impl(ctx):
         else:
             fail("direct_hdr_provider %s must contain either 'CcInfo' or 'objc' provider" % provider)
 
-    out = _make_headermap_input_file(ctx.attr.namespace, all_hdrs, ctx.attr.flatten_headers)
     ctx.actions.write(
```
segiddins, line 26: instead of manually writing this file, could we use `ctx.action.args()` to create a params file, so we don't ever need to materialize all the headers to a list during analysis?

rules/hmap.bzl
ob: Yeah, that's a good idea. I'll pick it up after we merge #61

rules/hmap.bzl
ob: I started going down this path but there are more changes required since the `hmapbuild` tool doesn't really support param files. Once the cleanups in #61 are in I can add param file support and turn this on.

segiddins: i meant just cheating to use the param file to write out the list, instead of using it as a legit param file. Feel free to land as-is and I can try that out

ob: I tried that first but it’s not that simple. You can’t really control _when_ Bazel will stop adding args and use a param file. You can set it to _always_ use a param file though.

ob:
I did a memory consumption and performance comparison for the tests in this same repo.

Memory usage goes from 117MB to 112MB and performance looks like this:


image

It's probably worth trying this in a bigger project. The way to do it is:

```
$ bazelisk build --profile=master-trace-profile.json --experimental_generate_json_trace_profile --experimental_profile_cpu_usage //...
$ bazelisk info used-heap-size-after-gc
$ cat master-trace-profile.json | jq -r '.traceEvents | .[] | select(.name | contains("HmapCreate")) | [.name , .dur] | @csv' > master.csv
```
Then do the same with the `hmapv2` branch and plot it in R with a program like:

```
library(ggplot2)
library(reshape2)

hmapv2 <- read.csv("hmapv2.csv", header=FALSE, col.names=c("hmap", "dur"))
master <- read.csv("master.csv", header=FALSE, col.names=c("hmap", "dur"))

hmapv2$branch <- "hmapv2"
master$branch <- "master"

data <- rbind(hmapv2, master)
ggplot(data, aes(x=branch, y=dur/1000)) + geom_boxplot() + labs(y="duration (ms)")
```

rules/hmap/hmapbuild.c
```diff
@@ -125,6 +89,59 @@ static void debug(char *format, ...) {
...
+                cli_args.name_space = strdup(optarg);
+                break;
+            case 'o':
+                cli_args.output_file = strdup(optarg);
+                break;
+            case 'v':
+                verbose = 1;
+                break;
+            default:
+                // TODO: don't
```
segiddins, line 107: TODO

rules/hmap/hmapbuild.c
```diff
@@ -125,6 +89,59 @@ static void debug(char *format, ...) {
...
+        }
+    }
+    ac -= optind;
+    av += optind;
+
+    // all remaining arguments are the actual headers
+    for (; *av; av++) {
+        if (**av == '@') {
+            // param file
+            parse_param_file(entries, *av);
```
segiddins, line 118: niiiiiice

rules/hmap.bzl
```diff
@@ -50,33 +20,31 @@ def _make_headermap_impl(ctx):
...
-    ctx.actions.write(
-        content = out,
-        output = input_f,
-    )
-
-    # Add a list of headermaps in text or hmap format
-    inputs = [input_f]
-    args = [input_f.path, ctx.outputs.headermap.path]
+    args.set_param_file_format(format = "multiline")
+    args.use_param_file("@%s", use_always = True)
```
segiddins, line 40: I don't think we need `use_always` anymore?

segiddins:
hmapv2 definitely looks better, but I think removing the "_framework_public_hmap" will have an even larger impact image

rules/hmap/hmapbuild.c
ob, line 107:
```suggestion
```

rules/hmap/hmapbuild.c
ob: 😁

rules/hmap.bzl
ob:
```suggestion
    args.use_param_file("@%s")
```
Yeah

rules/framework.bzl
```diff
@@ -209,24 +209,19 @@ def _apple_framework_packaging_impl(ctx):
...
-        output = mappings_file,
-    )
+    hmap_file = ctx.actions.declare_file(framework_name + "_framework_public_hmap.hmap")
+    args = ctx.actions.args()
+    args.add("--output", hmap_file.path)
+    args.add("--namespace", framework_name)
+    args.add_all(header_in)
+    args.add_all(private_header_in)
+    args.set_param_file_format("multiline")
+    args.use_param_file("@%s", use_always = True)
```
ob, line 219:
```suggestion
    args.use_param_file("@%s")
```

rules/hmap/hmapbuild.c
ob: removed

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           name: bazel-testlogs
           path: bazel-testlogs
   buildifier:
-    name: Check Starlark
+    name: Check Starlark adn Docs
```
segiddins, line 25: `and`

ob, line 25:
```suggestion
    name: Check Starlark and Docs
```

PR #53, opened 2020-04-28
Use Swift info provider during framework packaging to avoid symbol redefinition

gyfelton:
With updates on rules_swift, if a framework `Foo` depends on target `Bar1` and `Bar2`, while `Bar1` also depends on `Bar2`, without changes in this PR, bazel will produce 2 module maps for Bar2, causes symbol redefinition error (in `bazel-out`, it will produce Bar2.framework and `Bar2-modulemap` which contains `Bar2.modulemap`).

The fix is to provide the swift_info while packing framework `Foo` so that `rules_swift` won't produce the `Bar2-modulemap`.

This PR also points to latest `rules_swift` and `rules_apple` that we can support. 
This PR also removes a patch fix that is no longer useful with the intro of swift_info
Major work is done by @segiddins in this PR


segiddins:
> what is the output of this implementation? Is [objc_provider, swift_provider, default_info_provider] different from [swift_provider, objc_provider, default_info_provider] when I swap their order?

The order doesn't matter, it's functionally a set of the different providers -- there can only be one provider of each "type" per target returned

> what it means to have ** in front of objc_provider_fields?

It's a "splat" that turns key-value pairs from a dictionary into keyword arguments

> Is SwiftInfo the counter part of apple_common.Objc? Counter part meaning they are representation of a source file, where one is for objc class and one is for swift?

yes, they are counterparts. They don't represent source files, but rather are bazel "providers", which are the ways of propagating information about how to consume the target to the other targets that depend on this one

rules/framework.bzl
```diff
@@ -25,7 +25,7 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
 
 def _find_framework_dir(outputs):
     for output in outputs:
-        prefix = output.path.split(".framework/")[0]
+        prefix = output.path.rsplit(".framework/", 1)[0]
```
gyfelton, line 28: can give an example of how this changes what's inside the `prefix` variable?

rules/framework.bzl
```diff
@@ -274,13 +272,42 @@ def _apple_framework_packaging_impl(ctx):
...
+            )
+
+        swift_info_fields["modules"] = [
+            # only add the swift module, the objc modulemap is already listed as a header,
+            # and it will be discovered via the framework search path
+            swift_common.create_module(name = swiftmodule_name, swift = swift_module),
+        ]
+
+    return [
+        apple_common.new_objc_provider(**objc_provider_fields),
```
gyfelton, line 307: if no change on objc_provider, why not having `objc_provider = apple_common.new_objc_provider(**objc_provider_fields)` as b4?

rules/framework.bzl
```diff
@@ -274,13 +272,42 @@ def _apple_framework_packaging_impl(ctx):
...
-    objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
-    swift_provider = swift_common.create_swift_info(swift_infos=swift_infos)
-    default_info_provider = DefaultInfo(files = depset(framework_files))
-    # vfs_overlay_provider = VFSOverlay(
-    #     files = depset(items = file_map, transitive = [dep[VFSOverlay].files for dep in ctx.attr.transitive_deps if VFSOverlay in dep])
-    # )
-    return [objc_provider, swift_provider, default_info_provider]
+    # gather swift info fields
+    swift_info_fields = {
+        "swift_infos": [dep[SwiftInfo] for dep in ctx.attr.transitive_deps if SwiftInfo in dep],
```
gyfelton, line 277: for each of the transitive dependencies of ctr.attr (the current context we are in), if there is SwiftInfo in dep, take it out and add to an array. This array then becomes the value with key = "swift_infos"

rules/framework.bzl
```diff
@@ -274,13 +272,42 @@ def _apple_framework_packaging_impl(ctx):
...
+    }
+
+    if swiftmodule_out:
+        # only add a swift module to the SwiftInfo if we've actually got a swiftmodule
+        swiftmodule_name = paths.split_extension(swiftmodule_in.basename)[0]
```
gyfelton, line 282: example of `paths`? I am confused on this one
Also why not just use info inside swiftmodule_in as the `swiftmodule_name `? I would look at what's actually inside the swiftmodule_in to figure this part out

rules/framework.bzl
segiddins: It only would if `output.path` contained multiple instances of `.framework`, in which case now it will return the innermost framework, instead of the outermost

rules/framework.bzl
segiddins: there's no reason to capture the providers in locals, since they're only returned and never read from.

rules/framework.bzl
```diff
@@ -274,13 +272,42 @@ def _apple_framework_packaging_impl(ctx):
...
+        swift_module = swift_common.create_swift_module(
+            swiftdoc = swiftdoc_out[0],
+            swiftmodule = swiftmodule_out[0],
+        )
+
+        if swiftmodule_name != framework_name:
+            # Swift won't find swiftmodule files inside of frameworks whose name doesn't match the
+            # module name. It's annoying (since clang finds them just fine), but we have no choice but to point to the
+            # original swift module/doc, so that swift can find it.
+            swift_module = swift_common.create_swift_module(
```
gyfelton, line 295: so this overrides the swift_module we just created because its name does not match the framework name...

rules/framework.bzl
gyfelton: so `a.framework/b/c.framework/c` to give `a.framework/b/c.framework`

tests/frameworks/frameworkBar/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("//rules:framework.bzl", "apple_framework_packaging")
+load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
+
+swift_library(
+    name = "frameworkBar",
```
acecilia, line 5: What do you mean by `frameworkBar`? What is being tested in this example?

gyfelton: Can give some suggestion on the name? In this case what happened is there is a target to compile a swift library called `frameworkBar`  that depends `FrameworkFoo`, then it was packaged into a framework with a name `Unique_FrameworkBar ` which also depends on `FrameworkFoo`. Without the change introduced in this PR, it will result in redefinition of module `FrameworkBar` because it is not properly propagated through the swift info. Does this makes sense?

PR #56, opened 2020-05-03
Add target.info to list of all dependent targets

amberdixon:
If `include_transitive_targets` is true for a target that uses the xcodeproject rule, make sure to include the actual target in the provider's list for `all_dep_targets`.


tests/xcodeproj/BUILD.bazel
```diff
@@ -14,13 +15,56 @@ ios_unit_test(
...
+    module_name = "Single_Application_UnrunnableTestSuite",
+    runners = [],
+    sdk_frameworks = ["XCTest"],
+)
+
 xcodeproj(
     name = "Single-Application-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    include_transitive_targets = True,
```
gyfelton, line 62: Is this a new param available from xcodeproj rule? Or it is from somewhere else? I am always confused about this because after all it (to me) just shows up in the middle of no where :)

tests/xcodeproj/tests.sh
```diff
@@ -2,4 +2,5 @@ set -eux
 
 cd $(dirname $0)
 
-xcodebuild -project Single-Application-Project.xcodeproj -quiet -scheme Single-Application-UnitTests test
\ No newline at end of file
+xcodebuild -project Single-Application-Project.xcodeproj -quiet -scheme Single-Application-UnitTests test
+xcodebuild -project Single-Application-Project.xcodeproj -quiet -scheme Single-Application-RunnableTestSuite test
```
gyfelton, line 6: What's the diff between runnable and unrunnable?

tests/xcodeproj/BUILD.bazel
amberdixon: No, this param/attribute was already available in the xcodeproj rule. It has a default value of false.

This bug is only exhibited when this attribute is set to True.

If `include_transitive_targets` is false, the xcodeproject rule reads the "target" property of the provider returned by the aspect. This target just refers to a single target.

But if `include_transitive_targets` is true, the xcodeproject rule reads the "all_deps_target" property of the provider returned by the aspect. (I renamed "targets" to "all_deps_target", which is a more descriptive name imo.) The "all_deps_target" property refers to all the targets that this target depends on (directly and transitively.)

This PR fixes a bug. When the aspect is applied to a test_suite, prior to my changes, the "target" property of the provider would refer to the test_suite, but the "all_deps_target" did not refer to the test_suite itself ... and as a result, the xcodeproject would also fail to pick up the test_suite.

tests/xcodeproj/tests.sh
amberdixon: Ah, please disregard that for now, I removed the Unrunnable-Test-Suite in my latest changes.

The runnable test suite specifies a non-empty list for the attribute called `runners`. A runner is a combination of device-type and os-version (i.e. `iPhone6-9.3` or `iPad-Air-11.4`).

The Unrunnable test suite was testing what happens when the user specifies a target whose runners attribute is an empty list. After I observed that `bazel build` would fail on the Unrunnable-Test-Suite bazel target, I didn't see a great way of testing that case, so I removed that test.

rules/xcodeproj.bzl
```diff
@@ -61,7 +61,7 @@ def _xcodeproj_aspect_impl(target, ctx):
                 direct_srcs = [],
             ),
         )
-        target_info = _TargetInfo(target = info, targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
+        target_info = _TargetInfo(target = info, all_dep_targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "all_dep_targets")))
```
segiddins, line 64: im not sure this rename feels very "bazelly". In general, plural fields in a provider as assumed to return a depset, which will contain transitive information. If you did want to clean this up, renaming `target` to `direct_target` would make sense, as other providers use the `direct_*` fields to communicate info about the direct target that is returning that provider

tests/xcodeproj/BUILD.bazel
segiddins: Should we make a new project that sets `include_transitive_targets = True`, so we can have test coverage of both cases?

gyfelton: Thx for the explaination!

PR #60, opened 2020-05-13
Mixed module fixes

segiddins:
* Allow angle bracket imports of -Swift.h headers

* Ensure generated umbrella headers get added to hmaps

* Update tests with new cases

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -517,7 +517,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         )
         lib_names += [swift_libname]
 
-        # Add generated swift header to header maps for double quote imports
+        # Add generated swift header to header maps for angle bracket imports
         swift_doublequote_hmap_name = name + "_swift_doublequote_hmap"
         headermap(
             name = swift_doublequote_hmap_name,
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -444,9 +444,6 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     _append_headermap_copts(private_hmap_name, "-I", objc_copts, swift_copts, cc_copts)
     _append_headermap_copts(public_hmap_name, "-I", objc_copts, swift_copts, cc_copts)
     _append_headermap_copts(private_angled_hmap_name, "-I", objc_copts, swift_copts, cc_copts)
-    objc_copts.append("-I.")
-    cc_copts.append("-I.")
-    swift_copts.extend(("-Xcc", "-I."))
     _append_headermap_copts(private_hmap_name, "-iquote", objc_copts, swift_copts, cc_copts)
 
     objc_copts += [
@@ -493,6 +490,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         )
 
     if swift_sources:
+        swift_copts.extend(("-Xcc", "-I."))
         if module_map:
             swift_copts += [
                 "-Xcc",
@@ -503,6 +501,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         if module_map:
             swiftc_inputs.append(module_map)
         generated_swift_header_name = module_name + "-Swift.h"
+
         swift_library(
             name = swift_libname,
             module_name = module_name,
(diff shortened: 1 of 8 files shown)


rules/library.bzl
```diff
@@ -468,31 +490,8 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-            umbrella_header = paths.basename(umbrella_header),
-            public_headers = objc_hdrs,
-            private_headers = objc_private_hdrs,
-            module_name = module_name,
-            framework = False if swift_sources else True,
-            **kwargs
-        )
-
     if swift_sources:
+        swift_copts.extend(("-Xcc", "-I."))
```
gyfelton, line 494: what does `-I.` do again? Since it is used in multiple places down there, wonder if creating a macro representing this in this file will make it more readable

rules/library.bzl
```diff
@@ -381,6 +381,31 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     )
     deps += resource_bundles
 
+    # TODO: remove framework if set
```
gyfelton, line 384: does this mean when framework is set somewhere, we should remove this code? not clear to me what TODO wants

rules/library.bzl
```diff
@@ -524,12 +524,25 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
             direct_hdr_providers = [swift_libname],
-            flatten_headers = True,
+            flatten_headers = False,
             tags = _MANUAL,
         )
         internal_deps.append(swift_doublequote_hmap_name)
         _append_headermap_copts(swift_doublequote_hmap_name, "-iquote", objc_copts, swift_copts, cc_copts)
 
+        # Add generated swift header to header maps for double quote imports
+        swift_angle_bracket_hmap_name = name + "_swift_angle_bracket_hmap"
```
gyfelton, line 534: so this is for imports with format `<some imports>`? Also wonder why the comment suggests `double quote` imports

tests/mixed-source/ios/MixedSourceFramework/MixedSourceFramework.h
```diff
@@ -1,12 +1,2 @@
...
-
-//! Project version number for MixedSourceFramework.
-FOUNDATION_EXPORT double MixedSourceFrameworkVersionNumber;
-
-//! Project version string for MixedSourceFramework.
-FOUNDATION_EXPORT const unsigned char MixedSourceFrameworkVersionString[];
-
-// In this header, you should import all the public headers of your framework using statements like #import <MixedSourceFramework/PublicHeader.h>
+// Needed because the generated -Swift.h header is hard-coded to import this file, instead of the underlying headers / umbrella header for the module
+#import <MixedSourceFramework/MixedSourceFramework-umbrella.h>
```
gyfelton, line 2: there are cases when `SomeFramework.h` file has a list of includes, shouldn't we have a test case for this situation?

tests/mixed-source/ios/MixedSourceFramework/Logger/LoggerProtocol.h
```diff
@@ -0,0 +1,5 @@
+@import Foundation;
+
+@protocol LoggerProtocol <NSObject>
+
```
gyfelton, line 4: why not just copy `@objc public func swiftLog(_ message: String)` to this one so `SwiftLogger ` actually  implements this protocol?

rules/library.bzl
segiddins: `-I.` adds the current directory to the header search paths, and it's needed _after_ the `-I` flags pointing to the hmaps, otherwise relative paths inside hmaps don't resolve

rules/library.bzl
segiddins: When `framework` text is already included in the modulemap, it should be removed before going through this codepath

rules/library.bzl
segiddins: bad copy/paste, it should say angle brackets. ill fix

tests/mixed-source/ios/MixedSourceFramework/MixedSourceFramework.h
segiddins: Probably not needed, since I cant think of a functional difference?

rules/library.bzl
gyfelton: so to be clear `.` is for the current directory? Wonder if this is useful info to be included as a comment

tests/mixed-source/ios/MixedSourceFramework/Logger/LoggerProtocol.h
segiddins: Because it doesn't matter whats in the protocol, and `SwiftLogger` does indeed conform to this protocol (because there are no requirements). The conformance matters because it forces the generated swift header to have an import statement for the current module

tests/mixed-source/ios/MixedSourceFramework/MixedSourceFramework.h
gyfelton: if no functional difference why need to change here then?

segiddins: easier than manually adding each header when a new one gets added

rules/library.bzl
justinseanmartin: Not sure if you've already done this, but if not, I'd recommend looking at `swiftc --help` and `man clang` for what these args are for. There is also documentation up at https://clang.llvm.org/docs/

rayray: @segiddins What is the "current directory" that -I. refers to? Is it the directory of the BUILD file that calls this macro or another?

PR #63, opened 2020-05-19
Remove dependency of `jq` util

gyfelton:
Instead use a built-in json-util script (from https://github.com/dominictarr/JSON.sh) to parse the json and output list of file paths for `touch`
Existing tests covers this code (aka test will fail if this does not work) so no new test is added for now.


tools/xcodeproj-shims/swiftc-stub.sh
```diff
@@ -7,8 +7,9 @@ if [[ $# -eq 1 && $1 == "-v" ]]; then
     exec swiftc -v
 fi
 
+# Source of json-util: https://github.com/dominictarr/JSON.sh
 write_output_files() {
-    cat $1 | jq 'map(map(.)) | flatten | unique | .[]' -r | xargs touch
+    cat $1 | $(dirname "$0")/json-util -b -n  | sed 's/\[.*\]//' | xargs touch
```
gyfelton, line 12: example of one line of output from json-util -b -n is:

[".../register/Frameworks/Chartography/DemoApp/ConfigurableViewController.swift","dependencies"] ".../DerivedData/Chartography-Project-dljjinukhucgryfdyxkxdhbgnsrk/Build/Intermediates.noindex/Chartography-Project.build/Debug-iphonesimulator/Chartography-DemoApp.build/Objects-normal/x86_64/ConfigurableViewController.d"

where -b denotes get to the leaf nodes and remove empty entries, -n means remove anything that has no "head"

With this output, the next step is to use sed 's/\[.*\]//' to remove the first portion of the output (the path to the value) and so we end up a list of values, which is what jq 'map(map(.)) | flatten | unique | .[]' -r is doing

tools/xcodeproj-shims/swiftc-stub.sh
```diff
@@ -7,8 +7,9 @@ if [[ $# -eq 1 && $1 == "-v" ]]; then
     exec swiftc -v
 fi
 
+# Source of json-util: https://github.com/dominictarr/JSON.sh
```
segiddins, line 10: an alternative to this is to use python inline:

(something like `python -c` with)
```python
import json

files = set()
for derived_files in json.load(sys.stdin).values(): 
    files.extend(derived_files.values())

for file in files:
    open(file, 'a').close() 
```

or honestly just to rewrite the entire stub to use a language that isn't bash

segiddins: see, for example, https://github.com/lyft/envoy-mobile/pull/615/files#diff-8c6a45719413cc768d2e976b0c540484

gyfelton: my first thought is on python then I was like "wait does it then depend on python?" But now that the entire bazel heavily depend on python/starlark. Guess this is after all the viable solution

segiddins: Python is installed by default on macOS, same as bash/sh. Another alternative would be to write a wrapper that gets compiled into a native executable

amberdixon:
Agree with Sam that it would be better to use python for this.

tools/xcodeproj-shims/BUILD.bazel
```diff
@@ -18,6 +18,12 @@ sh_binary(
     visibility = STUB_VISIBILITY,
 )
 
+sh_binary(
```
segiddins, line 21: this isn't a shell script any more

tools/xcodeproj-shims/json-util.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/python
```
segiddins, line 1: `/usr/bin/env python`, and change the extension

tools/xcodeproj-shims/json-util.sh
```diff
@@ -0,0 +1,13 @@
...
+import sys, json; 
+j = json.load(sys.stdin)
+queue = [j]
+
+# Print all leaf nodes of a given json via stdin
```
segiddins, line 6: should we have a more desciptive file name, since this isn't a generic json utility and rather only does one specific thing?

tools/xcodeproj-shims/json-util.sh
```diff
@@ -0,0 +1,13 @@
...
+while len(queue) > 0:
+    entry = queue.pop(0)
+    if isinstance(entry, dict):
+      for key, value in entry.items():
+        queue.append(value)
+    else:
+      print entry
```
segiddins, line 13: `print(entry)` for python 3 compat

tools/xcodeproj-shims/json-util.sh
gyfelton: Agree, replaced all references of file names to `print-json-leaf-nodes`

tools/xcodeproj-shims/BUILD.bazel
gyfelton: right! changed

gyfelton:
Another look @segiddins ? Have to rename `xcodeproj-shim` to`xcodeproj_shim` as python don't want `-` in the path, let me know if there is an alternative.

segiddins:
One nit, it'd be nice to rebase to get rid of the commits that just say "fixup"

gyfelton:
@segiddins i always do squash and merge to get rid of these msgs

PR #64, opened 2020-05-27
The promised leftovers from the previous PR

acecilia:
Let me know if I missed something or you prefer other names

PR #58, opened 2020-05-07
Fix "module not found" when using prebuilt binaries + added prebuilt_framework_importer.bzl

acecilia:
Fixes https://github.com/bazel-ios/rules_ios/issues/55

* Fixed dependency on prebuilt frameworks for the problem described in https://github.com/bazel-ios/rules_ios/issues/55: added `apple_static_framework_import_patched` and `apple_dynamic_framework_import_patched`
* Added more examples

We may want to switch the implementation of `library.bzl` from `apple_**_framework_import` to `apple_**_framework_import_patched` in case this PR moves forward

Also adds `prebuilt_framework_importer.bzl`, which can be used to create binaries from your `carthage` and `cocoapods` dependencies, and integrate them with bazel (so they can be used and cached)

PR #65, opened 2020-05-27
Revert "Fix "module not found" when using prebuilt binaries + added prebuilt_framework_importer.bzl"

ob:
Reverts bazel-ios/rules_ios#58

See comments on #64


acecilia:
There is no need for this massive revert, I think you just need to comment out the contents of `https://github.com/bazel-ios/rules_ios/blob/master/tests/ios/frameworks/mixed-source/sources-with-prebuilt-binaries/BUILD.bazel`

ob:
Heh, fair point. I'll discard this PR and put together a more focused one.

PR #67, opened 2020-05-27
Move the repository rules to their own directory

ob:
Also move the tests to their own sub-repo to avoid a hit in load
times in the normal case when working on the rules.

CI still runs all the tests, including repository rules.


ob:
CC @acecilia

.github/workflows/tests.yml
```diff
@@ -19,6 +19,7 @@ jobs:
           bazelisk build --local_test_jobs=1 -- //... -//tests/ios/frameworks/mixed-source/sources-with-prebuilt-binaries/...
```
acecilia, line 19: This path `tests/ios/frameworks/mixed-source/sources-with-prebuilt-binaries/` needs to be updated

acecilia:
Looking great I think, very elegant

.github/workflows/tests.yml
```diff
@@ -19,6 +19,7 @@ jobs:
...
           bazelisk test --local_test_jobs=1  //tests/macos/...
           bazelisk test --local_test_jobs=1 --apple_platform_type=ios //tests/ios/...
```
segiddins, line 21:
```suggestion
          bazelisk test --local_test_jobs=1 -- //... -//tests/ios/... -//tests/macos/...
          bazelisk test --local_test_jobs=1 --apple_platform_type=macos -- //tests/macos/...
          bazelisk test --local_test_jobs=1 --apple_platform_type=ios -- //tests/ios/...
```

This way tests outside of the macos/ios dirs get run (such as the tests for the hmap tool)

ob:
Well, my idea didn't work so I guess for the time being we are still paying that loading time. I'll see if I can come up with something else.

PR #70, opened 2020-06-01
Adds a build wrapper script for Bazel build processing

ndizazzo:
## What Changed & Why
This PR changes the xcodeproj Bazel file to include an additional build wrapper that allows us to call an output processor.

The output processor is needed because:
1. Bazel outputs all build progress information to `STDERR`, causing Xcode to think that there was an error building the project.
2. Bazel outputs some of its build progress in a way that Xcode interprets as build error
3. The issue navigator in Xcode doesn't work with the way Bazel outputs the paths for files when building using Bazel. We can post-process each line of output to prepend the full path to `BAZEL_WORKSPACE_ROOT` to correct this.

Finally, developers are able to click an error or warning in the issue navigator and be directed to the corresponding warning / error.

## Screenshot
Screen Shot 2020-06-01 at 12 21 38 PM

Screen Shot 2020-06-01 at 12 21 59 PM

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -348,7 +348,7 @@ xcodeproj = rule(
         "_xcodeproj_installer_template": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:xcodeproj-installer.sh"), allow_single_file = ["sh"]),
         "_infoplist_stub": attr.label(executable = False, default = Label("//rules/test_host_app:Info.plist"), allow_single_file = ["plist"]),
         "_workspace_xcsettings": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:WorkspaceSettings.xcsettings"), allow_single_file = ["xcsettings"]),
-        "output_processor": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:output-processor.rb"), cfg = "host", allow_single_file = ["rb"]),
+        "output_processor": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:output-processor.rb"), cfg = "host", allow_single_file = True),
         "_xcodegen": attr.label(executable = True, default = Label("@com_github_yonaskolb_xcodegen//:xcodegen"), cfg = "host"),
         "clang_stub": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:clang-stub"), cfg = "host"),
         "ld_stub": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:ld-stub"), cfg = "host"),
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -166,6 +166,8 @@ def _xcodeproj_impl(ctx):
         "groupSortPosition": "none",
     }
     proj_settings = {
+        "BAZEL_BUILD_EXEC": "$BAZEL_STUBS_DIR/build-wrapper",
+        "BAZEL_OUTPUT_PROCESSOR": "$BAZEL_STUBS_DIR/output-processor.rb",
         "BAZEL_PATH": ctx.attr.bazel_path,
         "BAZEL_WORKSPACE_ROOT": "$SRCROOT/%s" % script_dot_dots,
         "BAZEL_STUBS_DIR": "$PROJECT_FILE_PATH/bazelstubs",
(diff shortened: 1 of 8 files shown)


rules/xcodeproj.bzl
```diff
@@ -342,12 +348,14 @@ xcodeproj = rule(
         "_xcodeproj_installer_template": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:xcodeproj-installer.sh"), allow_single_file = ["sh"]),
         "_infoplist_stub": attr.label(executable = False, default = Label("//rules/test_host_app:Info.plist"), allow_single_file = ["plist"]),
         "_workspace_xcsettings": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:WorkspaceSettings.xcsettings"), allow_single_file = ["xcsettings"]),
+        "output_processor": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:output-processor.rb"), cfg = "host", allow_single_file = ["rb"]),
```
segiddins, line 351: if we're going to allow people to override the default for this attribute (which is what's happening here, since the name isn't `_` prefixed), there's no need to limit to only `.rb` files

tools/xcodeproj_shims/output-processor.rb
```diff
@@ -0,0 +1,77 @@
...
+    rule = new(line)
+
+    rule.match_data.nil? ? nil : rule
+  end
+end
+
+# Documentation
+class SnapshotLine < RegexMatchLine
+  def initialize(line)
+    @regex = /^DEBUG: .+ Snapshot test (.+) (\(.+\)) does not specify it uses an app host$/
```
segiddins, line 42: this is square-specific, this i think should just match _all_ lines that begin with `DEBUG` and are followed by a path to a starlark file

tools/xcodeproj_shims/output-processor.rb
```diff
@@ -37,21 +37,21 @@ def self.create_from(line)
 end
 
 # Documentation
-class SnapshotLine < RegexMatchLine
+class StarlarkLine < RegexMatchLine
   def initialize(line)
-    @regex = /^DEBUG: .+ Snapshot test (.+) (\(.+\)) does not specify it uses an app host$/
+    @regex = /^DEBUG: (.+.bzl)(:\d+:\d+:) (.+)$/
```
segiddins, line 42: `(.+(\.bzl|\.bazel)?)` ?

tools/xcodeproj_shims/output-processor.rb
```diff
@@ -37,21 +37,21 @@ def self.create_from(line)
...
 
     super(line)
 
     return unless (@match_data = line.match(@regex))
 
-    test_name, test_path = @match_data.captures
-    @text = "warning: Snapshot test #{test_name} #{test_path} does not specify it uses an app host"
+    starlark_file, file_line, message = @match_data.captures
```
segiddins, line 48: should we be pointing to the file with the issue here?

rules/xcodeproj.bzl
ndizazzo: Going with the `allow_single_file = True` approach here instead

tools/xcodeproj_shims/output-processor.rb
ndizazzo: We could include offending file path - but in my testing Xcode seems to pick up any output matching `<path>:<line>:<character> ` and assumes it's an error unless `info:` or `warning:` is specified

segiddins: but i think we are specifying `warning` ?

ndizazzo: Yup, I ended up appending the path to the .bzl file for developers to dig further into if needed... What I was trying to get at was if we do this:

`warning: Some warning message (/some/path/to/a/file:20:10) and some additional warning data`

Xcode **still** flags that an error (as I observed), as lines formatted like that would show up as errors highlighted in red in the build log and in the issue navigator. Removing the line and character marker seems to correct this.

ndizazzo: I think this had to do with the STDOUT / STDERR though. I couldn't get repeatable behaviour before I made the the change to only route it to STDOUT.

ndizazzo:
Thanks for the pointer on `STDOUT.sync = true`. I was mulling over how to get this working the same way as it was before I moved the changes into `rules_ios` wasn't aware that flag existed.

PR #71, opened 2020-06-03
Remove test_host from list of deps to avoid including extra sources during compile sources phase

ndizazzo:
This PR undoes adding the `test_host` to the list of dependencies in order to prevent it's sources from being included in other target's compile sources build phase.

After discussion with @segiddins and @amberdixon, we considered preventing `_SrcInfo` from being appended to the list of providers, but after testing, that still didn't work.

See PR #46 for original context.

**Before**
Screen Shot 2020-06-03 at 4 27 36 PM

**Now**
Screen Shot 2020-06-03 at 4 38 38 PM


ndizazzo:
tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/project.pbxproj was regenerated with `bazel run tests/macos/xcodeproj:Test-Target-With-Test-Host-Project`

PR #72, opened 2020-06-03
Adds deployment target configs based on provided platform_type & minimum_os_version

ndizazzo:
This PR adds the suggested patch from @segiddins that allows platform and minimum deployment target to be passed through to target configuration option settings in Xcode.

If the platform is `iOS`, it will add `IPHONEOS_DEPLOYMENT_TARGET` with the corresponding minimum version.

If the platform is `macOS`, it will add `MACOSX_DEPLOYMENT_TARGET` with the corresponding minimum version (and set the `SDKROOT` to `macosx` if needed).

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -10,6 +10,17 @@ def _get_attr_values_for_name(deps, provider, field):
 
 _TargetInfo = provider()
 _SrcsInfo = provider()
+_PLATFORM_MAPPING = {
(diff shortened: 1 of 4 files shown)


ndizazzo:
@segiddins f6aec9a updates the min os version to 11 in an attempt to see if that corrects the build failures when the deployment target is set (and it's < iOS 11).

Any idea what additional configs we might need to specify to get it working again with iOS 10?

PR #73, opened 2020-06-09
Build frameworks/resource_bundles behind a transition

segiddins:
This way, their artifacts can be shared across multiple apps/test targets, regardless of the deployment targets used in the final packaged product


rules/library.bzl
```diff
@@ -318,9 +308,12 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     weak_sdk_frameworks = kwargs.pop("weak_sdk_frameworks", [])
     sdk_includes = kwargs.pop("sdk_includes", [])
     pch = kwargs.pop("pch", "@build_bazel_rules_ios//rules/library:common.pch")
-    deps = kwargs.pop("deps", [])
+    deps = [d for d in kwargs.pop("deps", [])]
     data = kwargs.pop("data", [])
     tags = kwargs.pop("tags", [])
+    platforms = kwargs.pop("platforms", {
+        "ios": "10.0",  # TODO: remove default
```
segiddins, line 315: Remove default

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -0,0 +1,207 @@
...
+    for k in dir(ctx.attr):
+        if k in ("to_json", "to_proto"):
+            continue
+        attr_dict[k] = getattr(ctx.attr, k)
+
+    current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
+    attr_dict["platform_type"] = str(current_apple_platform.platform.platform_type)
+    attr_dict["minimum_os_version"] = str(current_apple_platform.target_os_version)
+
+    # TODO: only do this dance if platform is not macos
```
segiddins, line 38: remove TODO, this is now done

tests/framework/platforms/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load(":versions.bzl", "make_applications", "make_frameworks", "make_frameworks_with_dependencies")
+
+# [
```
segiddins, line 3: remove commented code

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
segiddins, line 1: Deleted since I don't think these files actually get generated anywhere (except maybe by Xcode)

tools/xcodeproj_shims/xcodeproj-installer.sh
```diff
@@ -6,10 +6,21 @@ readonly project_path="${PWD}/$(project_short_path)"
...
 
-readonly stubs_dir="${dest}/bazelstubs"
+readonly installer="$(installer_short_path)"
+
+rm -fr "${tmp_dest}"
+mkdir -p "$(dirname $tmp_dest)"
+cp -r "${project_path}" "$tmp_dest"
+chmod -R +w "${tmp_dest}"
+
+readonly stubs_dir="${tmp_dest}/bazelstubs"
```
segiddins, line 16: installing the stubs into the tmp dir means they will get chmod'd, allowing them to be overwritten by subsequent invocations

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
segiddins: cc @amberdixon

tools/xcodeproj_shims/xcodeproj-installer.sh
segiddins: cc @amberdixon this is the change I showed you on Tuesday

rules/framework.bzl
```diff
@@ -12,15 +22,15 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
...
     library = apple_library(name = name, **kwargs)
     apple_framework_packaging(
         name = name,
         framework_name = library.namespace,
         transitive_deps = library.transitive_deps,
         deps = library.lib_names,
-        visibility = kwargs.get("visibility", None),
-        tags = kwargs.get("tags", None),
+        platforms = library.platforms,
+        **framework_packaging_kwargs
```
ob, line 33: Heh, cute.

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -0,0 +1,235 @@
...
+        top_level_attrs = ["resources"],
+    ), fake_ctx)
+
+    # Process the plist ourselves. This is required because
+    # include_executable_name defaults to True, which results in iTC rejecting
+    # our binary.
+    output_plist = ctx.actions.declare_file(
+        paths.join("%s-intermediates" % ctx.label.name, "Info.plist"),
+    )
+    bundle_id = ctx.attr.bundle_id or "com.cocoapods." + bundle_name
```
ob, line 66: Would it make sense to make the `bundle_id` required instead of providing a default?

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -0,0 +1,235 @@
...
+                Label("@build_bazel_rules_ios//rules/library:resource_bundle.plist"),
+            ],
+        ),
+        bundle_name = attr.string(
+            mandatory = False,
+            doc = "The name of the resource bundle. Defaults to the target name.",
+        ),
+        bundle_id = attr.string(
+            mandatory = False,
+            doc = "The bundle identifier of the resource bundle.",
```
ob, line 188: Or if we are providing a default (which I suggest we don't) it should be documented here.

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -0,0 +1,235 @@
...
+load("@build_bazel_rules_apple//apple/internal:partials.bzl", "partials")
+load("@build_bazel_rules_apple//apple/internal:resource_actions.bzl", "resource_actions")
+load("@build_bazel_rules_apple//apple/internal:rule_factory.bzl", "rule_factory")
+load("//rules:transition_support.bzl", "transition_support")
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleResourceBundleInfo", "AppleResourceInfo")
+
+_FAKE_BUNDLE_PRODUCT_TYPE_BY_PLATFORM_TYPE = {
+    "ios": apple_product_type.application,
+    "tvos": apple_product_type.application,
+    "watchos": apple_product_type.watch2_application,
```
ob, line 22: Should we add `macos` here since _why not_?

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
ob: +1

rules/precompiled_apple_resource_bundle.bzl
segiddins: Maybe? We could push the default out to the `apple_library` macro if you think that's better, but we do need a default at some layer

rules/precompiled_apple_resource_bundle.bzl
segiddins: macos is tested, it just doesnt need a fake bundle product type since the rule descriptors have one for a macos resource bundle

rules/library.bzl
```diff
@@ -157,26 +156,17 @@ FOUNDATION_EXPORT const unsigned char {module_name}VersionString[];
...
-                "PRODUCT_NAME": bundle_name,
-            },
-            tags = _MANUAL,
-        )
-        apple_resource_bundle(
+        precompiled_apple_resource_bundle(
             name = target_name,
             bundle_name = bundle_name,
             resources = [
                 library_tools["wrap_resources_in_filegroup"](name = target_name + "_resources", srcs = resource_bundles[bundle_name]),
```
segiddins, line 167: In a follow-up, it would be nice to get rid of this extra target as well

segiddins:
> Do we have concerns about blowing up the build graph by using transitions?

I _think_ that this will actually _shrink_ the build graph, because it's an incoming edge transition rather than an outgoing one. Before, we could build a framework (and therefore its library targets) N times, depending on what app/test binary was linking it in. Now, it's guaranteed to only be built once per platform when an explicit list of platform/deployment target pairs is passed in.

rules/framework.bzl
```diff
@@ -284,10 +300,25 @@ def _apple_framework_packaging_impl(ctx):
...
+            binary = binary_out,
+            bundle_id = bundle_id,
+            bundle_name = framework_name,
+            bundle_extension = bundle_extension,
+            entitlements = None,
+            infoplist = infoplist,
+            minimum_os_version = str(current_apple_platform.target_os_version),
+            platform_type = str(current_apple_platform.platform.platform_type),
+            product_type = ctx.attr._product_type,
+            uses_swift = swiftmodule_out != None,
```
amberdixon, line 315: nit, can you add parens here to keep the order of operations obvious?
suggestion:
uses_swift = (swiftmodule_out != None),

amberdixon: I hope the RHS doesn't turn into an array/tuple if you do that!

tests/framework/platforms/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
...
+    srcs = ["platform.m"],
+    resources_to_bundle = glob(["Resources/**/*"]),
+)
+
+make_frameworks_with_dependencies(
+    srcs = ["platform.m"],
+    resources_to_bundle = glob(["Resources/**/*"]),
+)
+
+make_applications(
```
amberdixon, line 13: Did you confirm that this application would fail to run, if this transition was not applied?

segiddins: Yes

Issue #74, opened 2020-06-10
Unknown class in Interface Builder file


barbasevich:
I have a simple project [TestXIB](https://github.com/barbasevich/bazel-test-xib) with `CustomView.swift` and `CustomView.xib`.
xib file has `customClass="CustomView" customModule="TestXIB" customModuleProvider="target"` settings.

When I'm running unit tests and trying to instantiate `CustomView` from xib I get an error:
`Unknown class _TtC39TestXIBTests.__internal__.__test_bundle10CustomView in Interface Builder file.`

If I either switch from `apple_framework` to `swift_library` or remove `customModuleProvider="target"` from xib all works fine.

Any ideas on how to solve the issue using `apple_framework` without xib modification.?

segiddins:
Ah yeah that could be it, good hunch

PR #75, opened 2020-06-11
Pass data to swift_library when relevant

barbasevich:
Pass resources to swift_library when there are swift sources.
Otherwise, pass to objc_library.

[#74] Fixes Unknown class in Interface Builder file issue.


barbasevich:
`rules_apple:74eca5857a136b9f1e2020886be76b791eb08231` could not be parsed.

Seems the ref should be updated.

segiddins:
It would be nice to add a xib to one of the tests, so we can test `customModuleProvider="target"` indeed works

barbasevich:
@segiddins Agree, added.

segiddins:
Thanks!

PR #77, opened 2020-06-16
Generated xcodeproj improvements

segiddins:
* Mark xcodeproj directories as generated
  
  This way, github will collapse diffs & have more accurate repo stats

* Write all xcodeproj files into the tmp_dest
  
  This way, everything will get chmod’d
  
  This fixes back-to-back runs of the installer failing due to permission issues

* Add a test that verifies that projects can regenerate without error

* Add support for specifying additional (non-srcs) files to be included in the project
  
  This would allow a project to, for example, include a README as a file reference, even though it isnt a src belonging to any particular target

* Download GitHub repos via http_archive

* Add IDEWorkspaceChecks.plist to generated projects
  
  So opening the project in Xcode doesn’t add a new file

* Include BUILD files in generated Xcode projects

* Explicitly opt out of xcodeproj settings presets
  
  It avoids the warning of the presets not being found

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.xcodeproj/ linguist-generated=true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,11 @@ jobs:
       - name: Select Xcode 11.2
         run: sudo xcode-select -s /Applications/Xcode_11.2.app
       - name: Generate Xcode projects and compare against existing record
-        run: (bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run) && git diff --exit-code tests/macos/xcodeproj
+        run: |
+          set -euo pipefail
+          bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run
+          bazel query 'attr(executable, 1, kind(genrule, tests/macos/xcodeproj/...))' | xargs -n 1 bazelisk run
+          git diff --exit-code tests/macos/xcodeproj
       - name: Run Xcode builds
         run: ./tests/macos/xcodeproj/build.sh
       - name: Run Unit tests
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -36,6 +36,13 @@ def _is_current_project_file(f):
 def _is_current_project_path(path):
     return not path.startswith("external/")
 
+def _srcs_info_build_files(ctx):
+    path = ctx.build_file_path
+    if not _is_current_project_path(path):
+        return []
+
+    return [path]
+
 def _xcodeproj_aspect_impl(target, ctx):
     providers = []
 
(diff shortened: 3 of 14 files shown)


rules/xcodeproj.bzl
```diff
@@ -282,18 +303,25 @@ $BAZEL_INSTALLER
...
         targets = xcodeproj_targets_by_name,
         schemes = xcodeproj_schemes_by_name,
+        fileGroups = project_file_groups,
     )
 
     ctx.actions.write(xcodegen_jsonfile, xcodeproj_info.to_json())
-
     ctx.actions.run(
         executable = ctx.executable._xcodegen,
+        # executable = "/Users/segiddins/Library/Developer/Xcode/DerivedData/XcodeGen-cilmddfxvdifhqgfmseshsaujtgb/Build/Products/Debug/XcodeGen",
```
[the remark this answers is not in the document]
segiddins: whoops!

.github/workflows/tests.yml
```diff
@@ -45,7 +45,11 @@ jobs:
       - name: Select Xcode 11.2
         run: sudo xcode-select -s /Applications/Xcode_11.2.app
       - name: Generate Xcode projects and compare against existing record
-        run: (bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run) && git diff --exit-code tests/macos/xcodeproj
+        run: |
```
ob, line 48: Nice, more readable! Can we count on all the lines being run in the same shell? Otherwise the `set -euo pipefail` might not be taking effect...

rules/xcodeproj.bzl
```diff
@@ -163,6 +176,7 @@ def _xcodeproj_impl(ctx):
         "createIntermediateGroups": True,
         "defaultConfig": "Debug",
         "groupSortPosition": "none",
+        "settingPresets": "none",
```
amberdixon, line 179: fyi to reviewers, this is described here: https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#options

tests/macos/xcodeproj/BUILD.bazel
```diff
@@ -82,3 +82,23 @@ xcodeproj(
...
+    name = "Test-Project-Regeneration",
+    testonly = True,
+    outs = ["Test-Project-Regeneration.sh"],
+    cmd = """
+cat <<'EOS' > $@
+#!/bin/sh
+set -euxo pipefail
+rm -fr {package_name}/{target_name}.xcodeproj
+bazel run {package_name}:{target_name}
+bazel run {package_name}:{target_name}
```
ob, line 96: Is running it twice intentional?

segiddins: yeah, it's testing re-generation (previously that was failing)

.github/workflows/tests.yml
segiddins: The whole `run` string gets executed at once by the shell

.gitattributes
```diff
@@ -0,0 +1 @@
+*.xcodeproj/ linguist-generated=true
```
amberdixon, line 1: Should we be seeing diffs for the `pbxproj` files in this PR?

Should this say **/*.xcodeproj/** instead?

PR #79, opened 2020-06-24
Separate objective-c indexstore units and records by objc_library

amberdixon:
This behavior is similar to swift_library: https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/derived_files.bzl#L66 (See its invocation of `actions.declare_directory` from `_indexstore_directory`.)

We observed an issue where it took a long time for index-import to remap bazel-generated indexstores. The slowdown was actually a result of the fact that all objective-c indexstores were being placed in the same .indexstore directory. When we broke them apart by objective-c library, things got considerably faster. Before these changes, it took 3 minutes to remap for a larger project, but now it takes 3 seconds.

There is an alternative approach we can take. This approach has been taken by Brentley here: https://github.com/lyft/index-import/pull/46 This PR is open right now, but once it lands, we can consider it, if we still see performance issues.
1. Group index unit/record files into the same indexstore directory.
2. Invoke index-import and have it run on parallel across units themselves, rather than indexstore directories.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -563,7 +563,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
 
     objc_copts.append("-I.")
 
-    objc_copts.extend(("-index-store-path", "$(GENDIR)/rules_ios_apple_library_objc.indexstore"))
+    objc_copts.extend(("-index-store-path", "$(GENDIR)/rules_ios_objc_library_{}.indexstore".format(objc_libname)))
     objc_library(
         name = objc_libname,
         srcs = objc_sources + objc_private_hdrs + objc_non_exported_hdrs,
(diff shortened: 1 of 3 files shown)


rules/library.bzl
```diff
@@ -563,7 +563,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
 
     objc_copts.append("-I.")
 
-    objc_copts.extend(("-index-store-path", "$(GENDIR)/rules_ios_apple_library_objc.indexstore"))
+    objc_copts.extend(("-index-store-path", "$(GENDIR)/rules_ios_objc_library_{}.indexstore".format(objc_libname)))
```
segiddins, line 566: this will resolve to the same name for targets with the same name in different packages -- is that going to be an issue?

amberdixon: oh yeah :( Well, that wouldn't create any problems (after all, before we were grouping all the units in the same directory), but it is a bit confusing to someone debugging why certain unit files don't show up where we want them to.

Let me see if I can add package in here.

PR #80, opened 2020-06-24
Collect diagnostics about indexing so we can debug problems more easily

amberdixon:
I considered using something like this build setting: https://pewpewthespells.com/blog/buildsettings.html#derived_file_dir

But I changed my mind, because I thought it might be easier to find the `bazel-xcode-diagnostics` directory if it was sitting closer to the root. Let me know if there are some other reasons I should consider using a deeper subdirectory of the derived data for the project.

### DIFF
--- a/tools/xcodeproj_shims/installers/indexstores.sh
+++ b/tools/xcodeproj_shims/installers/indexstores.sh
@@ -27,3 +27,5 @@ else
 fi
 
 echo "Finish remapping index files at `date`"
+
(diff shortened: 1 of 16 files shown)


tools/xcodeproj_shims/installers/indexstores.sh
```diff
@@ -27,3 +27,5 @@ else
 fi
 
 echo "Finish remapping index files at `date`"
+
+ls -ltR $BUILD_DIR/../../Index/DataStore/ > $BAZEL_DIAGNOSTICS_DIR/indexstores-contents-$DATE_SUFFIX.log
```
segiddins, line 31: trailing newline

PR #87, opened 2020-07-08
Unique the list of indexstores and move all of them, not just the first

amberdixon:
The bazel build process generates indexstores, if the necessary index-store-path options are enabled.

The xcodeproject invokes bazel build. Then it moves the generated indexstores to the derived data directory. This fixes a bug where only the FIRST indexstore was getting moved and not the others!!!!

### DIFF
--- a/tools/xcodeproj_shims/installers/indexstores.sh
+++ b/tools/xcodeproj_shims/installers/indexstores.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 
 echo "Start remapping index files at `date`"
 
-FOUND_INDEXSTORES=`pcregrep -o1 'command_line: "(.*\.indexstore)' $BAZEL_BUILD_EVENT_TEXT_FILENAME || true`
+FOUND_INDEXSTORES=`pcregrep -o1 'command_line: "(.*\.indexstore)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | uniq`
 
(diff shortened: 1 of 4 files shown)


tools/xcodeproj_shims/installers/indexstores.sh
```diff
@@ -6,7 +6,7 @@ set -euo pipefail
 
 echo "Start remapping index files at `date`"
 
-FOUND_INDEXSTORES=`pcregrep -o1 'command_line: "(.*\.indexstore)' $BAZEL_BUILD_EVENT_TEXT_FILENAME || true`
+FOUND_INDEXSTORES=`pcregrep -o1 'command_line: "(.*\.indexstore)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | uniq`
```
segiddins, line 9: will this now fail if no commandline matches the pattern?

amberdixon: i tested this locally, and it seems to just result in an empty array and an exit status code of 0:

```
~/Development/register master * fi6=`pcregrep -o1 'command_line: "(.*\.indexstore)' /tmp/foo | uniq`; echo $?
0
~/Development/register master * cat /tmp/foo
hi
```

segiddins:
Could you please add a short description to the PR saying what was broken & what this change is fixing?

PR #89, opened 2020-07-14
Fix a memory error in hmaptool

ob:
Don't assume doubling the buffer once makes it big enough

### DIFF
--- a/rules/hmap/hmap.c
+++ b/rules/hmap/hmap.c
@@ -411,9 +411,10 @@ static char* getStringFromTable(HeaderMap* hmap, unsigned int tblIndex) {
 static inline unsigned addStringToTable(HeaderMap* hmap, char* str) {
     unsigned idx = hmap->stringsTableNextEntry;
     unsigned len = strlen(str);
-    if (idx + len >= hmap->size) {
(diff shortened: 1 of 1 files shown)


rules/hmap/hmap.c
```diff
@@ -411,7 +411,7 @@ static char* getStringFromTable(HeaderMap* hmap, unsigned int tblIndex) {
 static inline unsigned addStringToTable(HeaderMap* hmap, char* str) {
     unsigned idx = hmap->stringsTableNextEntry;
     unsigned len = strlen(str);
-    if (idx + len >= hmap->size) {
+    while (idx + len >= hmap->size) {
```
segiddins, line 414: ooooof.

alternatively, instead of reallocating, it _might_ work to find the next highest power of two that's `>= idx + len` ? That way we theoretically could avoid a second `realloc` inside this function call?

Also, looking below, we `memcp` `len + 1` bytes, so should this be `idx + len + 1 >=` ?

ob: Yeah, good point... I don't think the realloc is a big deal but the code would be cleaner if we change it.

PR #90, opened 2020-07-15
Set xcode target build settings, so xcode indexer will work

amberdixon:
If xcode indexer fails, autocomplete operations will not work.


rules/xcodeproj.bzl
```diff
@@ -115,18 +118,28 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+        other_cflags = []
         for attr in _dir(ctx.rule.files):
             if attr == "srcs":
                 srcs += getattr(ctx.rule.files, attr, [])
             else:
                 asset_srcs += getattr(ctx.rule.files, attr, [])
         srcs = [f for f in srcs if _is_current_project_file(f)]
         asset_srcs = [f for f in asset_srcs if _is_current_project_file(f)]
+        if CcInfo in target:
+          for fi in target[CcInfo].compilation_context.framework_includes.to_list():
```
segiddins, line 130: instead of getting this as a list, we should directly propogate this depset -- alternatively, we might want to just propogate a raw `compilation_context` inside the `_SrcsInfo`, so we can use the `cc_common` aparatus for merging the values. then, in the implementation of the xcodeproj rule, we can expand out the depsets as needed, and turn them into `OTHER_CFLAGS` there (or alternatively use the native xcconfigs, such as `FRAMEWORK_SEARCH_PATHS` and `GCC_PREPROCESSOR_DEFINITIONS`)

rules/xcodeproj.bzl
```diff
@@ -115,18 +118,28 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+            if fi[0] != '/':
+              fi = "$BAZEL_WORKSPACE_ROOT/%s" % fi
+            other_cflags.append("-F%s" % fi)
+          other_cflags.extend(["-D%s" % di for di in target[CcInfo].compilation_context.defines.to_list()])
```
segiddins, line 134: do we also want to pull out defines from the `SwiftInfo` provider to pass to to `SWIFT_ACTIVE_COMPILATION_CONDITIONS` ? What about paths to swiftmodules for non-framework deps (such as a test that will try to import its test host)?

rules/xcodeproj.bzl
```diff
@@ -115,18 +118,28 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+          other_cflags.append("-fobjc-arc") # TODO: only apply to arc sources
```
segiddins, line 135: I know you have a TODO here, but we'll likely want to mimic what cocoapods does and default to `CLANG_ENABLE_OBJC_ARC=YES`,  and then set `-fno-objc-arc` on the individual no-arc srcs

rules/xcodeproj.bzl
```diff
@@ -115,18 +118,28 @@ def _xcodeproj_aspect_impl(target, ctx):
...
                 srcs += getattr(ctx.rule.files, attr, [])
             else:
                 asset_srcs += getattr(ctx.rule.files, attr, [])
         srcs = [f for f in srcs if _is_current_project_file(f)]
         asset_srcs = [f for f in asset_srcs if _is_current_project_file(f)]
+        if CcInfo in target:
+          for fi in target[CcInfo].compilation_context.framework_includes.to_list():
+            if fi[0] != '/':
+              fi = "$BAZEL_WORKSPACE_ROOT/%s" % fi
+            other_cflags.append("-F%s" % fi)
```
segiddins, line 133: you might want to quote these values, in case they contain spaces

rules/xcodeproj.bzl
```diff
@@ -117,29 +121,38 @@ def _xcodeproj_aspect_impl(target, ctx):
...
         if CcInfo in target:
-          for fi in target[CcInfo].compilation_context.framework_includes.to_list():
-            if fi[0] != '/':
-              fi = "$BAZEL_WORKSPACE_ROOT/%s" % fi
-            other_cflags.append("-F%s" % fi)
-          other_cflags.extend(["-D%s" % di for di in target[CcInfo].compilation_context.defines.to_list()])
-          other_cflags.append("-fobjc-arc") # TODO: only apply to arc sources
-          print("got other cflags %s" % other_cflags)
-
+          framework_includes = target[CcInfo].compilation_context.framework_includes.to_list()
```
segiddins, line 147: we shouldnt be calling `to_list` here, flattening depsets multiple times has a significant perf impact https://docs.bazel.build/versions/master/skylark/performance.html#avoid-calling-depsetto_list

rules/xcodeproj.bzl
```diff
@@ -248,8 +267,11 @@ def _xcodeproj_impl(ctx):
             "BAZEL_BIN_SUBDIR": target_info.bazel_bin_subdir,
             "MACH_O_TYPE": target_macho_type,
             "CLANG_ENABLE_MODULES": "YES",
+            "CLANG_ENABLE_OBJC_ARC": "YES",
         }
-        target_settings["OTHER_CFLAGS"] = " ".join(target_info.other_cflags.to_list())
+        # TODO
+        target_settings["FRAMEWORK_SEARCH_PATHS"] = " ".join(target_info.framework_includes.to_list())
```
segiddins, line 273: do we need to quote the values in these two settings?

rules/xcodeproj.bzl
```diff
@@ -217,6 +235,12 @@ def _xcodeproj_impl(ctx):
             "group": paths.dirname(s.short_path),
             "optional": True,
         } for s in target_info.srcs.to_list()]
+        compiled_non_arc_sources = [{
+            "path": paths.join(src_dot_dots, s.short_path),
+            "group": paths.dirname(s.short_path),
+            "optional": True,
+            "compilerFlags": "-fobjc-no-arc", # TODO: why is this not working
```
segiddins, line 242: probably because the flag is `-fno-objc-arc`

PR #92, opened 2020-07-21
Hide static framework target from scheme dropdown on xcode

gyfelton:
For Chartography (internal sample project) project for example, the scheme for Chartography will disappear after this change, good thing is it is still being built by the test or the actual app. 
Bad news is it is different from the xcodeproject generated by the Cocoapods. In the one by cocoapods, this scheme exists and building it will result in the framework itself being built.

It comes down to the original intention of this `continue` stmt in code, **do we hide frameworks from scheme list?** 

Also in this code: https://github.com/bazel-ios/rules_ios/blob/master/tools/xcodeproj_shims/install.sh#L11 we can see that `framework.static` won't be supported. If we add the scheme for a static framework, it will fail here nonetheless during build stage because it is not supported by bazel script.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -273,7 +273,7 @@ $BAZEL_INSTALLER
 """.format(bazel_build_target_name = target_info.bazel_build_target_name),
             }],
         }
-        if target_info.product_type == "framework":
+        if target_info.product_type == "framework" or target_info.product_type == "framework.static":
             continue
 
         scheme_action_name = "test"


segiddins:
I think we want to be able to build an individual framework in Xcode, given it's something that's frequently done in Xcode today (not every framework necessarily has a corresponding test target). (Additionally, `framework.static` isn't actually something Xcode will report, static frameworks have the above product type and the only thing that makes them static is their `MACH_O_TYPE` is set to `staticlib`)

PR #93, opened 2020-07-22
Set correct runnable actions for target/scheme

gyfelton:
### Problem this PR tries to tackle
Previously we have buildForTesting by default unless it is a Framework (no-op) or it is an app (buildForRunning).

Note: please review commit by commit

### What's changed
1. Changed to  buildForTesting and buildForRunning by default given a unit test/framework/app, since it is either buildable (buildForRunning needs to be enabled) and/or testable (which is further defined by each target to determine what is the testing behaviour).
2. For static framework, during the `build with bazel` build phase inside xcode (achieved mainly through xcodeproj_shims/installer.sh), it will grab the correct input path via bazel build event as oppose to a pre-determined path under `bazel-out` because for static framework, its path is not deterministic. With this change, a concrete framework binary file should be present under the output directory upon build succceed.
3. Also update the xcodegen version (to 2.16) we point to so it picks up more improements on xcodegen side

### Tests added
1. under ios, there is a new single-framework project which builds a static framework and has a real unit test testing its behaviour.

### What is tested (what's the new behaviour)
1. Framework target is now able to build, and after build succeed, the output folder will have the following content:
- Framework binary
- Header folder containing .h files (umbrella)
- Modules folder having modules.modulemap
2. scheme files are updated to reflect the new default behaviour


rules/xcodeproj.bzl
```diff
@@ -292,20 +288,20 @@ $BAZEL_INSTALLER
             else:
                 test_env_vars_dict[k] = v
 
-        scheme_action_details["environmentVariables"] = test_env_vars_dict
+        scheme_action_details_for_test["environmentVariables"] = test_env_vars_dict
```
gyfelton, line 291: `test_env_vars_dict` may contain key-value pairs from `scheme_existing_envvar_overrides` that is set up on top level, i wonder whether these overrides should be just for `test` or applies all actions (right now it is just for test)

amberdixon: I think you're right, it should apply to all actions. We could even make it apply to the run action and then have the test action "Use the Run action's arguments and environment variables."

tools/xcodeproj_shims/install.sh
```diff
@@ -12,6 +12,10 @@ case "${PRODUCT_TYPE}" in
     com.apple.product-type.framework)
         input_options=("bazel-bin/$BAZEL_BIN_SUBDIR/${TARGET_NAME}/${FULL_PRODUCT_NAME}")
         ;;
+    com.apple.product-type.framework.static)
+	#TODO: invalid input options here, we need to find the correct input options somewhere else
+        input_options=("bazel-bin/$BAZEL_BIN_SUBDIR/${TARGET_NAME}/${FULL_PRODUCT_NAME}")
```
amberdixon, line 17: Just a reminder to remove/fix this before you land this change.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
```diff
@@ -0,0 +1,68 @@
...
+	    # For static library, as the output is not under bazel-bin, we have to get it based on build event
+        # Example of such entry inside build event:
+        # important_output {
+        #   name: ".../SomeFramework.framework/SomeFramework"
+        #   uri: "file:///private/var/tmp/.../.../SomeFramework.framework/SomeFramework"
+        #   path_prefix:... 
+        #   ...
+        #  } 
+        # We only care about the entry ending with TARGET_NAME" so that we can get the path to its directory
+        QUERY=`grep -A 2 important_output "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep ${TARGET_NAME}\" | sed "s/uri: \"file:\/\///"`
```
gyfelton, line 25: `grep ${TARGET_NAME}\"` is what I have concern with, because right now in order to get the right path that is single and deterministic, I am only looking for the one path ending with the binary file (which is usually just the target name, but it can be different if user sets it to be a diff. thing?)

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
```diff
@@ -12,10 +12,10 @@
 
 /* Begin PBXFileReference section */
 		4BB7A0F977B290B8648FFDB0 /* ObjcFramework.framework */ = {isa = PBXFileReference; includeInIndex = 0; lastKnownFileType = wrapper.framework; path = ObjcFramework.framework; sourceTree = BUILT_PRODUCTS_DIR; };
-		6902D1207F875225638DE30A /* main.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; name = main.m; path = ../../ios/frameworks/objc/main.m; sourceTree = "<group>"; };
```
segiddins: why did these paths change?

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/xcshareddata/xcschemes/iOS-10.0-AppHost.xcscheme
```diff
@@ -9,7 +9,7 @@
          <BuildActionEntry
             buildForTesting = "YES"
             buildForRunning = "YES"
-            buildForProfiling = "YES"
```
segiddins: I think we still should be building for profiling?

rules/xcodeproj.bzl
```diff
@@ -64,21 +64,28 @@ def _xcodeproj_aspect_impl(target, ctx):
     bazel_build_target_name = ""
     if target.label.workspace_name != "":
         bazel_build_target_name = "@%s//" % target.label.workspace_name
-    bazel_build_target_name += "%s:%s" % (target.label.package, target.label.name)
```
segiddins, line 67: in the future, can unrelated formatting changes be made in a separate PR? it makes reviewing easier

rules/xcodeproj.bzl
```diff
@@ -89,17 +96,39 @@ def _xcodeproj_aspect_impl(target, ctx):
             bundle_extension = bundle_info.bundle_extension,
             bazel_build_target_name = bazel_build_target_name,
             bazel_bin_subdir = bazel_bin_subdir,
-            srcs = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "srcs")),
-            non_arc_srcs = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "non_arc_srcs")),
-            asset_srcs = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "asset_srcs")),
-            framework_includes = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "framework_includes")),
-            cc_defines = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "cc_defines")),
-            build_files = depset(_srcs_info_build_files(ctx), transitive = _get_attr_values_for_name(deps, _SrcsInfo, "build_files")),
+            srcs = depset([], transitive = _get_attr_values_for_name(
```
segiddins, line 99: is there any functional change here, or just formatting?

rules/xcodeproj.bzl
```diff
@@ -304,39 +382,43 @@ $BAZEL_INSTALLER
...
+        scheme_action_names = ["run", "test"]
         scheme_action_details = {"targets": [target_info.name]}
 
-        test_env_vars_dict = {}
+        env_vars_dict = {}
 
-        # target_info.test_env_vars looks like (("env_k1", "env_v1"), ("env_k2", "env_v2"))
-        for kvPair in getattr(target_info, "test_env_vars", ()):
+        # this looks like (("env_k1", "env_v1"), ("env_k2", "env_v2"))
+        for kvPair in getattr(target_info, "env_vars", ()):
```
segiddins, line 393: this could be `for (k, v) in getattr(target_info, "env_vars", ()):`

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/xcshareddata/xcschemes/Single-Application-RunnableTestSuite.xcscheme
```diff
@@ -27,18 +27,8 @@
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       onlyGenerateCoverageForSpecifiedTargets = "NO"
-      shouldUseLaunchSchemeArgsEnv = "NO">
+      shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
-         <TestableReference
```
segiddins, line 32: why is the testable reference gone?

tools/xcodeproj_shims/install.sh
```diff
@@ -12,6 +12,19 @@ case "${PRODUCT_TYPE}" in
...
+	    # For static library, as the output is not under bazel-bin, we have to get it based on build event
+        # Example of such entry inside build event:
+        # important_output {
+        #   name: ".../SomeFramework.framework/SomeFramework"
+        #   uri: "file:///private/var/tmp/.../.../SomeFramework.framework/SomeFramework"
+        #   path_prefix:... 
+        #   ...
+        #  } 
+        # We only care about the entry ending with TARGET_NAME" so that we can get the path to its directory
+        QUERY=`grep -A 2 important_output "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep ${TARGET_NAME}\" | sed "s/uri: \"file:\/\///"`
```
gyfelton, line 25: grep ${TARGET_NAME}\" is what I have concern with, because right now in order to get the right path that is single and deterministic, I am only looking for the one path ending with the binary file (which is usually just the target name, but it can be different if user sets it to be a diff. thing?)

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/xcshareddata/xcschemes/Single-Application-RunnableTestSuite.xcscheme
gyfelton: it's back now, turns out having targets under test action is what triggers XCodeGen side to include the TestableReference Section (what tests to run)

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/xcshareddata/xcschemes/Single-Application-RunnableTestSuite.xcscheme
```diff
@@ -78,16 +62,33 @@
...
+      </MacroExpansion>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-h"
+            isEnabled = "YES">
+         </CommandLineArgument>
+      </CommandLineArguments>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "test_envvar_key1"
```
gyfelton, line 82: these moved to Run action now

gyfelton:
I am aware that build is failing and it is failing on a new issue for the new test

tests/ios/frameworks/objc/main.m
```diff
@@ -4,3 +4,11 @@ int main(int argc, char *argv[])
...
     return 0;
 }
+
+@implementation FooFramework: NSObject
+
+- (int)alwaysReturnOne; {
+    return 1;
+}
+
+@end
```
segiddins, line 14: trailing newlines

gyfelton:
@amberdixon ^^ can take a look at this PR?

tools/xcodeproj_shims/install.sh
amberdixon: In the future, we may want to parse the the build_event_json_file.

PR #95, opened 2020-07-28
Disable generation of debug info (dSYM) file by default for all xcode projects

gyfelton:
### Why this change
For some reason instead of generating dwarf only for debug config, by default xcodeGen is specifying to generate dwarf+dSYM for all configs. Generating dsym file can take a long time for big targets, so it's time saving to disable it for now

Also we are seeing lots of warnings from `dsymutil` regarding `swiftmodule` files not found. This will turn off this warnings as well

### Additional changes
1. Fix an error during testing regarding NonArcObj not having NSObject as super class

### DIFF
--- a/tests/macos/xcodeproj/NonArcObject.h
+++ b/tests/macos/xcodeproj/NonArcObject.h
@@ -1,2 +1,2 @@
-@interface NonArcObject
+@interface NonArcObject : NSObject
 @end
--- a/tests/macos/xcodeproj/NonArcObject.h
+++ b/tests/macos/xcodeproj/NonArcObject.h
@@ -1,2 +1,2 @@
-@interface NonArcObject : NSObject
+@interface NonArcObject
 @end
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -209,6 +209,7 @@ def _xcodeproj_impl(ctx):
         "CXX": "$CC",
         "CLANG_ANALYZER_EXEC": "$CC",
         "CODE_SIGNING_ALLOWED": False,
+        "DEBUG_INFORMATION_FORMAT": "dwarf",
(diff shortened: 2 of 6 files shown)


tests/macos/xcodeproj/NonArcObject.h
```diff
@@ -1,2 +1,2 @@
-@interface NonArcObject
+@interface NonArcObject : NSObject
```
ndizazzo, line 1: Is this intentional? This will make it an ARC object now that you are specifying that inherits from `NSObject`?

gyfelton: I believe whether it is ARC or not is defined by compiler flag `-fno-obj-arc`, @amberdixon can correct me here? 
Also it's weird that my previous PR makes CI fail on this...

amberdixon: What we are really trying to test here is that the builder process treats this file like a non-arc object and adds the no-objc-arc flag. So with or without this change, our tests and builds should work as expected.

Even before ARC came along, NSObject was used I believe, so I'm not sure inheriting NSObject causes the file to become ARC ... i googled around a little, can you share any links/docs that indicate that that would be the case?

ndizazzo: For whatever reason, I thought ARC was managed at the object level... but brushing up on the [documentation for ARC](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#general) (it's been a while) reminded me it just inserts retain/release for you during compilation. Sorry for the confusion.

gyfelton: yeah i have not seen `retain` `release` for ages :)

gyfelton:
Waiting for @segiddins to give a final approval

PR #96, opened 2020-07-30
Pass swift defines into xcode build setting so indexing works on both swift and objc side

gyfelton:
### Why this change
This is intend to be a counter-part for swift as the objc side is done in this PR: https://github.com/bazel-ios/rules_ios/pull/90
With this change, any flag passed in as `defines` along the dependency graph gets surfaced up as `SWIFT_ACTIVE_COMPILATION_CONDITIONS` build setting inside xcode project.

note: this is NOT for building to work (it already works) but for indexing/showing error on XCode to work because during indexing, xcode will compile the code again using all the flags/settings on XCode (as evidenced by xcode producing error even though tests pass if flags are changed on build setting (xcode side)

note #2: since all flags are right now under one field of build.bazel: `defines`, it's possible that it has the form of `foo=bar` which means it should not be passed to swift side if `foo=0` (or `foo=false`?) This logic is now in this PR

### Tests done
1. Build pass even though swift and objc files both contain bad text (but wrapped under a macro that is evaluated to be zero). And xcode show no error on that line

2. Test still pass even though there are bad text and a `assertTure(false)` inside.

### Additional change for gcc_preprocessor_macro:
For debug config it is generally expected to have `DEBUG` as the macro showing up on both swift and objc side, this is now added as part of the project build setting as well to match this implicit expectation.


rules/xcodeproj.bzl
```diff
@@ -269,9 +278,22 @@ def _xcodeproj_impl(ctx):
...
             target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
             target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
+        else:
+            defines_without_equal_sign = []
+            for d in target_info.swift_defines.to_list():
+                splits = d.split("=")
+                append_this_define = True
+
+                # So that definition with value zero won't be populated
+                if len(splits) > 1 and splits[1] == "0":
```
gyfelton, line 292: so we have Foo=0 defined, it won't show up on swift side at all

rules/xcodeproj.bzl
```diff
@@ -136,9 +139,14 @@ def _xcodeproj_aspect_impl(target, ctx):
...
         framework_includes = _get_attr_values_for_name(deps, _SrcsInfo, "framework_includes")
         cc_defines = _get_attr_values_for_name(deps, _SrcsInfo, "cc_defines")
+        swift_defines = _get_attr_values_for_name(deps, _SrcsInfo, "swift_defines")
         if CcInfo in target:
             framework_includes.append(target[CcInfo].compilation_context.framework_includes)
             cc_defines.append(target[CcInfo].compilation_context.defines)
+
+        if SwiftInfo in target:
+            swift_defines.append(depset(target[SwiftInfo].direct_defines))
+            swift_defines.append(target[SwiftInfo].transitive_defines)
```
amberdixon, line 149: Should we also be looking at the `transitive_modulemaps` field in SwiftInfo? I dug in a little bit, but I'm not sure.
The field contains "The transitive module map files that will be passed to Clang using the <code>-fmodule-map-file</code> option."

rules/xcodeproj.bzl
```diff
@@ -269,9 +278,22 @@ def _xcodeproj_impl(ctx):
             framework_search_paths.append("\"%s\"" % fi)
         target_settings["FRAMEWORK_SEARCH_PATHS"] = " ".join(framework_search_paths)
         target_settings["GCC_PREPROCESSOR_DEFINITIONS"] = " ".join(["\"%s\"" % d for d in target_info.cc_defines.to_list()])
+
         if target_info.product_type == "application":
             target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
             target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
+        else:
```
amberdixon, line 285: Why do we only do this if the target is not an ios-application? Shouldn't we always do it if the `target_info.swift_defines` are there?

rules/xcodeproj.bzl
gyfelton: put them into this setting? Or into a separate build setting?

rules/xcodeproj.bzl
```diff
@@ -269,9 +278,22 @@ def _xcodeproj_impl(ctx):
...
+            defines_without_equal_sign = []
```
amberdixon, line 286: Based upon our discussion with @segiddins  this morning, I *think* we need to also add the `DEBUG` value to the list of defines for the debug xcconfig. This is done by cocoapods (which adds this build setting to xcode) and also by the bazel toolchain when it performs builds without xcode ... but our solution needs to do it as well I think.

rules/xcodeproj.bzl
gyfelton: nvm i found that for normal application it will show up. the one looked at is test host

rules/xcodeproj.bzl
gyfelton: i see will add it here! wonder whether we can tell which config we are on at this point (passing DEBUG to release config can be a bad thing no?)

rules/xcodeproj.bzl
segiddins: it'd go into `OTHER_SWIFT_FLAGS`, if its needed

rules/xcodeproj.bzl
segiddins: this is a bunch of logic here -- could you extract it into a separate function?

rules/xcodeproj.bzl
segiddins: See `swift/internal/compiling.bzl` in rules_swift -- this seems like a re-implementation of `_exclude_swift_incompatible_define`

rules/xcodeproj.bzl
segiddins: We should pass `DEBUG` for debug config, see https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#settings for docs on how to do per-config settings

rules/xcodeproj.bzl
gyfelton: i will create a separate ticket to track this

rules/xcodeproj.bzl
segiddins: I still think this would be easier to follow if the logic for transforming defines for swift was extracted to a separate function

rules/xcodeproj.bzl
gyfelton: wow how did you find this method... indeed it is identical in terms of the goal of the method. 
however based off the documentation of its caller `_conditional_compilation_flag_configurator` it only adds the flags to command line (that's how we get swift compile correctly?) and for us it's to add the flags into build setting.
If I remove this check the test will fail because there is a `foo=0` flag being added as well. Therefore I just borrowed their code to make them consistent to each other. Is that an okay thing to do?

rules/xcodeproj.bzl
gyfelton: separate ticket created

rules/xcodeproj.bzl
gyfelton: we now pass `DEBUG` for debug config in the project level, all targets will inherit it from the project setting

gyfelton:
@amberdixon another look since there are additional changes introduced into this PR? 🙏

rules/xcodeproj.bzl
```diff
@@ -268,10 +308,24 @@ def _xcodeproj_impl(ctx):
...
+        macros.append("$(inherited)")
+        target_settings["GCC_PREPROCESSOR_DEFINITIONS"] = " ".join(macros)
+
         if target_info.product_type == "application":
             target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
             target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
+        else:
+            defines_without_equal_sign = ["$(inherited)"]
+            for d in target_info.swift_defines.to_list():
+                defines_without_equal_sign.append(d)
```
amberdixon, line 321: I think line 321 should not be here. Or else you will end up appending everything.

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/project.pbxproj
```diff
@@ -415,11 +420,12 @@
 				CLANG_ENABLE_MODULES = YES;
 				CLANG_ENABLE_OBJC_ARC = YES;
 				FRAMEWORK_SEARCH_PATHS = "";
-				GCC_PREPROCESSOR_DEFINITIONS = "\"REQUIRED_DEFINED_FLAG\"";
+				GCC_PREPROCESSOR_DEFINITIONS = "\"REQUIRED_DEFINED_FLAG=1\" \"FLAG_WITH_VALUE_ZERO=0\" $(inherited)";
 				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
 				MACH_O_TYPE = "$(inherited)";
 				PRODUCT_NAME = "Single-Application-UnitTests";
 				SUPPORTS_MACCATALYST = 0;
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "\"$(inherited)\" \"REQUIRED_DEFINED_FLAG=1\" \"REQUIRED_DEFINED_FLAG\" \"FLAG_WITH_VALUE_ZERO=0\"";
```
amberdixon, line 428: Over here, `FLAG_WITH_VALUE_ZERO` should not be present, and `REQUIRED_DEFINED_FLAG` should only be present once. It should look like `REQUIRED_DEFINED_FLAG`

tests/macos/xcodeproj/NonArcObject.m
```diff
@@ -1,4 +1,11 @@
 #import "NonArcObject.h"
 
 @implementation NonArcObject
+#if REQUIRED_DEFINED_FLAG
 @end
+#endif
+
+#if FLAG_WITH_VALUE_ZERO
```
amberdixon, line 8: I'm not sure why this code is compiling .... it seems like `FLAG_WITH_VALUE_ZERO` has been mistakenly specified due to bug I noted above, where everything is getting appended to `defines_without_equals_sign`

tests/macos/xcodeproj/tests.sh
```diff
@@ -2,6 +2,9 @@ set -eux
 
 cd $(dirname $0)
 
+# Make sure there are simulators avilable as destinations
+xcrun simctl list
```
amberdixon, line 6: What will happen if there are no available sims? It seems like this command just lists the available sims. It doesn't error out the script or create any new sims, in the case that there are no sims available.

tests/macos/xcodeproj/NonArcObject.m
gyfelton: we have discussed and resolved this

tests/macos/xcodeproj/tests.sh
gyfelton: this implicitly creates the simulators if not exist, so that the next step is smooth and won't end up having no simulators to use, i have updated the comment here

gyfelton:
@segiddins merging this to unblock my next PR, going to set up time with you to discuss the specifics inside that needs to be improved/fixed

PR #97, opened 2020-07-31
Fix LLDB, swiftmodules xcode installers

amberdixon:
This PR is part of a fix for an issue where the debugger was returning the error "Cannot create Swift scratch context (couldn't load the Swift stdlib)Shared Swift state for <AppName> could not be initialized" for swift code. When breakpoints were hit, it was not possible to view variables values.

This partly fixes the issue, because it tells the debugger where to find swift modules and the SDK for swift. It also fixes an issue with the logic to copy swiftmodules into derived data.

In order for debugging to work, it is also important to set the `serialize-debugging-options` flag to true for the binary (so that the AST will refer to absolute paths on the machine that built the binary.) See further discussion about Swift AST Context here: https://forums.swift.org/t/improving-path-remappings-for-extraclangargs-swiftastcontext/26886


tools/xcodeproj_shims/installers/swiftmodules.sh
```diff
@@ -4,8 +4,21 @@ set -euo pipefail
 
 # Copy Bazel build `.swiftmodule` files to `DerivedData`. This is used by Xcode
 # and its indexing.
-find "$BAZEL_WORKSPACE_ROOT/bazel-out"/*/bin/ \
-    -name "*.swiftmodule" \
-     -not -path "*/_swift_module_cache/*" \
-     -print0 \
-    | xargs -0 "$BAZEL_INSTALLERS_DIR/_swiftmodule.sh"
+echo "Start copying swiftmodules at `date`"
+FOUND_SWIFTMODULES=`pcregrep -o1 'primary_output: "(.*\.swiftmodule)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | uniq`
```
gyfelton, line 8: I wonder what's the difference between `primary_output` and `important_output`....

tools/xcodeproj_shims/installers/swiftmodules.sh
```diff
@@ -4,8 +4,21 @@ set -euo pipefail
...
+  fi
+done
+echo "Found ${#EXISTING_SWIFTMODULES[@]} existing SWIFTMODULES"
+if [ ${#EXISTING_SWIFTMODULES[@]} -ne 0 ]
+then
+  "$BAZEL_INSTALLERS_DIR/_swiftmodule.sh" ${EXISTING_SWIFTMODULES[*]}
+else
+  echo "No SWIFTMODULES found"
+fi
+echo "Finish copying swiftmodules at `date`"
```
segiddins, line 24: trailing newline

tools/xcodeproj_shims/installers/swiftmodules.sh
amberdixon: After some digging, I learned that they are the same thing, but that `important_output` is deprecated:

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto

```
// Temporarily, also report the important outputs directly. This is only to
  // allow existing clients help transition to the deduplicated representation;
  // new clients should not use it.
  repeated File important_output = 4 [deprecated = true];
```

```
  // Identifier of an event reporting that an action was completed (not all
  // actions are reported, only the ones that can be considered important;
  // this includes all failed actions).
  message ActionCompletedId {
    string primary_output = 1;
    // Optional, the label of the owner of the action, for reference.
    string label = 2;
    // Optional, the id of the configuration of the action owner.
    ConfigurationId configuration = 3;
  }
```

PR #98, opened 2020-08-04
Add local debug options as a dependency, but only if local_debug_options_enabled

amberdixon:
Only do this if local_debug_options_enabled is specified at command line. Make xcode specify this config.

The changes in this PR are taken from the suggested changes in https://github.com/ios-bazel-users/ios-bazel-users/blob/master/DebuggableRemoteSwift.md to get debugging working.

Note that in order for the example applications and tests in the rules_ios repo to be debuggable from Xcode, you must:
1. Disable sandboxing in .bazelrc.
2. Update the flag in the tools/xcodeproj_shims/build-wrapper.sh to refer to `--//rules:local_debug_options_enabled` rather than `--@build_bazel_rules_ios//rules:local_debug_options_enabled`

There are still known issues with evaluating variables from the debugger for swift tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -15,5 +15,5 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 
 # Enable these features to test local and CI builds
 # when swiftmodule caching is enabled.
-build --features=swift.cacheable_swiftmodules
-build --features=swift.use_global_module_cache
+# build --features=swift.cacheable_swiftmodules
+# build --features=swift.use_global_module_cache
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,9 +1,19 @@
 # We can't create a bzl_library for rules-swift because of its visibility,
 # so circumvent by not using the sandbox
 build --strategy=Stardoc=standalone
+
+# Debugging does not work in sandbox mode. Uncomment these lines to turn off sandboxing.
+# build --genrule_strategy=standalone
+# build --spawn_strategy=standalone
+
 build --verbose_failures # Print the full command line for commands that failed
 build --test_output=errors # Prints log file output to the console on failure
 
-# By default do not build the tests for sources-with-prebuilt-binaries, 
+# By default do not build the tests for sources-with-prebuilt-binaries,
 # because it takes quite some time. They will only run on CI
 build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
+
+# Enable these features to test local and CI builds
+# when swiftmodule caching is enabled.
+build --features=swift.cacheable_swiftmodules
+build --features=swift.use_global_module_cache
--- a/rules/BUILD.bazel
+++ b/rules/BUILD.bazel
@@ -1,4 +1,39 @@
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
+
+
 exports_files(
     glob(["*.bzl"]),
     visibility = ["//docs:__pkg__"],
+)
+
+genrule(
+    name = "empty",
+    outs = ["empty.swift"],
+    cmd = "touch $(OUTS)",
+)
+
+# A dummy target that enables serialize-debugging-options but only in local development,
+# so that swiftmodule paths will refer to paths on the local machine and LLDB will work.
+swift_library(
+    name = "_LocalDebugOptions",
+    srcs = [":empty"],
+    copts = [
+            "-Xfrontend",
+            "-serialize-debugging-options",
+    ],
+    module_name = "_LocalDebugOptions",
+    tags = ["no-remote"],
+    visibility = ["//visibility:public"],
+)
+
+
+bool_flag(name = "local_debug_options_enabled", build_setting_default = False)
+
+config_setting(
+    name = "local_debug_options",
+    flag_values = {
+        ":local_debug_options_enabled": "True",
+    }
 )
--- a/rules/BUILD.bazel
+++ b/rules/BUILD.bazel
@@ -16,6 +16,7 @@ genrule(
 
 # A dummy target that enables serialize-debugging-options but only in local development,
 # so that swiftmodule paths will refer to paths on the local machine and LLDB will work.
+# See details here: https://github.com/ios-bazel-users/ios-bazel-users/blob/master/DebuggableRemoteSwift.md
 swift_library(
     name = "_LocalDebugOptions",
     srcs = [":empty"],
--- a/rules/BUILD.bazel
+++ b/rules/BUILD.bazel
@@ -1,7 +1,5 @@
-
 load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
-
 
 exports_files(
     glob(["*.bzl"]),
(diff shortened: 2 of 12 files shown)


rules/library.bzl
```diff
@@ -506,7 +506,10 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             generated_header_name = generated_swift_header_name,
             srcs = swift_sources,
             copts = swift_copts,
-            deps = deps + internal_deps + lib_names,
+            deps = deps + internal_deps + lib_names + select({
```
segiddins, line 509: do we really want to do this for every `swift_library` ? Ideally, it'd only be done for `ios_application` and `ios_unit_test`, so _most_ work can be shared between the flag being in either state

tests/macos/xcodeproj/build.sh
```diff
@@ -4,4 +4,5 @@ cd $(dirname $0)
 
 xcodebuild -project Single-Application-Project-DirectTargetsOnly.xcodeproj -quiet
 xcodebuild -project Single-Application-Project-AllTargets.xcodeproj -quiet
-xcodebuild -project Test-Target-With-Test-Host-Project.xcodeproj -quiet
\ No newline at end of file
+xcodebuild -project Test-Target-With-Test-Host-Project.xcodeproj -quiet
+# TODO: use strings to test that absolute paths and debug-prefix-map are present in the LocalDebug.swiftmodule. Not sure if this is the right palce to do it. We could also grep out the build event text file maybe.
```
segiddins, line 8: trailing newline

rules/library.bzl
amberdixon: Let me see if there's a way to do that. thanks!

amberdixon: I updated this macro to accept a parameter for whether or not to add swift local debug options.

rules/app.bzl
```diff
@@ -56,7 +56,7 @@ def ios_application(name, apple_library = apple_library, **kwargs):
     """
     infoplists = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
     application_kwargs = {arg: kwargs.pop(arg) for arg in _IOS_APPLICATION_KWARGS if arg in kwargs}
-    library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": application_kwargs.get("minimum_os_version")}, **kwargs)
+    library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": application_kwargs.get("minimum_os_version")}, add_swift_local_debug_options = True, **kwargs)
```
segiddins, line 59: should also be passed for tests as well?

segiddins: Alternatively, we could only add in the select to the `deps` passed to `rules_apple_ios_application `, and make it conditional on whether there were actually any swift files? That way, the swiftmodule that represents the app needn't be invalidated by switching the flag either

kastiglione: as of Xcode 11.5, we don't need this work around for apps, only tests.

rules/BUILD.bazel
```diff
@@ -1,4 +1,39 @@
...
+# so that swiftmodule paths will refer to paths on the local machine and LLDB will work.
+swift_library(
+    name = "_LocalDebugOptions",
+    srcs = [":empty"],
+    copts = [
+            "-Xfrontend",
+            "-serialize-debugging-options",
+    ],
+    module_name = "_LocalDebugOptions",
+    tags = ["no-remote"],
```
segiddins, line 27: Nit: can we also tag as manual?

tests/macos/xcodeproj/build.sh
segiddins: Nit: spelling of place

amberdixon:
I am still seeing issues with test bundles locally. I get errors like this. This happens for tests written in swift both with and without an app host:

```
AppRoot-Unit-Tests: error: cannot load underlying module for 'XCTest'

Debug info from this module will be unavailable in the debugger.

warning: Swift error in scratch context: error: missing required modules:

```

That said, I think the changes in this PR are still correct, but perhaps we need to do a few more things in terms of setting additional lldb settings. Going to open this up for review and then continue working on these issues I'm seeing with tests locally.

segiddins:
> AppRoot-Unit-Tests: error: cannot load underlying module for 'XCTest'

We might need to add in `$(PLATFORM_DIR)/Developer/Library/Frameworks` to `FRAMEWORK_SEARCH_PATHS` for test target types?

tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -euxo pipefail
 
-$BAZEL_PATH build --experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME --build_event_text_file=$BAZEL_BUILD_EVENT_TEXT_FILENAME --build_event_publish_all_actions $1 2>&1 | $BAZEL_OUTPUT_PROCESSOR
+$BAZEL_PATH build --experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME --build_event_text_file=$BAZEL_BUILD_EVENT_TEXT_FILENAME --build_event_publish_all_actions $1 --@build_bazel_rules_ios//rules:local_debug_options_enabled 2>&1 | $BAZEL_OUTPUT_PROCESSOR
```
gyfelton, line 4: run the following to get test fixtures updated:
```
bazelisk query 'kind(xcodeproj, tests/ios/xcodeproj/...)' | xargs -n 1 bazelisk run
bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run
bazelisk query 'attr(executable, 1, kind(genrule, tests/macos/xcodeproj/...))' | xargs -n 1 bazelisk run
bazelisk query 'attr(executable, 1, kind(genrule, tests/ios/xcodeproj/...))' | xargs -n 1 bazelisk run
find . -type f \( -name 'WORKSPACE' -o -name '*.bzl' -o -name '*.bazel' \) | xargs buildifier -lint=fix
```

rules/app.bzl
```diff
@@ -61,9 +61,16 @@ def ios_application(name, apple_library = apple_library, **kwargs):
     application_kwargs["launch_storyboard"] = application_kwargs.pop("launch_storyboard", library.launch_screen_storyboard_name)
     application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
 
+    local_debug_options_for_swift = []
+    if library.has_swift_sources:
+      local_debug_options_for_swift += ["@build_bazel_rules_ios//rules:_LocalDebugOptions"]
```
gyfelton, line 66: should we update some documentation / readme to reflect this important flag?

amberdixon: I added comments in the BUILD.bazel where the _LocalDebugOptions target is defined. I will also reference this README: https://github.com/ios-bazel-users/ios-bazel-users/blob/master/DebuggableRemoteSwift.md

.bazelrc
```diff
@@ -1,9 +1,19 @@
...
 
-# By default do not build the tests for sources-with-prebuilt-binaries, 
+# By default do not build the tests for sources-with-prebuilt-binaries,
 # because it takes quite some time. They will only run on CI
 build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
+
+# Enable these features to test local and CI builds
+# when swiftmodule caching is enabled.
+build --features=swift.cacheable_swiftmodules
+build --features=swift.use_global_module_cache
```
segiddins, line 19: nit: trailing newline

tests/macos/xcodeproj/build.sh
```diff
@@ -4,5 +4,4 @@ cd $(dirname $0)
 
 xcodebuild -project Single-Application-Project-DirectTargetsOnly.xcodeproj -quiet
 xcodebuild -project Single-Application-Project-AllTargets.xcodeproj -quiet
-xcodebuild -project Test-Target-With-Test-Host-Project.xcodeproj -quiet
-# TODO: use strings to test that absolute paths and debug-prefix-map are present in the LocalDebug.swiftmodule. Not sure if this is the right palce to do it. We could also grep out the build event text file maybe.
\ No newline at end of file
+xcodebuild -project Test-Target-With-Test-Host-Project.xcodeproj -quiet
```
segiddins: nit: trailing newline

rules/app.bzl
segiddins: Adding separate docs shouldn't be necessary, since the xcodeproj rule automatically enables the flag when necessary

PR #99, opened 2020-08-06
Pass hmap files to header search path inside xcode build setting

gyfelton:
### Why this change
In order for Xcode autocomplete/jump to definition etc. to work (dev/debugging experience), indexing has to  work fully against the files included in the project.

Right now it can be not working for various reasons, one of which is that the clang invocation required to populate index store and make sense of the symbols inside the file is not having the same set of flags as the invocation done inside bazel (`-I` in specific). This is because in `library.bzl` we generates a set of hmaps and include them as set of includes for clang. 

So this PR focuses on mimic the same behaviour inside `xcodeproj.bzl` aspect implementation so that xcode knows about the headers needed for each framework target. By default the root is included as well (tulsi does the same thing)

### Tests done
Because the test projects is too simple, it does not show error on indexing before this change. 
Therefore I need to come up with a situation (probably new dependencies on existing framework) so that it will have the `header not found` error before this change


rules/xcodeproj.bzl
```diff
@@ -43,6 +43,18 @@ def _srcs_info_build_files(ctx):
 
     return [path]
 
+def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
+    hmap_paths = []
```
amberdixon, line 47: You could instead look for the HeaderMapInfo in the providers for the dep. Also, for correctness, you should probably keep the private hmap flags in there too. That doesn't break anything, does it? 

```
def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
    hmap_paths = []
    kind = ctx.rule.kind
    if kind == "objc_library" or kind == "swift_library":
        for dep in deps:
            if dep and HeaderMapInfo in dep:  # type(dep) == "Target":
                path = ctx.expand_location("$(execpath :%s)" % dep.label.name)
                hmap_paths.append(path)
    return hmap_paths
```

amberdixon: You may also want to leave a comment explaining that we are doing this, because the library.bzl macro generates hmaps and refers to them in the value of the `-I` option.

rules/xcodeproj.bzl
```diff
@@ -263,13 +280,27 @@ def _xcodeproj_impl(ctx):
             "CLANG_ENABLE_MODULES": "YES",
             "CLANG_ENABLE_OBJC_ARC": "YES",
         }
+
+        # Just like framework, we need to have absoluate path to the hmap files
+        # so that Objc files can correctly import headers
+        header_search_paths = []
+        for hmap in target_info.hmap_paths.to_list():
+            if len(hmap) == 0:
```
amberdixon, line 288: Why would the hmap be an empty string?

rules/xcodeproj.bzl
```diff
@@ -263,13 +280,27 @@ def _xcodeproj_impl(ctx):
...
+                continue
+            if hmap != "." and hmap[0] != "/":
+                hmap = "$BAZEL_WORKSPACE_ROOT/%s" % hmap
+            header_search_paths.append("\"%s\"" % hmap)
+        header_search_paths.append("\"$BAZEL_WORKSPACE_ROOT\"")
```
amberdixon, line 293: For correctness (and could do this in a separate PR), we should read the appropriate values from the compilation context:

```
  /**
   * Returns the immutable list of include directories to be added with "-I" (possibly empty but
   * never null). This includes the include dirs from the transitive deps closure of the target.
   * This list does not contain duplicates. All fragments are either absolute or relative to the
   * exec root (see {@link
   * com.google.devtools.build.lib.analysis.BlazeDirectories#getExecRoot(String)}).
   */
  public ImmutableList<PathFragment> getIncludeDirs() {
    return commandLineCcCompilationContext.includeDirs;
  }

  /**
   * Returns the immutable list of include directories to be added with "-iquote" (possibly empty
   * but never null). This includes the include dirs from the transitive deps closure of the target.
   * This list does not contain duplicates. All fragments are either absolute or relative to the
   * exec root (see {@link
   * com.google.devtools.build.lib.analysis.BlazeDirectories#getExecRoot(String)}).
   */
  public ImmutableList<PathFragment> getQuoteIncludeDirs() {
    return commandLineCcCompilationContext.quoteIncludeDirs;
  }
  /**
   * Returns the immutable list of include directories to be added with "-isystem" (possibly empty
   * but never null). This includes the include dirs from the transitive deps closure of the target.
   * This list does not contain duplicates. All fragments are either absolute or relative to the
   * exec root (see {@link
   * com.google.devtools.build.lib.analysis.BlazeDirectories#getExecRoot(String)}).
   */
  public ImmutableList<PathFragment> getSystemIncludeDirs() {
    return commandLineCcCompilationContext.systemIncludeDirs;
  }

  /**
   * Returns the immutable set of declared include directories, relative to a "-I" or "-iquote"
   * directory" (possibly empty but never null).
   */
  public NestedSet<PathFragment> getLooseHdrsDirs() {
    return looseHdrsDirs;
  }
```
So this means you can access this info by calling `target[CcInfo].compilation_context.include_dirs, quote_include_dirs, system_include_dirs and loose_hdrs_dirs` in the same way I accessed the `framework_includes` property of CompilationContext.

(I searched CcCompilationContext.java in bazel source for references to -I and -i.)

There is probably a corresponding set of info for swift in the swift_library providers.

amberdixon: Oddly Tulsi does not seem to access these properties, unless I'm grepping wrong. I'm not sure why, it seems like the right thing to do.

rules/xcodeproj.bzl
gyfelton: So regarding the `private_hmap ` one it is because `library.bzl` only has it under `-iquote` instead of `-I`, if we include this then private headers will be exposed as public ones? 
Now think again what I realize is that if someone did this, then build will not pass (assuming what bazel did is source of truth), then it does not matter whether I match the flags here or not. Therefore in the latest commit I will not do this check to make code cleaner

amberdixon: `if we include this then private headers will be exposed as public ones?` Bazel's clang invocation does reference the private hmaps. We aren't exposing private headers per se, we are just making sure that the way we build these in bazel is the same as the way we do it in xcode.

It seems like library.bzl adds private_hmap as both `-I` and `-iquote`.

rules/xcodeproj.bzl
gyfelton: yeah definitely a separate PR to explore this way of getting the dirs. Given existing way is one of the correct ways to achieve what we want

rules/xcodeproj.bzl
```diff
@@ -44,6 +45,20 @@ def _srcs_info_build_files(ctx):
 
     return [path]
 
+def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
+    hmap_paths = []
+    kind = ctx.rule.kind
+    is_library = kind == "objc_library" or kind == "swift_library"
```
segiddins, line 51: why is the `kind` check necessary? shouldn't we be able to just look for the `HeaderMapInfo` provider without caring what rule it was instantiated in?

rules/xcodeproj.bzl
```diff
@@ -44,6 +45,20 @@ def _srcs_info_build_files(ctx):
...
+    if is_library:
+        for dep in deps:
+            if type(dep) == "Target" and HeaderMapInfo in dep:
```
segiddins, line 54: this `type` check shouldn't be necessary

rules/xcodeproj.bzl
```diff
@@ -303,6 +323,19 @@ def _xcodeproj_impl(ctx):
             "CLANG_ENABLE_MODULES": "YES",
             "CLANG_ENABLE_OBJC_ARC": "YES",
         }
+
+        # Just like framework, we need to have absoluate path to the hmap files
```
segiddins, line 327: sp: `absolute`

rules/xcodeproj.bzl
```diff
@@ -303,6 +323,19 @@ def _xcodeproj_impl(ctx):
...
+        # so that Objc files can correctly import headers
+        header_search_paths = []
```
segiddins, line 329: could this logic also be extracted out into a separate function?

rules/xcodeproj.bzl
gyfelton: got `unsupported binary operation: Provider in NoneType`, should i do sth like `if type(dep) != "NoneType"`? Tried this and still have same error 😢

segiddins: sounds like we're somehow passing in a list that includes `None` into the list of `deps`, and we probably should filter that out ahead of time?

rules/xcodeproj.bzl
gyfelton: I thought not doing this check will bring unrelated hmaps here, but after testing this is not the case. Will remove this check. Thanks!

rules/xcodeproj.bzl
```diff
@@ -44,6 +45,17 @@ def _srcs_info_build_files(ctx):
 
     return [path]
 
+def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
+    hmap_paths = []
+    for dep in deps:
+        if type(dep) == "Target" and HeaderMapInfo in dep:
```
segiddins: still shouldnt need the `type` check, as the `deps` param should only be passed a list of Target objects

rules/xcodeproj.bzl
```diff
@@ -52,6 +64,8 @@ def _xcodeproj_aspect_impl(target, ctx):
     deps += getattr(ctx.rule.attr, "infoplists", [])
     deps.append(getattr(ctx.rule.attr, "entitlements", None))
```
segiddins, line 65: insted of checking for `None` all over the place, why don't we only `append` the entitlements target here conditionally?

```starlark
entitlements = getattr(ctx.rule.attr, "entitlements", None)
if entitlements:
    deps.append(entitlements)
```

gyfelton: so if entitlements:  will fail if entitlements  is None?
I wonder if other places will add None to it upstream.

PR #101, opened 2020-08-14
Force swift local debug option to be turned on so that framework search path has correct payload after project is built

gyfelton:
### Why this change
With the change introduced in https://github.com/bazel-ios/rules_ios/pull/98, when bazel builds behind the scene (inside 1build-wrapper1), it will generates one that enables swift to be debuggable. However the framework search paths generated during the xcodeproj generation stage (`bazel run some_xcodeprojec_rule`) will generate a different configuration cache than the one generated by bazel build (since inside build wrapper there are additional flags: 1. compilation mode and 2. local_debug_option)
As evidenced by the test fixture changes, the search paths are changed in this PR because 
1. 1compilation_mode1 is set to debug to match with when xcode builds under debug config (not release config, so under release this is still broken)
2. `local_debug_option` is set to True for both xcode proj rule (through transition) and inside build-wrapper

The direct effect is that the framework search path is now pointing at the correct location.

### Known issue:
Release config won't work with this change because compilation mode for debug is only set for Debug config, wondering if we should just have Release config pointing to `compilation_mode = dbg` too since the expectation is that ppl use release config under xcode is because they want to test the behaviors of the MACROS? 

### Tests
1. Existing fixtures
2. TODO: add a step after xcodebuild to verify that each framework specified under framework search path and header search path can be all found

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,3 +17,6 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # when swiftmodule caching is enabled.
 # build --features=swift.cacheable_swiftmodules
 # build --features=swift.use_global_module_cache
+
+# Match this with the flag used in build-wrapper.sh
+build --compilation_mode=dbg
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,6 +18,3 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # build --features=swift.cacheable_swiftmodules
 # build --features=swift.use_global_module_cache
 
-# Match this with the flag used in build-wrapper.sh
-build --compilation_mode=dbg
-
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,3 +20,4 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 
 # Match this with the flag used in build-wrapper.sh
 build --compilation_mode=dbg
+
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,4 +17,3 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # when swiftmodule caching is enabled.
 # build --features=swift.cacheable_swiftmodules
 # build --features=swift.use_global_module_cache
-
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Generate Xcode projects and compare against existing record
         run: |
           set -euo pipefail
+          bazelisk clean
           bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run
           bazelisk query 'attr(executable, 1, kind(genrule, tests/macos/xcodeproj/...))' | xargs -n 1 bazelisk run
           git diff --exit-code tests/macos/xcodeproj
@@ -64,10 +65,15 @@ jobs:
       - name: Generate Xcode projects and compare against existing record
         run: |
           set -euo pipefail
+          bazelisk clean
           bazelisk query 'kind(xcodeproj, tests/ios/xcodeproj/...)' | xargs -n 1 bazelisk run
           bazelisk query 'attr(executable, 1, kind(genrule, tests/ios/xcodeproj/...))' | xargs -n 1 bazelisk run
           git diff --exit-code tests/ios/xcodeproj
+      - name: Run pre build checks
+        run: ./tests/ios/xcodeproj/pre_build_check.sh
       - name: Run Xcode builds
         run: ./tests/ios/xcodeproj/build.sh
+      - name: Run post build checks
+        run: ./tests/ios/xcodeproj/post_build_check.sh
       - name: Run Unit tests
         run: ./tests/ios/xcodeproj/tests.sh
--- a/rules/transition_support.bzl
+++ b/rules/transition_support.bzl
@@ -118,7 +118,23 @@ _apple_rule_transition = transition(
     ],
 )
 
+def _xcodeproj_rule_transition_impl(settings, attr):
+    _ignore = (settings, attr)
+    return [
+        {"@build_bazel_rules_ios//rules:local_debug_options_enabled": True},
+    ]
+
+
+_xcodeproj_rule_transition = transition(
+    implementation=_xcodeproj_rule_transition_impl,
+    inputs=[],
+    outputs=[
+        "@build_bazel_rules_ios//rules:local_debug_options_enabled"
+    ]
+)
+
 transition_support = struct(
-    apple_rule_transition = _apple_rule_transition,
-    current_apple_platform = _current_apple_platform,
+    apple_rule_transition=_apple_rule_transition,
+    current_apple_platform=_current_apple_platform,
+    xcodeproj_rule_transition=_xcodeproj_rule_transition,
 )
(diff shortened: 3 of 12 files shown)


rules/transition_support.bzl
```diff
@@ -118,7 +118,22 @@ _apple_rule_transition = transition(
...
+    inputs = [],
+    outputs = [
+        "@build_bazel_rules_ios//rules:local_debug_options_enabled",
+    ],
+)
+
 transition_support = struct(
     apple_rule_transition = _apple_rule_transition,
     current_apple_platform = _current_apple_platform,
+    xcodeproj_rule_transition = _xcodeproj_rule_transition,
```
segiddins, line 138: maybe it's a bit more clear to name this `force_swift_local_debug_options_transition` ?

.bazelrc
```diff
@@ -17,3 +17,7 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # when swiftmodule caching is enabled.
 # build --features=swift.cacheable_swiftmodules
 # build --features=swift.use_global_module_cache
+
+# Match this with the flag used in build-wrapper.sh
+build --compilation_mode=dbg
```
segiddins, line 22: why is this necessary? builds in this rules repo should work in all 3 compilation modes

gyfelton: hmmm so the `bazel-out` directory name is a composition of compilation mode and configruation hash and other architecture parameters, right? If we don't specify this one, the framework search path will be `applebin.....-fastbuild-<somehash>` instead of `applebin.....-debug-<somehash>`. Am I correct here? (the change inside fixture also confirms this)

gyfelton: so unless we make the build-wrapper respect the compilation mode in first place instead of hardcoding?

segiddins: either the build wrapper should respect the compilation mode or, my hunch is, we'll need a split transition so we can get the full matrix of settings for {debug,release}/{simulator,device}

gyfelton: will do the `wrapper respect setting` one and do the split transition to get full matrix of settings for  {debug,release}/{simulator,device} in future

gyfelton: okay turns out adding the macro in the first place is not needed because if we specify compile mode to be `dbg` under `.bzelrc`, it will be respected inside the build-warpper. unless i missed some case @amb

.github/workflows/tests.yml
```diff
@@ -47,11 +47,18 @@ jobs:
       - name: Generate Xcode projects and compare against existing record
         run: |
           set -euo pipefail
+          bazlisk clean
           bazelisk query 'kind(xcodeproj, tests/macos/xcodeproj/...)' | xargs -n 1 bazelisk run
           bazelisk query 'attr(executable, 1, kind(genrule, tests/macos/xcodeproj/...))' | xargs -n 1 bazelisk run
           git diff --exit-code tests/macos/xcodeproj
+      - name: Ensure hmap files do not exist (or remove if they do)
```
segiddins, line 54: why is this necessary? should we instead put this logic in `./tests/macos/xcodeproj/build.sh` so it can be used when running locally?

tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -euxo pipefail
 
-# TODO: If the target is being built from the rules_ios repo, then remove the `@build_bazel_rules_ios`
-# prefix to the `local_debug_options_enabled` build flag
-$BAZEL_PATH build --experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME --build_event_text_file=$BAZEL_BUILD_EVENT_TEXT_FILENAME --build_event_publish_all_actions $1 --@build_bazel_rules_ios//rules:local_debug_options_enabled $BAZEL_DEBUG_SYMBOLS_FLAG 2>&1 | $BAZEL_OUTPUT_PROCESSOR
+$BAZEL_PATH build \
+--experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME \
```
segiddins, line 5: nit: indent subsequent lines so it's visually clear they're part of a single command invocation

.github/workflows/tests.yml
```diff
@@ -64,10 +65,17 @@ jobs:
       - name: Generate Xcode projects and compare against existing record
         run: |
           set -euo pipefail
+          bazelisk clean
           bazelisk query 'kind(xcodeproj, tests/ios/xcodeproj/...)' | xargs -n 1 bazelisk run
           bazelisk query 'attr(executable, 1, kind(genrule, tests/ios/xcodeproj/...))' | xargs -n 1 bazelisk run
           git diff --exit-code tests/ios/xcodeproj
+      - name: Ensure hmap files do not exist (or remove if they do)
+        run: |
+          grep "HEADER_SEARCH_PATHS" tests/ios/xcodeproj/*.xcodeproj/project.pbxproj | grep -o  "bazel-out\S*\.hmap" | xargs -I {} bash -c 'F=$0; echo $F; if [ -f $F ]; then rm $F; echo "removed file at $F"; fi' "{}"
```
amberdixon, line 74: I think this was brought by Sam earlier ... but can you move these steps into the build.sh ? Thanks.

.bazelrc
amberdixon: In its current form, this PR has removed `compilation_mode=dbg` from the build-wrapper. Could you put it back? You can add it into the new `BAZEL_RULES_IOS_OPTIONS` build setting you have created and are using.

.github/workflows/tests.yml
amberdixon: Also, I think these tests should all be referring to FRAMEWORK_SEARCH_PATHS, not HEADER_SEARCH_PATHS.

This is because in the lldb-settings.sh, we tell LLDB about the $FRAMEWORK_SEARCH_PATHS.

.bazelrc
gyfelton: we have discussed and agree that above comment is invalid as now the single point of truth is on user's .bazelrc file @amberdixon

rules/transition_support.bzl
gyfelton: done

gyfelton:
@segiddins @amberdixon should be ready to review again 🙏

.github/workflows/tests.yml
gyfelton: new test case will fail if changes are not in :)

.github/workflows/tests.yml
gyfelton: for now it does the same thing of checking against the hash value to see if what xcodeproj wants is what bazel build generates inside `build-warapper.sh`
This test case will fail if changes in this PR is reverted (the transitive rule) so i do see value in this test

.github/workflows/tests.yml
```diff
@@ -64,10 +65,15 @@ jobs:
       - name: Generate Xcode projects and compare against existing record
         run: |
           set -euo pipefail
+          bazelisk clean
           bazelisk query 'kind(xcodeproj, tests/ios/xcodeproj/...)' | xargs -n 1 bazelisk run
           bazelisk query 'attr(executable, 1, kind(genrule, tests/ios/xcodeproj/...))' | xargs -n 1 bazelisk run
           git diff --exit-code tests/ios/xcodeproj
+      - name: Run pre build checks
+        run: ./tests/ios/xcodeproj/pre_build_check.sh
```
segiddins, line 73: why not include these in the existing `build` shell script?

gyfelton: What's added are not part of build process (aka removing them won't affect build). might as well use a pre_build and post_build hooks for future iteration on such checks

PR #103, opened 2020-08-19
Disable swift module caching in xcodeprojects for now until we figure out how to get debugging working

amberdixon:


segiddins:
Do the changed options also need to go into the transition so paths are correct?

amberdixon:
> Do the changed options also need to go into the transition so paths are correct?

I'll update this PR, although I don't think that the SHA is changing when these flags are added.

amberdixon:
> > Do the changed options also need to go into the transition so paths are correct?
> 
> I'll update this PR, although I don't think that the SHA is changing when these flags are added.

I tried updating the transition. First of all, I don't think it needs to be updated, because the SHA of the bazel-out directory does not seem to change.

Second of all, I did try making the change  to the transition as follows and it caused the SHA of bazel-out directory referenced by Framework_search_paths to point to a non-existent dir:
```
+++ b/rules/transition_support.bzl
@@ -120,15 +120,17 @@ _apple_rule_transition = transition(
 
 def _force_swift_local_debug_options_transition_impl(settings, attr):
     _ignore = (settings, attr)
-    return [
-        {"@build_bazel_rules_ios//rules:local_debug_options_enabled": True},
-    ]
+    return  {
+      "@build_bazel_rules_ios//rules:local_debug_options_enabled": True,
+      "//command_line_option:features": ["-swift.cacheable_swiftmodules", "-swift.use_global_module_cache"]
+    }
 
 _force_swift_local_debug_options_transition = transition(
     implementation = _force_swift_local_debug_options_transition_impl,
     inputs = [],
     outputs = [
         "@build_bazel_rules_ios//rules:local_debug_options_enabled",
+        "//command_line_option:features",
     ],
 )
```

rules/xcodeproj.bzl
```diff
@@ -282,7 +282,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_BUILD_EXEC": "$BAZEL_STUBS_DIR/build-wrapper",
         "BAZEL_OUTPUT_PROCESSOR": "$BAZEL_STUBS_DIR/output-processor.rb",
         "BAZEL_PATH": ctx.attr.bazel_path,
-        "BAZEL_RULES_IOS_OPTIONS": "--@build_bazel_rules_ios//rules:local_debug_options_enabled",
+        "BAZEL_RULES_IOS_OPTIONS": "--@build_bazel_rules_ios//rules:local_debug_options_enabled --features=-swift.cacheable_swiftmodules --features=-swift.use_global_module_cache",
```
gyfelton, line 285: will this cause `bazel run` producing a different path than `bazel build` just like before? (the tests are now not doing anything btw)

amberdixon: It was my observation that this does not change the path from what it was before.

Spoke with gyfelton offline about the tests ... they are doing almost the same thing as before, except for they are now looking for framework search paths (instead of header search paths.)

tests/ios/xcodeproj/pre_build_check.sh
```diff
@@ -2,14 +2,12 @@ set -eux
 
 cd $(dirname $0)
 
-export HEADER_SERACH_PATHS=`grep "HEADER_SEARCH_PATHS" *.xcodeproj/project.pbxproj | grep -o  "bazel-out\S*\.hmap"`
-
-echo "Ensure hmap files do not exist (remove if they do)"
-for path in ${HEADER_SERACH_PATHS}; do
+export FSP=`grep "FRAMEWORK_SEARCH_PATHS" *.xcodeproj/project.pbxproj | grep -o "bazel-out[^ \\]*"`
```
gyfelton, line 5: i wonder why not just do both? (hmap paths and framework paths)

gyfelton:
my feeling is that transition overrides everything under the feature set? (to reply what amber observed when adding it to the transition)

amberdixon:
> my feeling is that transition overrides everything under the feature set? (to reply what amber observed when adding it to the transition)

If we explicitly pass the feature when we call `bazel build`, then the targets will be built with the feature disabled. 

The transition is an incoming edge transition to the xcodeproj rule itself. It only influences how the xcodeproj gets generated. I am finding that adding these features to the bazel build command does not change the SHA for the location of the generated frameworks. Therefore, I think we don't need to put these values in the transition when we generate xcodeproj.

PR #104, opened 2020-08-20
Fix bad query from framework causing disk-wasting rsync operation

gyfelton:
### What happened
In this PR (https://github.com/bazel-ios/rules_ios/pull/93/files) a change is added to installer script to obtain path to framework as input and copy it over to another location. However since `rsync` can be a dangerous operation where it might try to copy entire disk or root dir over to another location, it's important to do some basic sanity checks.
This PR introduced such basic checks, please suggest some better ways to validate the input. Also given that current problem arises from framework's `QUERY`, it has a strict check of `not empty` so that in future dev can figure out which target broke this sanity check

###Priority
High


tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
```diff
@@ -41,6 +45,11 @@ output="$TARGET_BUILD_DIR/$FULL_PRODUCT_NAME"
 mkdir -p "$(dirname "$output")"
 
 for input in "${input_options[@]}"; do
+    if [[ -z $input ]] || [ $input = "." ] || [ $input = "/" ]; then
+        # rsync can be a dangerous operation when it tries to copy entire root dir into the $output
+        echo "Error: illegal input for installing ${TARGET_NAME} of type ${PRODUCT_TYPE}" >&2
```
segiddins, line 50: nit: include the bad input in the error message?

PR #105, opened 2020-08-20
New tag to help ignore targets, also checks to make sure no target overwrite with different type

gyfelton:
### Why this change
Due to the need to have a "framework_unlinked" version of a app so that tests can run, we end up having a framework target with name `Foo` while there is an app target also with name `Foo`. If we don't ignore one of them, during `xcodeproj` generation phase one will overwrite the other.

This PR is therefore putting efforts to prevent this from happening:
1. a new tag so dev can specify to ignore one of them
2. a check to prevent such thing from happening

### Tests done
The new xcodeproj rule is testing against above situation so without the tag, new test cases will fail during `bazel run <some_xcodeproj_rule>` stage

### Outstanding question
1. Where to document this new tag?

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -26,6 +26,8 @@ _ARCH_MAPPING = {
 
 _PRODUCT_SPECIFIER_LENGTH = len("com.apple.product-type.")
 
+_IGNORE_AS_TARGET_TAG = "xcodeproj-ignore-as-target"
+
 def _dir(o):
     return [
         x
@@ -134,7 +136,6 @@ def _xcodeproj_aspect_impl(target, ctx):
             env_vars = env_vars,
             hmap_paths = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "hmap_paths")),
             commandline_args = commandline_args,
-            tags = tags,
         )
         if ctx.rule.kind != "apple_framework_packaging":
             providers.append(
@@ -150,10 +151,16 @@ def _xcodeproj_aspect_impl(target, ctx):
                     hmap_paths = depset(hmap_paths),
                 ),
             )
-        direct_targets = [info]
+
+        direct_targets = []
+        transitive_targets = []
+        if not _IGNORE_AS_TARGET_TAG in tags:
+            direct_targets.append(info)
+            transitive_targets.append(info)
+
         if test_host_target:
             direct_targets.extend(test_host_target[_TargetInfo].direct_targets)
-        target_info = _TargetInfo(direct_targets = direct_targets, targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
+        target_info = _TargetInfo(direct_targets = direct_targets, targets = depset(transitive_targets, transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
         providers.append(target_info)
     else:
         srcs = []
@@ -328,16 +335,14 @@ def _xcodeproj_impl(ctx):
     for target_info in targets:
         target_name = target_info.name
 
-        if "xcodeproj-ignore-as-target" in target_info.tags:
-            continue
-
         if target_name in xcodeproj_targets_by_name:
             existing_type = xcodeproj_targets_by_name[target_name]["type"]
             if target_info.product_type != existing_type:
-                fail("""
+                fail("""\
+Failed to generate xcodeproj for "{}" due to conflicting targets:
 Target "{}" is defined with type "{}", but a same-name target of type "{}" wants to override.
-Please double check your rule declaration or add `xcodeproj-ignore-as-target` as a tag to choose ignore which target.
-""".format(target_name, existing_type, target_info.product_type))
+Double check your rule declaration for naming or add `xcodeproj-ignore-as-target` as a tag to choose which target to ignore.
+""".format(ctx.label, target_name, existing_type, target_info.product_type))
 
         target_macho_type = "staticlib" if target_info.product_type == "framework" else "$(inherited)"
         compiled_sources = [{
@@ -543,6 +548,11 @@ $BAZEL_INSTALLER
 
 xcodeproj = rule(
     implementation = _xcodeproj_impl,
+    doc = """\
+Generates a XCode project file (.xcodeproj) with a reasonable set of defaults
+Tags for configuration:
+    xcodeproj-ignore-as-target: Add this to a rule declaration so that this rule will not generates a scheme for this target
+""",
     cfg = transition_support.force_swift_local_debug_options_transition,
     attrs = {
         "deps": attr.label_list(mandatory = True, allow_empty = False, providers = [], aspects = [_xcodeproj_aspect]),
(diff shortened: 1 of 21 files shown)


rules/xcodeproj.bzl
```diff
@@ -132,6 +134,7 @@ def _xcodeproj_aspect_impl(target, ctx):
             env_vars = env_vars,
             hmap_paths = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "hmap_paths")),
             commandline_args = commandline_args,
+            tags = tags,
```
segiddins, line 137: instead of propogating the tags, could we skip adding a `direct` entry into the depset when `xcodeproj-ignore-as-target` is in the tags? that theoretically should let us short-circuit some work

rules/xcodeproj.bzl
```diff
@@ -323,6 +326,19 @@ def _xcodeproj_impl(ctx):
...
+        target_name = target_info.name
+
+        if "xcodeproj-ignore-as-target" in target_info.tags:
+            continue
+
+        if target_name in xcodeproj_targets_by_name:
+            existing_type = xcodeproj_targets_by_name[target_name]["type"]
+            if target_info.product_type != existing_type:
+                fail("""
+Target "{}" is defined with type "{}", but a same-name target of type "{}" wants to override.
```
segiddins, line 338: can we get the full labels of the conflicting targets in the error message somehow?

rules/xcodeproj.bzl
```diff
@@ -323,6 +326,19 @@ def _xcodeproj_impl(ctx):
     xcodeproj_schemes_by_name = {}
 
     for target_info in targets:
+        target_name = target_info.name
+
+        if "xcodeproj-ignore-as-target" in target_info.tags:
```
amberdixon, line 331: What if we just skipped all targets that had the `manual` tag? If we don't want a target to show up from doing a `bazel query`, do we really want it to show up their xcodeproject? I would guess not.

rules/xcodeproj.bzl
gyfelton: added a line before this to indicate which label we are in

rules/xcodeproj.bzl
```diff
@@ -147,10 +151,16 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+        direct_targets = []
+        transitive_targets = []
+        if not _IGNORE_AS_TARGET_TAG in tags:
+            direct_targets.append(info)
+            transitive_targets.append(info)
+
         if test_host_target:
             direct_targets.extend(test_host_target[_TargetInfo].direct_targets)
-        target_info = _TargetInfo(direct_targets = direct_targets, targets = depset([info], transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
+        target_info = _TargetInfo(direct_targets = direct_targets, targets = depset(transitive_targets, transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
```
gyfelton, line 163: @segiddins are above what you have in mind? I have to do the same to `transitive_targets ` because the sanity check I added fails if `include_transittive_targets` is set to True, I am adding a new xcodeproj rule for Test project where it is set to True but it fails current checks on hmap and framework search path. I need to figure out what's going on then

segiddins: Yes, this is roughly what I had in mind

rules/xcodeproj.bzl
```diff
@@ -147,10 +151,16 @@ def _xcodeproj_aspect_impl(target, ctx):
                     hmap_paths = depset(hmap_paths),
                 ),
             )
-        direct_targets = [info]
+
+        direct_targets = []
+        transitive_targets = []
+        if not _IGNORE_AS_TARGET_TAG in tags:
```
segiddins, line 157: nit: `if x not in y:`

rules/xcodeproj.bzl
segiddins: but what about the labels of the iOS targets that the error is saying are conflicting?

rules/xcodeproj.bzl
```diff
@@ -527,6 +548,11 @@ $BAZEL_INSTALLER
 
 xcodeproj = rule(
     implementation = _xcodeproj_impl,
+    doc = """\
+Generates a XCode project file (.xcodeproj) with a reasonable set of defaults
```
segiddins, line 552: sp: `Xcode`

tests/ios/xcodeproj/post_build_check.sh
```diff
@@ -1,4 +1,4 @@
-set -eux
+set -e
```
segiddins, line 1: why this change?

tests/ios/xcodeproj/build.sh
```diff
@@ -5,7 +5,9 @@ cd $(dirname $0)
...
 xcrun simctl list
 
-export PROJECT_AND_SCHEME="-project Single-Static-Framework-Project.xcodeproj -scheme ObjcFrameworkTests"
-export SIM_DEVICE_ID=`xcodebuild $PROJECT_AND_SCHEME -showdestinations | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last"`
+export SAMPLE_PROJECT_AND_SCHEME="-project Single-Static-Framework-Project.xcodeproj -scheme ObjcFrameworkTests"
+export SIM_DEVICE_ID=`xcodebuild $SAMPLE_PROJECT_AND_SCHEME -showdestinations | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last"`
 
-xcodebuild $PROJECT_AND_SCHEME -destination "id=$SIM_DEVICE_ID" -quiet 
\ No newline at end of file
+for i in ` find *.xcodeproj -maxdepth 0 -type d`; do
```
segiddins, line 11: nit: extra space after "`"

tests/ios/xcodeproj/post_build_check.sh
gyfelton: i don't think these 2 scripts should output all the execution details, especially when the list of file paths to search can be long. Usually this script output nothing unless there is a path not found

segiddins: still would want `u` then, so we error on unset variables?

tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -11,6 +11,18 @@ xcodeproj(
     ],
 )
 
+xcodeproj(
+    name = "Test-Imports-App-Project",
+    testonly = True,
+    bazel_path = "bazelisk",
+    include_transitive_targets = False,
```
gyfelton, line 18: Current post-build check will fail on `include_transitive_targets` being set to `True` because there are search paths are not found. I am planning to make this into an issue under github and deal with it later in a separate PR, what do you folks think? @segiddins @amberdixon

segiddins: if it's not a regression, feel free to make an issue for it and fast-follow

tests/ios/xcodeproj/post_build_check.sh
gyfelton: got it!

gyfelton:
@segiddins @amberdixon another look? 🙏

PR #107, opened 2020-08-25
Added a flag to disable logging of execution log from Bazel by default

gyfelton:
### Why this change
Execution log file from bazel can eat a lot of disk space and is not useful when builds are working.
Only when it is not working, where engineer can ask the end user of this repo to change the `BAZEL_EXECUTION_LOG_ENABLED` inside xcode to 1 (at the highest level possible, usually project level) so that during bazel build, the execution log will be dumped to the specified path

### Tests done
no test since it can be easily verified manually.

### Priority
High, as it impacts dev's disk space

### Opent question
Should I document this user defined flag inside xcodeproj somewhere?


amberdixon:
> Should I document this user defined flag inside xcodeproj somewhere?
IMO that's not necessary.

PR #110, opened 2020-08-27
Copy when “merging” a single binary

segiddins:
(a) this saves a libtool invocation
(b) it ensures that input & output can share storage in the CAS cache (as e.g. libtool might change the timestamp due to -D, which isn’t present for ld invocations in all versions of the c++ crosstool)

### DIFF
--- a/rules/framework/framework_packaging.py
+++ b/rules/framework/framework_packaging.py
@@ -60,13 +60,18 @@ def _merge_binaries(framework_root, framework_name, binary_in):
         binary_in: a list of binary paths
     """
     output_path = os.path.join(framework_root, framework_name)
-    commands = ["libtool",
-                "-static",
-                "-D",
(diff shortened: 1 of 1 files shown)


rules/framework/framework_packaging.py
```diff
@@ -60,13 +60,18 @@ def _merge_binaries(framework_root, framework_name, binary_in):
...
     """
     output_path = os.path.join(framework_root, framework_name)
-    commands = ["libtool",
-                "-static",
-                "-D",
-                "-no_warning_for_no_symbols",
-                "-o", output_path
-                ] + binary_in
-    subprocess.check_call(commands)
+    if len(binary_in) == 1:
```
gyfelton, line 63: do we have sample project with a single binary?

segiddins: anything pure objc will only have a single archive

PR #111, opened 2020-08-27
Add support for .xcassets and .xcstickers (catalogue groups)

gyfelton:
Also refactored code around xcdatamodel a bit to match new changes
What looks like now:
Screen Shot 2020-08-27 at 2 47 00 PM


rules/xcodeproj.bzl
```diff
@@ -310,10 +339,22 @@ def _gather_asset_sources(target_info, path_prefix):
                 "buildPhase": "none",
             }
             asset_sources.append(payload)
-    for datamodel_short_path in datamodel_groups.keys():
+
+    for datamodel_key in datamodel_groups.keys():
```
segiddins, line 343: could use `for (datamodel_key, datamodel_path) in datamodel_groups.items():`

rules/xcodeproj.bzl
```diff
@@ -280,28 +280,57 @@ def _gather_asset_sources(target_info, path_prefix):
...
-
-    # Each key will be a short_path to a directory ending either .xcdatamodeld or .xcdatamodel
     datamodel_groups = {}
+    catalog_groups = {}
+
     for s in target_info.asset_srcs.to_list():
         short_path = s.short_path
         group = paths.dirname(short_path)
+        short_path_components = short_path.split("/")
+        is_xcassets = False
```
segiddins, line 290: instead of 5 different boolean variables,  maybe its simpler to store the extension? also, probably makes sense to extract this out into a `_classify_asset` helper function?

rules/xcodeproj.bzl
```diff
@@ -310,10 +339,22 @@ def _gather_asset_sources(target_info, path_prefix):
...
+        datamodel_path = datamodel_groups[datamodel_key]
+        payload = {
+            "path": paths.join(path_prefix, datamodel_path),
+            "group": paths.dirname(datamodel_path),
+            "optional": True,
+            "buildPhase": "none",
+        }
+        asset_sources.append(payload)
+
+    for asset_key in catalog_groups.keys():
```
segiddins, line 353: same as above, iterate over the `items()` tuples

amberdixon:
In general, why do we filter `xcdatamodel` and `xcasset` files out from the data passed to objc_library? Is it because they are not needed at runtime?

rules/xcodeproj.bzl
```diff
@@ -280,28 +280,57 @@ def _gather_asset_sources(target_info, path_prefix):
...
+        elif is_xcdatamodeld or is_xcdatamodel or is_xcmappingmodel:
+            # Any file under .xcdatamodeld or .xcdatamodel is ignored but itself will be added.
+            # However there two possibilities for .xcdatamodel to exist:
+            # 1. foo/bar.xcdatamodel (standalone)
+            # 2. foo/foo2.xcdatamodeld/bar(v2).xcdatamodel (versioned datadmodel)
+            # Iteration of path components above stops at xcdatamodeld first
+            # so the second case is taken care of
+            datamodel_name = paths.basename(path_so_far)
+            if datamodel_name not in datamodel_groups:
+                datamodel_groups[datamodel_name] = path_so_far
```
amberdixon, line 332: I think this might be backwards, right? It should be:

```
datamodel_groups[path_so_far] = datamodel_name
```

or alternatively:
```
datamodel_groups[paths.dirname(path_so_far)] = paths.basename(path_so_far)
```

amberdixon: I think there's the same problem with the settings of key-value pairs in catalog_groups dictionary.

gyfelton: this approach avoids a situation where same file name sittting in diff. folder are considered as the same asset? Good point!

rules/xcodeproj.bzl
gyfelton: keeping as is since the values are not that important now, now use dict just to de-duplciate the keys)

gyfelton:
@amberdixon another look? changed to make key = unique path to avoid the situation when assets with same name but different path got ignored for one of them

amberdixon:
> @amberdixon another look? changed to make key = unique path to avoid the situation when assets with same name but different path got ignored for one of them

Updates look good, thanks!

rules/xcodeproj.bzl
```diff
@@ -269,6 +269,43 @@ def _joined_header_search_paths(hmap_paths):
     header_search_paths.append("\"$BAZEL_WORKSPACE_ROOT\"")
     return " ".join(header_search_paths)
 
+_XCASSETS = "xcassets"
+_XCDATAMODELD = "xcdatamodeld"
+_XCDATAMODEL = "xcdatamodel"
+_XCMAPPINGMODEL = "xcmappingmodel"
+_XCSTICKERS = "xcstickers"
+
+def _classify_asset(path):
```
segiddins, line 278: 🎉

Issue #113, opened 2020-08-28
Error "this class is not key value coding-compliant for the key" when using XIBs inside resource bundles


acecilia:
Hi:

I have been trying to debug this issue, but with little success. 

What I did:
1. Take the example `tests/ios/frameworks/target-xib` from this repository and copy it to `tests/ios/frameworks/target-xib-resource-bundle`
2. Modify the `BUILD.bazel` file to use a resource bundle
3. Update the source code to point to the right resource bundle when reading the XIB

When running the test from the example, the resource bundle is found, and the XIB is also found. But after that I am getting the error: `<unknown>:0: error: -[TargetXIBTestsLib.CustomViewTests testCustomView] : failed: caught "NSUnknownKeyException", "[<UIView 0x7f8d9fc6de50> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key label."`

Find the code with the failing test here: https://github.com/bazel-ios/rules_ios/pull/112
To try it out, just run `git clone https://github.com/acecilia/rules_ios.git -b customModuleProviderBug && cd rules_ios && bazelisk test --apple_platform_type=ios //tests/ios/frameworks/target-xib-resource-bundle:TargetXIBTests`

Thanks :)

acecilia:
Issue may be related with the call to `ibtool`. From `man ibtool`:

```
       --module moduleName
              Specifies  the  module  name  to include in applicable custom class names at runtime. The module name should match the PRODUCT_MODULE_NAME build
              setting of the target for which the Interface Builder document is being compiled. This option is required for compiling any document  that  con-
              tains an object whose custom class is a Swift class and whose customModuleProvider attribute is set to target.
```

Reference: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleBundleDescription.java#L435

segiddins:
We don't use `customModuleProvider` so I'm not surprised it's not working -- probably need to plumb the module name through to the precompiled_resource_bundle target somehow

acecilia:
Some more information, to be used as a future reference:

* The command line tool used for compiling storyboards, nibs and xibs is `ibtool`. Running `man ibtool` it is possible to see that the swift module is passed as a parameter:
    ```
    --module moduleName
        Specifies the module name to include in applicable custom class  names  at  runtime.  The  module  name
        should  match the PRODUCT_MODULE_NAME build setting of the target for which the Interface Builder docu-
[...]

PR #114, opened 2020-08-29
Fix the precompiled_apple_resource_bundle to use a correct swift module

acecilia:
Fixes https://github.com/bazel-ios/rules_ios/issues/113

I couldnt find a way to properly pass the `swift_module` parameter to the `xibCompile` step, so instead I changed the label for the `fake_ctx` (which seems to be non important) to match the required `swift_module`: in the end, such label is used as a fallback value when `swift_module` is not specified.

You can see the fallback here: https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/support/resources_support.bzl#L446

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -51,6 +51,19 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         fake_ctx_dict[k] = getattr(ctx, k)
     fake_ctx_dict["attr"] = struct(**attr_dict)
     fake_ctx_dict["file"] = struct(**file_dict)
+
+    # The label of this fake_ctx is used as the swift module associated with storyboards, nibs and xibs.
+    # See: https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/support/resources_support.bzl#L446
+    #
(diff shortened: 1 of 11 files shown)


rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -51,6 +51,19 @@ def _precompiled_apple_resource_bundle_impl(ctx):
...
+    #
+    # Such swift module is required when the storyboard, nib or xib contains the value <customModuleProvider="target">.
+    # Otherwise, the swift module is not important and could be any arbitrary string.
+    # For the full context see https://github.com/bazel-ios/rules_ios/issues/113
+    #
+    # Usage:
+    # The most common scenario happens when the bundle name is the same as the corresponding swift module.
+    # If that is not the case, it is possible to customize the swift module by explicitly
+    # passing a swift_module attr
+    fake_ctx_dict["label"] = Label("//fake_package:" + (ctx.attr.swift_module or bundle_name))
```
acecilia, line 66: @keith Do you know if there is any better way of passing the `swift_module`?

keith: we haven't ever tried since we don't rely on this being set for anything

PR #116, opened 2020-09-04
Update to Bazel 3.5 & HEAD of upstream rules

segiddins:


.bazelrc
```diff
@@ -1,3 +1,7 @@
+# Required for bazel 3.4+
+# See https://github.com/bazelbuild/bazel/issues/10674#issuecomment-658270895
+build --incompatible_objc_compile_info_migration
```
gyfelton, line 3: damn this migration thing keep coming up :(

rules/test_host_app/BUILD.bazel
```diff
@@ -7,17 +7,17 @@ exports_files([
 
 [
     ios_application(
-        name = "iOS-%s-AppHost" % version,
```
gyfelton, line 10: want to learn why this change is needed, no more support of this `%` syntax?

rules/repositories.bzl
```diff
@@ -93,10 +93,24 @@ native_binary(
     out = "index-import",
     visibility = ["//visibility:public"],
 )
+
+native_binary(
+    name = "validate_index",
```
gyfelton, line 98: want to learn what's this from (new bazel versions' requirement?)

gyfelton (approved):

.bazelrc
segiddins: It's a good thing though! the objc provider is bad, and consolidation with CcInfo will make things simpler in the long run

rules/test_host_app/BUILD.bazel
segiddins: No it's still supported, but using the explicit `format` method reads clearer

rules/repositories.bzl
segiddins: No, not required, but will make debugging using these tools easier since now we can just `bazel run` them (they were added with the bump to index-import)

.bazelrc
gyfelton: totally agree!

rules/test_host_app/BUILD.bazel
gyfelton: i see i thought this is a better syntax, will use `format` in future instead of this

justinseanmartin (approved):

WORKSPACE
```diff
@@ -46,9 +46,9 @@ load(
 
 git_repository(
     name = "io_bazel_stardoc",
-    commit = "4378e9b6bb2831de7143580594782f538f461180",
-    remote = "https://github.com/bazelbuild/stardoc.git",
-    shallow_since = "1570829166 -0400",
+    commit = "fa1878a1c4eae78a8aff016500777bedbc0bdf32",
+    remote = "https://github.com/segiddins/stardoc.git",
```
amberdixon, line 50: Isn't bazelbuild master up to date with the segiddins branch? (Assuming this is true, because of https://github.com/bazelbuild/stardoc/compare/master...segiddins:master) Can we point to bazelbuild master instead?

segiddins: no, stardoc hasn't merged my PR yet

Issue #106, opened 2020-08-24
Post build script fails if `include_transitive_targets` is set to True


gyfelton:
Following PR https://github.com/bazel-ios/rules_ios/pull/105:
if we have this where include_transitive_targets is set to True, post build check on this project will fail
```
xcodeproj(
    name = "Test-Imports-App-Project",
    testonly = True,
    bazel_path = "bazelisk",
    include_transitive_targets = True,
    deps = [
        "//tests/ios/unit-test/test-imports-app:TestImports-App",
        "//tests/ios/unit-test/test-imports-app:TestImports-App_framework_unlinked",
        "//tests/ios/unit-test/test-imports-app:TestImports-Unit-Tests",
    ],
)
```
Not a regression as this uncovers an existing issue

PR #117, opened 2020-09-07
Some fixes to the xcodeproj rule

acecilia:
Fix xcodeproj: 
* Pass a custom name correctly
* Add the host app to the project when adding the transitive dependencies (seems it was forgotten)

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -159,7 +159,11 @@ def _xcodeproj_aspect_impl(target, ctx):
             transitive_targets.append(info)
 
         if test_host_target:
-            direct_targets.extend(test_host_target[_TargetInfo].direct_targets)
+            # Add the test_host_target to the targets to be added to the xcode project
+            test_host_direct_targets = test_host_target[_TargetInfo].direct_targets
+            direct_targets.extend(test_host_direct_targets)
+            transitive_targets.extend(test_host_direct_targets)
+
         target_info = _TargetInfo(direct_targets = direct_targets, targets = depset(transitive_targets, transitive = _get_attr_values_for_name(deps, _TargetInfo, "targets")))
         providers.append(target_info)
     else:
@@ -384,7 +388,7 @@ def _xcodeproj_impl(ctx):
     xcodegen_jsonfile = ctx.actions.declare_file(
         "%s-xcodegen.json" % ctx.attr.name,
     )
-    project_name = ctx.attr.project_name if ctx.attr.project_name else ctx.attr.name + ".xcodeproj"
+    project_name = (ctx.attr.project_name or ctx.attr.name) + ".xcodeproj"
     if "/" in project_name:
         fail("No / allowed in project_name")
 
--- a/tests/ios/unit-test/BUILD.bazel
+++ b/tests/ios/unit-test/BUILD.bazel
@@ -29,6 +29,8 @@ ios_unit_test(
     ],
     minimum_os_version = "12.0",
     test_host = "//rules/test_host_app:iOS-9.3-AppHost",
+    # Adding visibility so the xcodeproject tests can depend on this target
+    visibility = ["//visibility:public"],
 )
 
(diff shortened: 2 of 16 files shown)


segiddins:
cc @amberdixon / @gyfelton to review

Would you mind adding a test that covers the issues you're addressing? Thanks!

acecilia:
Test added :)

acecilia:
I would like to get you opinion on the following, if you guys dont mind :)

I see that the `xcodeproj` rule generates targets that build with bazel. I was thinking about tweaking the rule, so it is possible to also generate normal targets that run with `xcodebuild` (control it with a parameter). Either have one project with `MyTarget-bazel` and `MyTarget-xcodebuild`, or two projects `MyProject-bazel.xcodeproj` and `MyProject-xcodebuild.xcodeproj`. What I would like to do is to easily be able switch between bazel and xcodebuild. Seems like something that should be not so difficult to do. What do you think about this?

gyfelton (approved):
Thanks for covering this case! Yeah I think we generally lack of tests when include_transitive_targets = true, I have another issue to track a similar problem: https://github.com/bazel-ios/rules_ios/issues/106

gyfelton:
> I would like to get you opinion on the following, if you guys dont mind :)
> 
> I see that the `xcodeproj` rule generates targets that build with bazel. I was thinking about tweaking the rule, so it is possible to also generate normal targets that run with `xcodebuild` (control it with a parameter). Either have one project with `MyTarget-bazel` and `MyTarget-xcodebuild`, or two projects `MyProject-bazel.xcodeproj` and `MyProject-xcodebuild.xcodeproj`. What I would like to do is to easily be able to switch between bazel and xcodebuild. Seems like something that shouldn't be so difficult to do. What do you think about this? Did you try something like this before?

I say that this should be inside a brand new rule that its only job is to generate a normal xcodeproj (via XcodeGen).

acecilia:
> I say that this should be inside a brand new rule that its only job is to generate a normal xcodeproj (via XcodeGen).

I see, thanks :) In that case, how would you share code between the `xcodeproj-bazel` rule and the `xcodeproj-xcodebuild` rule? My initial thought was that both would be very similar

amberdixon (approved):
Thanks for addressing these issues!

amberdixon:
> > I say that this should be inside a brand new rule that its only job is to generate a normal xcodeproj (via XcodeGen).
> 
> I see, thanks :) In that case, how would you share code between the `xcodeproj-bazel` rule and the `xcodeproj-xcodebuild` rule? My initial thought was that both would be very similar: since files, targets and dependencies are shared, seems like the only differences would be a bunch of build settings

This is a good idea, @gyfelton . As far as implementing it, I think it would be pretty simple, the xcodeproject could just contain targets and schemes that match the ones listed in the bazel project and with the same set of source files.

gyfelton:
> > > I say that this should be inside a brand new rule that its only job is to generate a normal xcodeproj (via XcodeGen).
> > 
> > 
> > I see, thanks :) In that case, how would you share code between the `xcodeproj-bazel` rule and the `xcodeproj-xcodebuild` rule? My initial thought was that both would be very similar: since files, targets and dependencies are shared, seems like the only differences would be a bunch of build settings
> 
> This is a good idea, @gyfelton . As far as implementing it, I think it would be pretty simple, the xcodeproject could just contain targets and schemes that match the ones listed in the bazel project and with the same set of source files.

I worried more about the maintenances needed if sharing of code is done, specifically each time when we are adding a new build setting/changing behaviors, we have to verify the correctness on both sides. Nevertheless if this enforces a better separation of logics inside existing rule (what is for a normal project, and what is for bazel), I am definitely in favor of it.

acecilia:
Is this ready to merge? :)

gyfelton:
@acecilia go ahead!

gyfelton:
@acecilia did not realize you cannot merge it, I merged. Thanks!

Issue #108, opened 2020-08-26
Local breakpoints do not work when sandbox is enabled


amberdixon:
See https://github.com/bazel-ios/rules_ios/pull/98#issue-463041614 for context.

"In order for the example applications and tests in the rules_ios repo to be debuggable from Xcode, you must:

Disable sandboxing in .bazelrc.
Update the flag in the tools/xcodeproj_shims/build-wrapper.sh to refer to --//rules:local_debug_options_enabled rather than --@build_bazel_rules_ios//rules:local_debug_options_enabled"

acecilia:
Hey, I am trying to get breakpoints working on Xcode while using the `xcodeproj` rule. What I did:
* I pulled rules_ios to the most updated commit, containing https://github.com/bazel-ios/rules_ios/pull/119
* I added the following to my `bazel.rc` file:
```
build --genrule_strategy=standalone
build --spawn_strategy=standalone
build --compilation_mode=dbg
```

But I cant get it to work. Are there any other step required? 🤔

amberdixon:
Will take a look, thank you

wileykestner:
@amberdixon We tried the same setup steps as @acecilia and also reproduced the same result: we did not hit a local breakpoint when set in Xcode 11.  Were you able to reproduce @acecilia's issue?  

We would be happy to help investigate / fix if possible, but don't want to duplicate any investigative work that you have already done on this issue.

acecilia:
@amberdixon could you give an update on this? 🙏 
What is the status of breakpoints? Are they supposed to work and I have some faulty configuration, or are they not working?

amberdixon:
Hi all! This is working for me, if I do the following:
1. `bazelisk run tests/ios/xcodeproj:Test-With-Host-App --@build_bazel_rules_ios//rules:local_debug_options_enabled`
2. `open tests/ios/xcodeproj/TestWithHostApp.xcodeproj`
3. Set a break point on line 5 of empty.swift where the EmptySwiftTests class is defined.
3. Select the ExplicitHosted scheme
4. Run Cmd+U.
My breakpoint gets hit.

Note that I do not need to disable sandbox mode anymore. I think there have been some changes to this repo since October that may have fixed the issue. I'm using Xcode 11.6 on BigSur (Mac OS v11.0.1).

[...]

PR #119, opened 2020-09-09
Set clang flags in lldb, propagate linker options

amberdixon:
Debugging apps and tests with transitive swift when `cacheable_swiftmodules` was enabled was not working for us, because:
1. the LLDB swift-extra-clang-flags setting was not set, and
2. Frameworks that generated ASTs were not propagating the appropriate linker options to apps/tests that depended on them. `link_inputs` is a critical piece to propagate forward to the linker; according to the documentation (https://docs.bazel.build/versions/master/skylark/lib/ObjcProvider.html#link_inputs): "Link time artifacts from dependencies that do not fall into any other category such as libraries or archives. This catch-all provides a way to add arbitrary data (e.g. Swift AST files) to the linker. The rule that adds these is also responsible to add the necessary linker flags to 'linkopt'."

Note that when `cacheable_swiftmodules` is disabled, libraries are built with the `serialize-debugging-options` flag, so debugging works just fine. When `cacheable_swiftmodules` is enabled however, we need to give LLDB some additional information and also make sure all the dependent ASTs are linked into the final app/test binary.

The `dsymutil` command is being used in the tests to view the debugging sections of the generated app/test binary to ensure ALL necessary ASTs are linked to it.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,9 +6,6 @@ build --incompatible_objc_compile_info_migration
 # so circumvent by not using the sandbox
 build --strategy=Stardoc=standalone
 
-# Debugging does not work in sandbox mode. Uncomment these lines to turn off sandboxing.
-# build --genrule_strategy=standalone
-# build --spawn_strategy=standalone
 
 build --verbose_failures # Print the full command line for commands that failed
 build --test_output=errors # Prints log file output to the console on failure
@@ -19,5 +16,10 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 
 # Enable these features to test local and CI builds
 # when swiftmodule caching is enabled.
-# build --features=swift.cacheable_swiftmodules
-# build --features=swift.use_global_module_cache
+build --features=swift.cacheable_swiftmodules
+build --features=swift.use_global_module_cache
+build --compilation_mode=dbg
+
+# Debugging does not work in sandbox mode
+build --genrule_strategy=standalone
+build --spawn_strategy=standalone
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,6 +6,9 @@ build --incompatible_objc_compile_info_migration
 # so circumvent by not using the sandbox
 build --strategy=Stardoc=standalone
 
+# Debugging does not work in sandbox mode. Uncomment these lines to turn off sandboxing.
+# build --genrule_strategy=standalone
+# build --spawn_strategy=standalone
 
 build --verbose_failures # Print the full command line for commands that failed
 build --test_output=errors # Prints log file output to the console on failure
@@ -14,12 +17,6 @@ build --test_output=errors # Prints log file output to the console on failure
 # because it takes quite some time. They will only run on CI
 build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 
-# Enable these features to test local and CI builds
-# when swiftmodule caching is enabled.
-build --features=swift.cacheable_swiftmodules
-build --features=swift.use_global_module_cache
+# Enable dbg compilation mode in this repo, so we can test xcodeproj-built
+# binaries contain debug symbol tables.
 build --compilation_mode=dbg
-
-# Debugging does not work in sandbox mode
-build --genrule_strategy=standalone
-build --spawn_strategy=standalone
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -250,9 +250,13 @@ def _apple_framework_packaging_impl(ctx):
         "imported_library",
         "force_load_library",
         "multi_arch_linked_archives",
+        "multi_arch_linked_binaries",
+        "multi_arch_dynamic_libraries",
         "source",
         "define",
         "include",
+        "link_inputs",
+        "linkopt",
     ]:
         set = depset(
             direct = [],
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -407,7 +407,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_BUILD_EXEC": "$BAZEL_STUBS_DIR/build-wrapper",
         "BAZEL_OUTPUT_PROCESSOR": "$BAZEL_STUBS_DIR/output-processor.rb",
         "BAZEL_PATH": ctx.attr.bazel_path,
-        "BAZEL_RULES_IOS_OPTIONS": "--@build_bazel_rules_ios//rules:local_debug_options_enabled --features=-swift.cacheable_swiftmodules --features=-swift.use_global_module_cache",
+        "BAZEL_RULES_IOS_OPTIONS": "--@build_bazel_rules_ios//rules:local_debug_options_enabled",
         "BAZEL_WORKSPACE_ROOT": "$SRCROOT/%s" % script_dot_dots,
         "BAZEL_STUBS_DIR": "$PROJECT_FILE_PATH/bazelstubs",
         "BAZEL_INSTALLERS_DIR": "$PROJECT_FILE_PATH/bazelinstallers",
@@ -456,7 +456,7 @@ def _xcodeproj_impl(ctx):
             if target_info.product_type != existing_type:
                 fail("""\
 Failed to generate xcodeproj for "{}" due to conflicting targets:
-Target "{}" is already defined with type "{}". 
+Target "{}" is already defined with type "{}".
 A same-name target with label "{}" of type "{}" wants to override.
 Double check your rule declaration for naming or add `xcodeproj-ignore-as-target` as a tag to choose which target to ignore.
 """.format(ctx.label, target_name, existing_type, target_info.bazel_build_target_name, target_info.product_type))
@@ -510,6 +510,10 @@ Double check your rule declaration for naming or add `xcodeproj-ignore-as-target
         target_settings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = " ".join(
             ["\"%s\"" % d for d in defines_without_equal_sign],
         )
+        target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(
+            ["-D%s" % d for d in target_info.cc_defines.to_list()],
+        )
+
 
         if target_info.product_type == "application":
             target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -514,7 +514,6 @@ Double check your rule declaration for naming or add `xcodeproj-ignore-as-target
             ["-D%s" % d for d in target_info.cc_defines.to_list()],
         )
 
-
         if target_info.product_type == "application":
             target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
             target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
--- a/tools/xcodeproj_shims/installers/lldb-settings.sh
+++ b/tools/xcodeproj_shims/installers/lldb-settings.sh
@@ -20,21 +20,17 @@ settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
 END
 
-set +u
-if [[ -n $BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS ]]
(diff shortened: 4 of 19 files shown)


rules/xcodeproj.bzl
```diff
@@ -506,6 +506,10 @@ Double check your rule declaration for naming or add `xcodeproj-ignore-as-target
         target_settings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = " ".join(
             ["\"%s\"" % d for d in defines_without_equal_sign],
         )
+        target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(
```
gyfelton, line 509: this is the key for getting swift to work again right?

amberdixon: There are two issues: (1) LLDB settings need to be right or else LLDB will dump a bunch of errors and (2) the linker command for the application/test needs to include the `add_ast_path` option. This linker option needs to get propagated down to the application/test target by setting `linkopts` in the ObjcProvider returned from the framework target for the dependency.

rules/framework.bzl
```diff
@@ -250,9 +250,13 @@ def _apple_framework_packaging_impl(ctx):
         "imported_library",
         "force_load_library",
         "multi_arch_linked_archives",
+        "multi_arch_linked_binaries",
+        "multi_arch_dynamic_libraries",
         "source",
         "define",
         "include",
+        "link_inputs",
+        "linkopt",
```
amberdixon, line 259: Here is the reference for objcprovider: https://docs.bazel.build/versions/master/skylark/lib/ObjcProvider.html

.bazelrc
```diff
@@ -6,9 +6,6 @@ build --incompatible_objc_compile_info_migration
 # so circumvent by not using the sandbox
 build --strategy=Stardoc=standalone
 
-# Debugging does not work in sandbox mode. Uncomment these lines to turn off sandboxing.
-# build --genrule_strategy=standalone
```
amberdixon, line 10: I left these in here by accident, will revert.

rules/xcodeproj.bzl
gyfelton: thx for this! I should have caught them on the description section. 
Looking back, I wonder how can we know these are the steps needed without Sam's input...
Also did you manually ensure that breakpoints work in test projects in this repo? If not that's fine I can do another round of tests

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -20,3 +20,10 @@ settings append target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
 END
+
+if [ -v BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS ]
+then
+cat <<-END > ~/.lldbinit-source-map
```
gyfelton, line 26: I wonder since we are modifying this file `~/.lldbinit-source-map` why not modify `~/.lldbinit` as well here? Is it because this is our own file while .lldbinit can be shared by multiple projects/other IDE that uses lldb?

gyfelton (changes requested):
Hmmmm this breaks existing tests because `SomeFramework` is not found, interesting.

gyfelton:
i assume you have tested against the downstream project that depends on this project? I am referring to the test script we have to ensure the correct debugging experience in various apps/tests?

gyfelton:
I did a test run for Test-Imports-App under tests/ios/unit-test/test-imports-app, got the following error in debug window:
`error: the replacement path doesn't exist: "/Users/abc/Development/rules_ios/tests/ios/xcodeproj/../../../bazel-../external"`
In the debug window, I can see variable showing up with correct types but auto complete does not work:
Screen Shot 2020-09-09 at 3 53 51 PM
Wonder if this is a regression (actually I don't think so because previously debugging is not even supported entirely)

tools/xcodeproj_shims/installers/lldb-settings.sh
amberdixon: I think it's because the developer owns ~/.lldbinit and should be allowed to put what they want in there without us clobbering it. For example, they might put Tulsi LLDB debug settings in there too or any custom debug settings they want xcode to pick up.

If we ask the developer to explicitly include this other file (~/lldbinit-source-map) in their ~/lldb-init, then they maintain control of the contents of lldbinit.

rules/xcodeproj.bzl
amberdixon: Sure, I can explain how I went about finding these problems. I first added the line `log enable lldb types -f /tmp/types.log` to my lldbinit, so that I would get very verbose diagnostic lldb logging in a file called /tmp/types.log

1. I noticed the linker invocations being made by bazel to generate the final binary did not reference all the AST paths for transitive dependencies. I saw that LLDB was (1) sometimes complaining about missing ASTs and (2) in the types.log file , I noticed that the AST for transitive deps were not being loaded (They were being loaded in the working case.) So I just experimented with adding on the `add_ast_path` option to the linker and that did the trick.

As far as the actual implementation of this change, Sam noticed that framework.bzl was not propagating linker opts for transitive deps, which would explain why these linker options were not present.

2. After I made that fix, debugging was still not working. The LLDB output was complaining about an error with importing GPBProtocolBuffers.h. I noticed that one of our projects at Square has a clang flag with GPB in the name (this flag affects the way we build source files that depend on google protobufs.) So I used the lldb setting for extra-clang-flags to add all the necessary clang flags.

amberdixon: I will double check the breakpoints in the test project, but note that debugging in the test project only works when sandboxing is disabled (which requires changing local bazelrc.)

In order to test this change, I updated build.sh to check for the presence of AST files in the symbol table for the binary. I think that should be sufficient.

amberdixon:
> I did a test run for Test-Imports-App under tests/ios/unit-test/test-imports-app, got the following error in debug window:
> `error: the replacement path doesn't exist: "/Users/abc/Development/rules_ios/tests/ios/xcodeproj/../../../bazel-../external"`
> In the debug window, I can see variable showing up with correct types but auto complete does not work:
> Screen Shot 2020-09-09 at 3 53 51 PM
> Wonder if this is a regression (actually I don't think so because previously debugging is not even supported entirely)

I'm seeing this too, I'll dig in. Not sure if it's regression, this wasn't working too well to begin with :( Note that because of issue #108 , you have to disable sandboxing in this repo's bazelrc to get debugging working locally. That alone is not enough however, because I still see this issue.

amberdixon:
> Hmmmm this breaks existing tests because `SomeFramework` is not found, interesting.

I appear to have green builds now. I was probably missing a few things, but they should be resolved.

amberdixon:
> > I did a test run for Test-Imports-App under tests/ios/unit-test/test-imports-app, got the following error in debug window:
> > `error: the replacement path doesn't exist: "/Users/abc/Development/rules_ios/tests/ios/xcodeproj/../../../bazel-../external"`
> > In the debug window, I can see variable showing up with correct types but auto complete does not work:
> > Screen Shot 2020-09-09 at 3 53 51 PM
> > Wonder if this is a regression (actually I don't think so because previously debugging is not even supported entirely)
> 
> I'm seeing this too, I'll dig in. Not sure if it's regression, this wasn't working too well to begin with :( Note that because of issue #108 , you have to disable sandboxing in this repo's bazelrc to get debugging working locally. That alone is not enough however, because I still see this issue.

I identified the fix to this issue and will push it shortly. LLDB settings should not point to directories that don't exist! will fix.

amberdixon:
> i assume you have tested against the downstream project that depends on this project? I am referring to the test script we have to ensure the correct debugging experience in various apps/tests?

Yes, I have tested with one of our internal projects at Square and confirmed that these changes fix the problem we were seeing with debugging swift dependencies. Before my changes were applied, it was not possible to evaluate variables in the debugger.

rules/xcodeproj.bzl
```diff
@@ -403,7 +403,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_BUILD_EXEC": "$BAZEL_STUBS_DIR/build-wrapper",
         "BAZEL_OUTPUT_PROCESSOR": "$BAZEL_STUBS_DIR/output-processor.rb",
         "BAZEL_PATH": ctx.attr.bazel_path,
-        "BAZEL_RULES_IOS_OPTIONS": "--@build_bazel_rules_ios//rules:local_debug_options_enabled --features=-swift.cacheable_swiftmodules --features=-swift.use_global_module_cache",
```
segiddins, line 406: any particular reason these features are removed?

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -15,8 +15,26 @@ set -euo pipefail
...
 cat <<-END > ~/.lldbinit-source-map
-settings set target.source-map ./external/ "$BAZEL_WORKSPACE_ROOT/bazel-$(basename "$BAZEL_WORKSPACE_ROOT")/external"
-settings append target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
+settings set target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
 END
+
+set +u
+if [[ -n $BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS ]]
```
segiddins, line 24: could do `-n ${BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS:-}` so you don't need to turn off `-u`

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -15,8 +15,26 @@ set -euo pipefail
...
+then
+set -u
+cat <<-END >> ~/.lldbinit-source-map
```
segiddins, line 27: nit: indent the contents of conditional bodies

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -15,8 +15,26 @@ set -euo pipefail
...
+settings set -- target.swift-extra-clang-flags $BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS
+END
+fi
+set -u
+
+BAZEL_EXTERNAL_DIRNAME="$BAZEL_WORKSPACE_ROOT/bazel-$(basename "$BAZEL_WORKSPACE_ROOT")/external"
+if [ -d $BAZEL_EXTERNAL_DIRNAME ]
```
segiddins, line 34: quote the variable in case it contains a space?

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -15,8 +15,26 @@ set -euo pipefail
...
+then
+cat <<-END >> ~/.lldbinit-source-map
+settings append target.source-map ./external/ $BAZEL_EXTERNAL_DIRNAME
```
segiddins, line 37: does it need to be quoted?

segiddins (approved):
looks good! left a few comments, feel free to merge once they're addressed

rules/xcodeproj.bzl
gyfelton: Thanks for the detailed explanation! Learnt that we can do logging for lldb just like we can do it for SourceKIT (so in future get the logs is the priority so we can not shooting in the dark)
As far as the testing go, I think it might be worth to call out in readme that sandbox needs to be disabled to get debugging working? Definitely a separate concern tho

gyfelton (approved):
There are some comments from Sam that are definitely worth fixing

tools/xcodeproj_shims/installers/lldb-settings.sh
amberdixon: thank you!! i was having trouble figuring this out.

rules/xcodeproj.bzl
gyfelton: @amberdixon Can point to me where are the changes on build.sh for presence of AST files? 🙏

gyfelton: > @amberdixon Can point to me where are the changes on build.sh for presence of AST files? 🙏

saw the updated checks, that's awesome!

rules/xcodeproj.bzl
amberdixon: Because of this change, debugging works in local development, even when cacheable swiftmodules are enabled. Therefore, we don't need to explicitly turn off the cacheable swiftmodules feature ... the developer can decide what to do with the feature in their bazelrc if they wish.

segiddins: awesome!!

PR #121, opened 2020-09-18
New attribute to only generate of schemes for certain types of targets

gyfelton:
### Description
Aim is to introduce a new attribute `generate_schemes_for_product_types`:
1. refactor on affect part that deals with generating dicts for schemes and targets under xcode proj context: extracted into methods and hardcoded strings (this is in a separate commit)
2. added a new attribute `generate_schemes_for_product_types ` to `xcodeproj` rule in order to pass in an array of `product_types` so that` xcodeproj` implementation can do a filter on what schemes to generate if this array is not empty

### Tests
1. test fixtures are updated as schemes are gone in this PR
2. added the new attribute to some targets under `tests/ios/xcodeproj` so that a scheme for tests and a scheme for framework under `Test-Imports-App-Project.xcodeproj` is gone
4. all `.xcodeproj` is removed during the `regenerate project` steps in workflow to ensure that the regenerated file has the same files as before.


.github/workflows/tests.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         run: sudo xcode-select -s /Applications/Xcode_11.6.app
       - name: Generate Xcode projects and compare against existing record
         run: |
+          rm -rf tests/macos/xcodeproj/*.xcodeproj
```
segiddins, line 49: this should go after the `set -euo pipefail`

segiddins (approved):

.github/workflows/tests.yml
gyfelton: will do! thanks!

amberdixon: Should this proj get removed by bazelisk clean ?

amberdixon (approved):

rules/xcodeproj.bzl
```diff
@@ -670,6 +692,7 @@ Tags for configuration:
         "project_name": attr.string(mandatory = False),
         "bazel_path": attr.string(mandatory = False, default = "bazel"),
         "scheme_existing_envvar_overrides": attr.string_dict(allow_empty = True, default = {}, mandatory = False),
+        "generate_schemes_for_product_types": attr.string_list(mandatory = False, allow_empty = True, default = [], doc = "Generate schemes only for the specified product types if this list is not empty"),
```
amberdixon, line 695: The valid values are all valid uniform type identifiers for the product type. They are the keys for the dictionary listed here: https://www.rubydoc.info/gems/xcodeproj/Xcodeproj/Constants under the `PRODUCT_TYPE_UTI`.

Valid examples are `application`,`bundle.unit-test`, `framework`, etc.

amberdixon: It would be good if you could include this info in the documentation string. You could say:

"Product types must be valid apple product types, e.g. `application`, `bundle.unit-test`, `framework`."

.github/workflows/tests.yml
gyfelton: clean only clears the intermediate products under bazel-out/bazel-bin, certainly not the output right?

amberdixon: Oh yeah, you're right. Thanks for the clarification.

PR #123, opened 2020-09-22
Bazel build produces error when adding generated header to .h file

gyfelton:
See failed build in this PR, only change is adding the -swift.h file
If doing the same thing in a xcode generated project, it will be fine and pointing to the generated `-swift` file

Unless this is not supported?


segiddins:
So I think this isn't working because the header doing the include is in turn referenced by the modulemap, which is imported by swift. For this to work, it's possible you need to wrap the entire header in a `#if __has_include` directive? I'm not 100% sure if Xcode will go out of it's way to make this work though (maybe by pointing the headermap used for swift compilation to an empty header representing the -Swift.h file?)

PR #127, opened 2020-10-01
Setting cwd explicitly to ensure lldb remote debugging works on relative paths in debug info etc.

gyfelton:
This is important for lldb debugger to work correctly according to https://lldb.llvm.org/use/remote.html#id6
In future we can ask developers to run `platform status` inside lldb to ensure some metadata is correct.

Without this change, do `platform shell pwd` in lldb says `/` and after this it becomes whatever workspace you are actually in.

This might be a solution to some past/existing issues related to lldb debugger.
Testing against downstream project to make sure this works.

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
@@ -11,10 +11,15 @@ set -euo pipefail
 # the paths in the debug info. This lldb setting ensures that _project relative_
 # paths are remapped to _project absolute_ paths.
 #
+# The platform settings one points the current working directory (cwd) to workspace root.
+# This is needed for features relying on lldb remote debugging, such as `oso_prefix_is_pwd`.
+#
 # NOTE: In order to use this, add this line to `~/.lldbinit`:
 #
 #     command source ~/.lldbinit-source-map
 cat <<-END > ~/.lldbinit-source-map
+platform settings -w "$BAZEL_WORKSPACE_ROOT/"
+
 settings set target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
--- a/tools/xcodeproj_shims/installers/lldb-settings.sh
+++ b/tools/xcodeproj_shims/installers/lldb-settings.sh
@@ -11,10 +11,15 @@ set -euo pipefail
 # the paths in the debug info. This lldb setting ensures that _project relative_
 # paths are remapped to _project absolute_ paths.
 #
+# The platform settings one points the current working directory (cwd) to workspace root.
+# This is needed for features relying on lldb remote debugging, such as `oso_prefix_is_pwd`.
+#
 # NOTE: In order to use this, add this line to `~/.lldbinit`:
 #
 #     command source ~/.lldbinit-source-map
 cat <<-END > ~/.lldbinit-source-map
+platform settings -w "$BAZEL_WORKSPACE_ROOT/"
+
 settings set target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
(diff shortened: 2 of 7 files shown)


tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
```diff
@@ -11,13 +11,19 @@ set -euo pipefail
...
+#
 # NOTE: In order to use this, add this line to `~/.lldbinit`:
 #
 #     command source ~/.lldbinit-source-map
 cat <<-END > ~/.lldbinit-source-map
 settings set target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
+
+platform settings -w "$BAZEL_WORKSPACE_ROOT/"
```
keith, line 25: Interesting, I've only tested this by doing `script import os; os.chdir("$BAZEL_WORKSPACE_ROOT")` but this seems nicer if it works 🤔 

Do you have to reset this setting in other LLDB sessions? you do for my option

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -11,13 +11,20 @@ set -euo pipefail
...
 # NOTE: In order to use this, add this line to `~/.lldbinit`:
 #
 #     command source ~/.lldbinit-source-map
 cat <<-END > ~/.lldbinit-source-map
+platform settings -w "$BAZEL_WORKSPACE_ROOT/"
+
 settings set target.source-map ./ "$BAZEL_WORKSPACE_ROOT/"
 settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
+
```
segiddins, line 26: are these trailing newlines intentional?

segiddins (approved):

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
gyfelton: @keith do you do `script import os; os.chdir("$BAZEL_WORKSPACE_ROOT")` in `~/.lldbinit-source-map` as well like here? It fails for me for some reason

gyfelton: You have a good point since this is on user directory level. let me test this out!

gyfelton: and this works btw, tested in downstream project that uses this

keith: IIRC with the script option lldb-rpc-server kept around some state about that change, where it looks like it doesn't with your change, which is awesome

gyfelton: @keith so this is going to be a problem for bazel projects not built using this rules_ios because this script is embedded inside every xcodeproj generated here so running one will overwrite the content from the previous one.
As we are switching to use per scheme .lldbinit file in future, we will eventually pointing this script to a per-project file not user directory

tools/xcodeproj_shims/installers/lldb-settings.sh
gyfelton: nope, let me fix this and merge

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
gyfelton: > IIRC with the script option lldb-rpc-server kept around some state about that change, where it looks like it doesn't with your change, which is awesome

Wanted to learn how u know about the state in this lldb-rpc-server, also i wonder what does `platform status` output looks like for you when use the script `os.chdir("$BAZEL_WORKSPACE_ROOT")`. This can be invoked in the debugging window with `lldb`

PR #125, opened 2020-09-25
Stop specifiying a manual group for xcodeproj sources

segiddins:
Xcodegen can now intuit groups properly for paths that start with .., and are above the current directly on disk

(thanks to https://github.com/yonaskolb/XcodeGen/pull/892).

Pointing to my fork for now, since there hasn't been a recent release

PR #128, opened 2020-10-01
Correctly fix the path of the workspace root group when installing xcodeproj into root

segiddins:
It does not have a trailing slash, since the path was ../../..;

Fixes a regression in https://github.com/bazel-ios/rules_ios/pull/125


gyfelton (approved):

gyfelton:
tested and it works on downstream projects

PR #132, opened 2020-10-06
Add support for xcconfigs that use conditions & default values

segiddins:
This... isn't ideal. But given the inability to infinite loop or recurse in starlark, we need to generate boilerplate to evaluate these expressions.

The trickiest bit of this PR is the tracking over whether any user values have been used for a particular xcconfig -- we do this tracking to ensure that we don't blindly add in Xcode's defaults (i.e. if a user passes in an empty xcconfig, no copts will be added).

This allows using a much broader range of xcconfig options, such as those that are used only inside conditions for other options.


data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
     "com.apple.xcode.tools.swift.compiler",
 )
 
+cond_contents = """
+# foo
```
segiddins, line 69: add a real header & rename this var

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
...
 bzl_contents = """
 # Extracted from Xcode %s
 # To update, in rules_ios run `bazel run data_generators:extract_xcspecs`
 
 SETTINGS = """ % xcode_version
 
+meths = {}
+
+def str_to_thing(string, key):
+    r = re.compile("\$\(([^\$)]+)\)")
```
segiddins, line 82: hoist into constant

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
...
+        if "DefaultValue" in opt:
+            ({var}_used_user_content, {var}) = XCSPEC_EVALS[opt["DefaultValue"]](xcconfigs, id_configs)
+            used_user_content = used_user_content or {var}_used_user_content""".format(var=var, name=name, key=key, key_value=format_str(name))
+        for (name, (var, _i, key)) in d.items()
+    ))
+
+    if key == 'DefaultValue':
+        return s + '\n\n    return (used_user_content, %s)' % (format_str(string), )
+    elif key == 'Condition':
+        string = re.sub(pattern="(?:^|[(&|!= ])(?!eval_val_\d+)([A-Za-z0-9_]+)", repl= lambda m: repr(m[1]), string=string)
```
segiddins, line 132: constant re

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
...
+        return s + '\n\n    return (used_user_content, (%s))' % string.replace('||', 'or').replace('!=', '__NEQ__').replace('!', 'not ').replace('&&', 'and').replace('__NEQ__', '!=')
+    
+    raise "Unknown key %s" % key
+
+def _add_eval(id: str, name, key, string, meths):
+    meth = '__'.join((id.replace('.', '_'), name.replace('.', '_'), key))
+    if meth in meths:
+        raise "Dup"
+
+    meths[meth] = """
```
segiddins, line 142: de-dupe on contents of method?

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
...
+    if key == 'DefaultValue':
+        return s + '\n\n    return (used_user_content, %s)' % (format_str(string), )
+    elif key == 'Condition':
+        string = re.sub(pattern="(?:^|[(&|!= ])(?!eval_val_\d+)([A-Za-z0-9_]+)", repl= lambda m: repr(m[1]), string=string)
+        return s + '\n\n    return (used_user_content, (%s))' % string.replace('||', 'or').replace('!=', '__NEQ__').replace('!', 'not ').replace('&&', 'and').replace('__NEQ__', '!=')
+    
+    raise "Unknown key %s" % key
+
+def _add_eval(id: str, name, key, string, meths):
+    meth = '__'.join((id.replace('.', '_'), name.replace('.', '_'), key))
```
segiddins, line 138: maybe mangle more, so search results don't get cluttered when grepping for a particular xcconfig name?

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,117 @@ def xcspec_from_file(path):
...
+    # {string}
+    {expr}
+""".format(meth=meth, string=string, expr=str_to_thing(string, key))
+
+    return meth
+
 to_print = {
     x['Identifier']:
-        {k: ({o['Name']: o for o in v} if k == 'Options' else v)
+        {k: ({o['Name']: {ok: (_add_eval(x['Identifier'], o['Name'], ok, ov, meths) if ok in ('DefaultValue', 'Condition') else ov) for (ok, ov) in o.items() if ok not in ('Name',)} for o in v if o['Name'] not in ('SDKROOT',)} if k == 'Options' else v)
```
segiddins, line 152: pull into a method, this is gross

rules/library.bzl
```diff
@@ -192,12 +192,12 @@ def _prepend(list, other):
 
 def _prepend_copts(copts_struct, objc_copts, cc_copts, swift_copts, linkopts, ibtool_copts, momc_copts, mapc_copts):
     _prepend(objc_copts, copts_struct.objc_copts)
-    _prepend(copts_struct.cc_copts, cc_copts)
-    _prepend(copts_struct.swift_copts, swift_copts)
-    _prepend(copts_struct.linkopts, linkopts)
-    _prepend(copts_struct.ibtool_copts, ibtool_copts)
-    _prepend(copts_struct.momc_copts, momc_copts)
-    _prepend(copts_struct.mapc_copts, mapc_copts)
+    _prepend(cc_copts, copts_struct.cc_copts)
```
segiddins, line 195: ... bad sam, this has been broken for a while?

rules/library/xcconfig.bzl
```diff
@@ -10,56 +12,84 @@ _MAPC = "com.apple.compilers.model.coredatamapping"
...
+            ))
+
+        (condition_used_user_config, condition) = condition_function(xcconfigs, SETTINGS[xcspec]["Options"])
+
+        if not condition:
+            return
+
+        used_user_config = used_user_config or condition_used_user_config
+
+    if not used_user_config:
```
segiddins, line 59: add a comment here:

```
the point of this is to not pollute copts with all of xcode's defaults.
only take into account settings that the user's explicit config have influenced (e.g. their config was used in the conditional or default value, if not the value for the setting itself)
```

tests/macos/xcconfig/swift_internal.swift
```diff
@@ -0,0 +1,3 @@
+internal class HiThere: NSObject {
+    @objc internal class func sayHi() {}
+}
```
segiddins, line 3: trailing newline

tests/macos/xcconfig/tests.bzl
```diff
@@ -165,7 +165,12 @@ def xcconfig_unit_test_suite():
             assert_xcconfig(
                 name = "option_with_unknown_command_line",
                 xcconfig = {"OTHER_CFLAGS": ["-IFOO"]},
-                error = 'OTHER_CFLAGS: unable to extract value for ["-IFOO"] in com.apple.compilers.llvm.clang.1_0',
+                expected = {},
+            ),
+            assert_xcconfig(
+                name = "only_used_as_condition",
+                xcconfig = {"APPLICATION_EXTENSION_API_ONLY": "YES"},
+                expected = {"linkopts": ["-fapplication-extension", "-fapplication-extension"], "objc_copts": ["-fapplication-extension"], "swift_copts": ["-application-extension"]},
```
segiddins, line 173: de-dupe somehow?

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,147 @@ def xcspec_from_file(path):
     "com.apple.xcode.tools.swift.compiler",
 )
 
-bzl_contents = """
-# Extracted from Xcode %s
+header = """
+# Extracted from Xcode {xcode_version}
 # To update, in rules_ios run `bazel run data_generators:extract_xcspecs`
+""".format(xcode_version=xcode_version)
```
ob, line 71: Let's add a much bigger warning here. It's easy to miss that those two files are generated code.

```suggestion
header = """
#########################################################
#                                  THIS IS GENERATED CODE                                             #
# Extracted from Xcode {xcode_version}                                                           #
# To update, in rules_ios run `bazel run data_generators:extract_xcspecs`  #
#########################################################
""".format(xcode_version=xcode_version)
```

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,147 @@ def xcspec_from_file(path):
...
+            )
+
+    raise "Unknown key %s" % key
+
+
+def add_eval(id, name, key, string, eval_methods):
+    method_name = '__'.join(
+        (id.replace('.', '_'), name.replace('.', '_'), key))
+    if method_name in eval_methods:
+        raise "Dup"
```
ob, line 162: I'd add more information here... at least `method_name`

data_generators/xcspec_extractor.py
```diff
@@ -63,29 +65,147 @@ def xcspec_from_file(path):
...
+def _{method_name}(xcconfigs, id_configs):
+    # {string}
+    {expr}
+""".format(method_name=method_name, string=string, expr=generate_method_body_for_expression(string, key))
+
+    return method_name
 
-SETTINGS = """ % xcode_version
 
 to_print = {
```
ob, line 173: Not a biggie and we can fix it later but I tend to find python scripts without a `main` and with code scattered all around the global namespace hard to read. I'd take all the logic and shove it in a `main()` and do the standard:

```
if __name__ == '__main__':
    main()
````

dance.

File: rules/library.bzl [hunk 14]
```diff
@@ -296,7 +296,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     namespace = module_name if namespace_is_module_name else name
     module_map = kwargs.pop("module_map", None)
     cc_copts = kwargs.pop("cc_copts", [])
-    swift_copts = kwargs.pop("swift_copts", [])
+    swift_copts = list(kwargs.pop("swift_copts", []))
```
ob, line 299: I'm curious why you need the cast to `list` here...

ob (approved):
Nice! Pretty soon we'll have a compiler-within-a-compiler :-)

One question though... does this mean we are somewhat loosely tied to an Xcode release now? Since we are checking in the generated files...

rules/library.bzl
justinseanmartin: It this just a matter of precedence?

rules/library.bzl
justinseanmartin, line 299: Why the move to list container here but not the others?

rules/library/xcconfig.bzl
```diff
@@ -134,6 +162,20 @@ def settings_from_xcconfig(xcconfig):
     ibtool_copts = []
     linkopts = []
 
+    xcconfig = dicts.add(
+        {
+            "CLANG_ENABLE_MODULE_DEBUGGING": "NO",
```
justinseanmartin, line 167: Why are these two special cased here? Seems potentially a bit buried. Would it be worth pulling these out to somewhere more visible?

rules/library/xcconfig.bzl
```diff
@@ -10,56 +12,87 @@ _MAPC = "com.apple.compilers.model.coredatamapping"
...
+                value = repr(value),
+            ))
     elif _type in ("String", "Path") and not types.is_string(value):
         fail("{name}: {value} not a valid value, must be a string".format(
             name = name,
             value = value,
         ))
 
-    if option.get("Condition"):
-        fail("{name}: option specifies a condition ({cond}), which is currently unsupported".format(
```
justinseanmartin, line 58: No more!

rules/library/xcconfig.bzl
```diff
@@ -112,12 +145,7 @@ def _add_copts_from_option(xcspec, option, value, value_escaper, copts, linkopts
         new = [command_line_prefix_flag + "$(value)"]
 
     if new == None:
-        fail("{name}: unable to extract value for {value} in {xcspec}\n\t{option}".format(
```
justinseanmartin, line 115: Looks like you're explicitly expecting this now with `Command LineFlag` setting `new = None`. Was this catching anything else that will now no longer be caught?

justinseanmartin (approved):

rules/library.bzl
segiddins: forgot to revert it

segiddins:
> One question though... does this mean we are somewhat loosely tied to an Xcode release now? Since we are checking in the generated files...

We already were checking in a generated file, now we're checking in a second

rules/library/xcconfig.bzl
segiddins: Not that I know of -- now that we support conditions, there are some options that are _only_ used as options

rules/library/xcconfig.bzl
segiddins: can pull into a constant and comment, but we explicitly disable these because they turn on flags we really don't want when `CLANG_ENABLE_MODULES = YES`

rules/library.bzl
segiddins: No -- we were mutating the wrong variable, meaning swift copts from xcconfigs weren't being applied to `swift_library` targets.  the new test framework shows that this is fixed now

rules/library/xcconfig.bzl
justinseanmartin: Should this be up to the caller to pass this in?

segiddins: IMO no -- it would mean that every build would have warnings (from the second setting), and would be non-reproducible (from the first)

tests/macos/xcconfig/tests.bzl
```diff
@@ -177,11 +182,15 @@ def xcconfig_unit_test_suite():
                 xcconfig = {"GCC_PREPROCESSOR_DEFINITIONS": ["DISPLAY_VERSION=1.0.0-beta.1", "SDK_NAME=WHY WOULD YOU ADD SPACES"]},
                 expected = {"objc_copts": ["-D'DISPLAY_VERSION=1.0.0-beta.1'", "-D'SDK_NAME=WHY WOULD YOU ADD SPACES'"]},
             ),
-            # TODO: we should eventually support conditioned vars somehow
             assert_xcconfig(
-                name = "conditioned_option_unsupported",
+                name = "conditioned_option",
                 xcconfig = {"CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER": "YES"},
-                error = 'CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER: option specifies a condition ("$(CLANG_UNDEFINED_BEHAVIOR_SANITIZER) == YES"), which is currently unsupported',
+                expected = {},
```
amberdixon, line 188: Why isn't the expected value ```{"objc_copts": ["-fsanitize=integer"]}```

amberdixon (approved):

tests/macos/xcconfig/tests.bzl
amberdixon: I think there might be a bug in xcspecs_eval.bzl. It should say:

```
def _com_apple_compilers_llvm_clang_1_0__CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER__Condition(xcconfigs, id_configs):
    # $(CLANG_UNDEFINED_BEHAVIOR_SANITIZER) == YES

    used_user_content = False
    
    eval_val_0 = ""
   # bug: currently, eval_key_0 is "CLANG_UNDEFINED_BEHAVIOR_SANITIZER", but the key passed to xcconfig is "CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER"
    eval_key_0 = "CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER"
```

segiddins: why is this a bug? The actual condition for the option is `$(CLANG_UNDEFINED_BEHAVIOR_SANITIZER) == YES`, which isn't set here

amberdixon: OK, so you're saying that setting the `CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER` key in xcconfig is a no-op. Can we rename it to something else in this test ... like:
```
                xcconfig = {"foobar": "YES"},
                expected = {},
```

To someone reading this test case, the `CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER` looks just like the `CLANG_UNDEFINED_BEHAVIOR_SANITIZER` key, which would actually result in the "objc_copts" getting set to ["-fsanitize=integer"].

segiddins: but `foobar` isn't a valid xcconfig.... you can look in the xcspec bzl file to find all the one's that we're operating over

amberdixon: All right. Maybe there should be a comment here ... because when I read this test case at first, I would have assumed that setting `CLANG_UNDEFINED_BEHAVIOR_SANITIZER_INTEGER` to YES would change the copts ... when in fact, it is only setting `CLANG_UNDEFINED_BEHAVIOR_SANITIZER` to YES that would affect cops.

PR #133, opened 2020-10-06
Do not apply force_swift_local_debug_options transition. 

amberdixon:
There should be no difference between applying the local_debug_options_enabled build setting via the command line vs applying it with a transition.

But (possibly due to this bug: https://github.com/bazelbuild/bazel/issues/12171), when bazel sees this transition is present, it adds a transition directory name fragment, even if the build settings are identical to having the local_debug_options_enabled flag specified on command line. This results in a different output directory fragment for dependent artifacts when developers call `bazel run xcodeproject` with the transition applied vs. when the xcodeproject invokes `bazel build <targetname> --local_debug_options_enabled`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,5 +93,7 @@ ios_application(
 See the [tests](https://github.com/bazel-ios/rules_ios/tree/master/tests)
 directory for sample applications.
 
-## Special note about debugging
+## Special notes about debugging xcode projects
+Note that when generating xcodeprojects, it is necessary to supply the `--//rules:local_debug_options_enabled` so that debugging of swift files will work.
+
 Debugging does not work in sandbox mode, due to issue #108. The workaround for now is to disable sandboxing in the .bazelrc file.
--- a/README.md
+++ b/README.md
@@ -94,6 +94,6 @@ See the [tests](https://github.com/bazel-ios/rules_ios/tree/master/tests)
 directory for sample applications.
 
 ## Special notes about debugging xcode projects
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -93,5 +93,7 @@ ios_application(
 See the [tests](https://github.com/bazel-ios/rules_ios/tree/master/tests)
 directory for sample applications.
 
-## Special note about debugging
+## Special notes about debugging xcode projects
+Note that when generating xcodeprojects, it is necessary to supply the `--//rules:local_debug_options_enabled` so that debugging of swift files will work.
```
segiddins, line 97:
```suggestion
Note that when generating Xcode projects, it is necessary to supply the `--//rules:local_debug_options_enabled` so that debugging of swift files will work.
```

also, should include the repository in the flag name

rules/xcodeproj.bzl
```diff
@@ -674,7 +674,7 @@ def _xcodeproj_impl(ctx):
 xcodeproj = rule(
     implementation = _xcodeproj_impl,
     doc = """\
-Generates a Xcode project file (.xcodeproj) with a reasonable set of defaults
+Generates a Xcode project file (.xcodeproj) with a reasonable set of defaults. Specify the --//rules:local_debug_options_enabled when generating the project to ensure debugging of swiftmodules works.
```
segiddins, line 677: include the repository in the flag name

segiddins (approved):

PR #134, opened 2020-10-07
Add sources to project, instead of the target

wileykestner:
- Ensures that the files will continue to show up in Xcode's 'Project
  Navigator'

- Ensures that the same files will _not_ show up in any target's
  'Compile Sources' build phase

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -425,6 +425,7 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots):
         compiled_sources = [{
             "path": paths.join(src_dot_dots, s.short_path),
             "optional": True,
+            "buildPhase": "none",
         } for s in target_info.srcs.to_list()]
         compiled_non_arc_sources = [{
             "path": paths.join(src_dot_dots, s.short_path),
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
@@ -6,11 +6,6 @@
 	objectVersion = 51;
 	objects = {
 
-/* Begin PBXBuildFile section */
-		0BD7FD856ED50175FA3D3BC9 /* FooFrameworkTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 091655034CF618D208C22AE8 /* FooFrameworkTests.m */; };
-		5C8EED4DAAF3B4B069337569 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = CF7A2F537C999952921627A8 /* main.m */; };
-/* End PBXBuildFile section */
-
 /* Begin PBXFileReference section */
 		091655034CF618D208C22AE8 /* FooFrameworkTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = FooFrameworkTests.m; sourceTree = "<group>"; };
 		11B98E6654D00F1E74B2DA47 /* BUILD.bazel */ = {isa = PBXFileReference; path = BUILD.bazel; sourceTree = "<group>"; };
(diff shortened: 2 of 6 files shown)


segiddins:
What impact does this have on indexing & debugging functionality?

jschear:
Hey @segiddins, the short answer is that we're not sure yet (and it very well may break indexing, debugging, or some other IDE features) -- we're trying to grok how the "compile sources" phase fits into those features right now.

Wiley opened this PR to demonstrate a possible solution to an issue we posted about in [ios-dev-at-scale](https://ios-dev-at-scale.slack.com/archives/CRPRV58UF/p1602097501016800?thread_ts=1601655120.009800&cid=CRPRV58UF) (we're hitting `ARG_MAX` because we have too many files in a single target & the legacy build system appears to not use an argument file when compiling), but this might be the wrong approach.

gyfelton:
Going to check on the impact of removing compile source phase but most likely this is going to break indexing. 
Will pull down this branch to test soon because I want to investigate if compile source phase can help us with getting a health check on indexing.

wileykestner:
> to check on the impact of removing compile source phase but most likely this is going to break indexing.
> Will pull down this branch to test soon because I want to investigate if compile source phase can help us with

Thanks so much @gyfelton - as @jschear noted above, this is more of a conversation starter than a legit pull request - we're trying to get up to speed on how all this stuff works.

amberdixon:
Thanks for starting this discussion! A couple thoughts:

1. I would guess that this breaks indexing but will give it a whirl. Xcode's indexer needs to (1) be turned on in order for autocomplete to work and (2) needs to successfully be able to compile source files that are opened/edited. In order to know which compile flags to pass for the source file, xcode looks at the target build settings for that source file.
2. If a developer adds files to the xcodeproject target, note that Xcode will add a new compile sources phase to the target with this source file in it. Assuming you stub out clang, you probably wouldn't hit an argmax issue under those conditions, but it would be very weird to see just one source in the target phase.
3. Maybe the right thing to do here is to explore switching the new build system. (The other reason to switch to new build system is because someday the old one is going to go away!) The original version of the xcode rule *was* generating projects that ran against new build system ... but I switched it over to use the old build system and here was the reason for it: https://github.com/bazel-ios/rules_ios/pull/42#issuecomment-629967829 The last time I checked, Tulsi was also using the old build system.

I will confirm that this does in fact break indexing and revisit using the new build system.

amberdixon:
> Thanks for starting this discussion! A couple thoughts:
> 
> 1. I would guess that this breaks indexing but will give it a whirl. Xcode's indexer needs to (1) be turned on in order for autocomplete to work and (2) needs to successfully be able to compile source files that are opened/edited. In order to know which compile flags to pass for the source file, xcode looks at the target build settings for that source file.
> 2. If a developer adds files to the xcodeproject target, note that Xcode will add a new compile sources phase to the target with this source file in it. Assuming you stub out clang, you probably wouldn't hit an argmax issue under those conditions, but it would be very weird to see just one source in the target phase.
> 3. Maybe the right thing to do here is to explore switching the new build system. (The other reason to switch to new build system is because someday the old one is going to go away!) The original version of the xcode rule _was_ generating projects that ran against new build system ... but I switched it over to use the old build system and here was the reason for it: [#42 (comment)](https://github.com/bazel-ios/rules_ios/pull/42#issuecomment-629967829) The last time I checked, Tulsi was also using the old build system.
> 
> I will confirm that this does in fact break indexing and revisit using the new build system.

Hi all! I have put up a draft PR that seems to successfully switch out xcode projects to using the new build system: https://github.com/bazel-ios/rules_ios/pull/141

rules/xcodeproj.bzl
```diff
@@ -425,6 +425,7 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots):
         compiled_sources = [{
             "path": paths.join(src_dot_dots, s.short_path),
             "optional": True,
+            "buildPhase": "none",
```
gyfelton, line 428: Have u tested whether files not inside compiled source phase will be indexed by XCode? I am pretty they are added to the compiled source list is so that XCode knows them and will index the files.

gyfelton: @amberdixon am i correct on this?

PR #137, opened 2020-10-09
Supports the select method in the deps of apple_library

congt:
cc @thiagohmcruz

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -308,7 +308,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     weak_sdk_frameworks = kwargs.pop("weak_sdk_frameworks", [])
     sdk_includes = kwargs.pop("sdk_includes", [])
     pch = kwargs.pop("pch", "@build_bazel_rules_ios//rules/library:common.pch")
-    deps = [d for d in kwargs.pop("deps", [])]
+    deps = [] + kwargs.pop("deps", [])
     data = kwargs.pop("data", [])
     tags = kwargs.pop("tags", [])
     tags_manual = tags if "manual" in tags else tags + _MANUAL
@@ -341,6 +341,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         )
         internal_deps.append(linkopts_name)
 
+    vendored_deps = []
     for vendored_static_framework in kwargs.pop("vendored_static_frameworks", []):
         import_name = "%s-%s-import" % (name, paths.basename(vendored_static_framework))
         apple_static_framework_import(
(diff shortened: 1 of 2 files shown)


rules/library.bzl
```diff
@@ -308,7 +308,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     weak_sdk_frameworks = kwargs.pop("weak_sdk_frameworks", [])
     sdk_includes = kwargs.pop("sdk_includes", [])
     pch = kwargs.pop("pch", "@build_bazel_rules_ios//rules/library:common.pch")
-    deps = [d for d in kwargs.pop("deps", [])]
+    deps = [] + kwargs.pop("deps", [])
```
congt, line 311: Got `Error: trying to mutate a frozen list value` if I use `deps = kwargs.pop("deps", [])` when building tests/framework/platforms/BUILD.bazel

tests/ios/frameworks/selectable-deps/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+apple_framework(
+    name = "SelectableDepsFramework",
+    srcs = [
+        "main.m",
+    ],
+    platforms = {"ios": "10.0"},
+    visibility = ["//visibility:public"],
+    deps = select({
+        ":release": [":SelectableFramework"],
+        "//conditions:default": [":SelectableFramework"]
```
thiagohmcruz, line 32: Missing a `,` here at the end

thiagohmcruz (approved):

tests/ios/frameworks/selectable-deps/BUILD.bazel
thiagohmcruz: Just noticed the unit tests are failing, never mind my previous approval we should check and see why

rules/library.bzl
```diff
@@ -357,17 +358,18 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
         objc_import(
             name = import_name,
             archives = [vendored_static_library],
             tags = _MANUAL,
         )
-        deps.append(import_name)
+        vendored_deps.append(import_name)
     for vendored_dynamic_library in kwargs.pop("vendored_dynamic_libraries", []):
         fail("no import for %s" % vendored_dynamic_library)
+    deps += vendored_deps
```
segiddins, line 372: nice! thanks for keeping this a single array, it'll make reading `bazel query --output=build` output much nicer

segiddins (approved):
Awesome! I think for the sake of completeness we should also have an `ios_application` and `ios_unit_test` that have selected deps in the tests?

PR #139, opened 2020-10-15
Add resources arg

fnazarios:


segiddins (approved):

rules/app.bzl
```diff
@@ -14,6 +14,7 @@ _IOS_APPLICATION_KWARGS = [
     "provisioning_profile",
     "resources",
     "app_icons",
+    "resources",
```
jschear, line 17: It looks like this is already present on line 15 (which was added somewhat recently, in #120).

PR #143, opened 2020-10-22
Enable building mixed-language frameworks that have user-specified modulemaps

segiddins:
The main way this is accomplished is by providing a vfsoverlay, since headers from a modulemap aren't found by following hmaps.

This PR also adds support for bridging headers in non-framework targets, for pairing with https://github.com/bazel-ios/cocoapods-bazel/pull/23.

I will look into xcode support for frameworks that use these features in a follow-up.


gyfelton:
one thing to verify is to add back this module to clang_modules and make sure it builds: https://git.sqcorp.co/projects/IOS/repos/register/pull-requests/48821/overview

rules/framework/vfs_overlay.bzl
```diff
@@ -0,0 +1,81 @@
...
+    vfsoverlay_file = ctx.outputs.vfs_overlay
+
+    ctx.actions.write(
+        content = vfsoverlay_yaml,
+        output = vfsoverlay_file,
+    )
+
+    files = depset(direct = [vfsoverlay_file])
+    objc_provider = apple_common.new_objc_provider(
+        # framework_search_path_only = FRAMEWORK_SEARCH_PATH,
```
congt, line 58: nit: remove the dead code?

congt (approved):

rules/framework/vfs_overlay.bzl
thiagohmcruz, line 58: Is this a debugging/testing leftover?

thiagohmcruz (approved):

rules/framework/vfs_overlay.bzl
thiagohmcruz: Oops @congt has pointed that out already, never mind

PR #144, opened 2020-10-28
Point to generated swift objc interface header

amberdixon:
During build and indexing step when the framework scheme is selected, xcode will leave an empty swiftmodule in the build products dir, which causes other autocomplete and debugging operations to fail with an error about a "malformed swiftmodule". Setting the SWIFT OBJC INTERFACE HEADER build setting in xcode will make it so Xcode will not leave empty swiftmodule files around.

When the framework is built from xcode, xcode tries to copy the swiftmodule, swiftdoc and (in the new build system) the swiftsourceinfo files from DerivedData's Build/Intermediates directory into Build/Products using `ditto`. If those files are not present in Build/Intermediates (which they will not be, if the target is built with bazel), then xcode will create empty swiftmodule and swiftdoc files in Build/Products. This PR adds a step to `install.sh` where we copy the Bazel-built swiftmodule, swiftdoc, swiftsourceinfo and -Swift.h file into Xcode's Build/Intermediates directory.


tools/xcodeproj_shims/installers/_swiftmodule.sh
```diff
@@ -5,15 +5,24 @@ set -euo pipefail
...
-    cp "$module" "$module_bundle/$ios_module_name.swiftmodule"
-    cp "$doc" "$module_bundle/$ios_module_name.swiftdoc"
+    cp "$module" "$dest_products_module_bundle/$ios_module_name.swiftmodule"
+    cp "$doc" "$dest_products_module_bundle/$ios_module_name.swiftdoc"
 
-    chmod -R +w "$module_bundle"
+    chmod -R +w "$dest_products_module_bundle"
+	
+	# XCode expects to find swiftmodules in Build/Intermediates and has a build phase where it copies
+	# the Intermediates swiftmodules to Built Products Dir using `ditto`
```
amberdixon, line 23: This is an unfortunate additional step. Even if the xcode build setting DEFINES_MODULE is false, XCode will still try to copy the swiftmodule from Intermediates into Build/Products dir using `ditto`. If the intermediates file doesn't exist, it will generate an empty swiftmodule in Build/Products!

rules/xcodeproj.bzl
```diff
@@ -99,6 +99,12 @@ def _xcodeproj_aspect_impl(target, ctx):
     bazel_bin_subdir = "%s/%s" % (target.label.workspace_root, target.label.package)
 
     if AppleBundleInfo in target:
+        swift_objc_header_path = None
+        if SwiftInfo in target:
```
segiddins, line 103: this seems ick, and the iteration over all headers a potential performance concern: could we instead look at implementing https://github.com/bazelbuild/rules_swift/issues/291 ?

rules/xcodeproj.bzl
```diff
@@ -99,6 +99,12 @@ def _xcodeproj_aspect_impl(target, ctx):
...
+            for h in target[apple_common.Objc].header.to_list():
```
segiddins, line 104: could this iterate over `direct_headers` to limit the amount of depset flattening that needs to be done?

segiddins (approved):

rules/xcodeproj.bzl
amberdixon: Based upon that issue's conversation, it looks like the bazel team is thinking about returning a totally separate provider (called IDEInfo or something along those lines) and returning this type of information in there, as they don't think it should belong in the SwiftInfo. I'm not sure if it's worthwhile for me to try to implement that ticket right now, but I did leave a comment and +1 the issue.

I also looked at Lyft's workaround (https://github.com/lyft/envoy-mobile/pull/797/files) which is very similar to my approach here. In their case, they are pulling the headers list from the CCInfo provider.

gyfelton:
just went through the code, awesome job Amber! 🎉

PR #145, opened 2020-10-29
Lets XCode build based on the selected destination

congt:
When builing for an iOS device in XCode, TARGET_DEVICE_IDENTIFIER is exported, and PLATFORM_NAME is "iphoneos". PLATFORM_NAME is "iphonesimulator" when building for an iOS simulator.
Sets the build option --ios_multi_cpus=arm64 when building for a iOS device. Otherwise, uses the default option.

Also pins GoogleUserMessagingPlatform to fix the master.


tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -6,6 +6,10 @@ if [ $BAZEL_EXECUTION_LOG_ENABLED -gt 0 ]; then
     EXEC_LOG_OPTION="--experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME"
 fi
 
+if [ -n "$(export -p | grep BAZEL_IOS_MULTI_CPUS=)" ]; then
+    EXEC_LOG_OPTION="$EXEC_LOG_OPTION --ios_multi_cpus=$BAZEL_IOS_MULTI_CPUS"
```
segiddins, line 10: this probably doesn't belong in `$exec_log_option` -- i'd reccomend changing this to an array of args passed to bazel that we can freely append to

segiddins (commented):
Can we add a test in here that does an `xcodebuild` for device? Does this have any impact on indexing-based functionality?

rules/xcodeproj.bzl
```diff
@@ -417,6 +417,9 @@ mkdir -p $BAZEL_DIAGNOSTICS_DIR
 export DATE_SUFFIX="$(date +%Y%m%d.%H%M%S%L)"
 export BAZEL_BUILD_EVENT_TEXT_FILENAME="$BAZEL_DIAGNOSTICS_DIR/build-event-$DATE_SUFFIX.txt"
 export BAZEL_BUILD_EXECUTION_LOG_FILENAME="$BAZEL_DIAGNOSTICS_DIR/build-execution-log-$DATE_SUFFIX.log"
+if [ "$PLATFORM_NAME" = "iphoneos" ]; then
```
segiddins, line 420: could this logic be moved entirely into the build wrapper?

tools/xcodeproj_shims/build-wrapper.sh
congt: Apparently I cannot read.

rules/xcodeproj.bzl
congt: Yes. I put it here to make build-wrapper.sh agnostic to the xcode behavior, but probably it should not be in xcodeproj.bzl since it doesn't use any knowledge of xcodeproj.bzl.

tools/xcodeproj_shims/build-wrapper.sh
congt: Done

rules/xcodeproj.bzl
congt: Done

tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -1,15 +1,21 @@
...
+BAZEL_BUILD_OPTIONS+=" --build_event_publish_all_actions"
+
 if [ $BAZEL_EXECUTION_LOG_ENABLED -gt 0 ]; then
-    EXEC_LOG_OPTION="--experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME"
+    BAZEL_BUILD_OPTIONS+=" --experimental_execution_log_file=$BAZEL_BUILD_EXECUTION_LOG_FILENAME"
+fi
+
+# When building for an iOS device in XCode, TARGET_DEVICE_IDENTIFIER is exported, and PLATFORM_NAME is iphoneos.
+if [ -n "$(export -p | grep TARGET_DEVICE_IDENTIFIER=)" ] && [ "$PLATFORM_NAME" = "iphoneos" ]; then
+    BAZEL_BUILD_OPTIONS+=" --ios_multi_cpus=arm64"
```
gyfelton, line 13: should we echo `using arm64 since detected abc` ad a way to tell the scirpt is altering build bahavior/config?

gyfelton (approved):

tools/xcodeproj_shims/build-wrapper.sh
amberdixon: Why are you not able to do `if [ -n ${TARGET_DEVICE_IDENTIFIER:-} ] && [ "$PLATFORM_NAME" = "iphoneos" ]; then`

amberdixon (approved):

tools/xcodeproj_shims/build-wrapper.sh
segiddins: should we maybe instead check the archs env vars? Because I think building for a "generic" device is a thing?

segiddins (approved):
Looks good, but it would be great to add a test in CI that does an `xcodebuild` for device

tools/xcodeproj_shims/build-wrapper.sh
congt: PLATFORM_NAME is `iphoneos` when building for generic device.

congt:
@segiddins I tried to add a test, but xcodebuild emits wrong env variables, e.g., PLATFORM_NAME is iphoneos even if the target is a simulator. Saw someone complained [the same issue](https://stackoverflow.com/questions/13509734/xcodebuild-platform-name-always-iphoneos). Fortunately if a developer wants to build from commandline, she should directly use bazel, instead of xcodebuild.

segiddins:
@congt I think we could check one of the `ARCHS` env vars instead of just the platform name?

congt:
@segiddins xcodebuild has the similar issue with CURRENT_ARCH and some other arch variables. It's arm64 even when the target device is a simulator. I used PLATFORM_NAME instead of CURRENT_ARCH to differentiate iOS and macos. Read somewhere that macos can also use arm64.

PR #151, opened 2020-11-05
Add ability to put xconfig copts behind select() of build settings

thiagohmcruz:
Add `copts` and `infoplists` behind `select()` statements when 'xcconfig_by_build_setting' and 'infoplists_by_build_setting' are set. For example:

Note 'infoplists' here

```
# bazel query //tests/ios/app:AppWithConfigurableAttrs --output=build

ios_application(
  name = "AppWithConfigurableAttrs",
  generator_name = "AppWithConfigurableAttrs",
  generator_function = "ios_application",
  generator_location = "tests/ios/app/BUILD.bazel:139:16",
  bundle_id = "com.example.app",
  infoplists = select({"//tests/ios/app:blue": ["//rules/test_host_app:Info.plist", "//tests/ios/app:AppWithConfigurableAttrs.blue.infoplist.0"], "//tests/ios/app:red": ["//rules/test_host_app:Info.plist", "//tests/ios/app:AppWithConfigurableAttrs.red.infoplist.0"], "//conditions:default": ["//rules/test_host_app:Info.plist", "//tests/ios/app:AppWithConfigurableAttrs.infoplist.0"]}),
  minimum_os_version = "10.0",
  families = ["iphone", "ipad"],
  entitlements = "//tests/ios/app:AppWithConfigurableAttrs_entitlements",
  dylibs = [],
  deps = ["//tests/ios/app:AppWithConfigurableAttrs_objc", "//tests/ios/app:AppWithConfigurableAttrs_private_headers"] + select({"//rules:local_debug_options": [], "//conditions:default": []}) + ["//tests/ios/app:AppWithConfigurableAttrs_entitlements", "//tests/ios/app:AppWithConfigurableAttrs.swift_runtime_linkopts"] + [],
  launch_storyboard = "//tests/ios/app:AppWithConfigurableAttrs_launch_screen_storyboard",
)
```

and `copts` here

```
# bazel query //tests/ios/app:AppWithConfigurableAttrs_objc --output=build

objc_library(
  name = "AppWithConfigurableAttrs_objc",
  tags = ["__CC_RULES_MIGRATION_DO_NOT_USE_WILL_BREAK__", "manual"],
  generator_name = "AppWithConfigurableAttrs",
  generator_function = "ios_application",
  generator_location = "tests/ios/app/BUILD.bazel:139:16",
  deps = ["//tests/ios/app:Empty", "//tests/ios/app:FW", "//tests/ios/app:OnlySources", "//tests/ios/app:AppWithConfigurableAttrs_public_hmap", "//tests/ios/app:AppWithConfigurableAttrs_private_hmap", "//tests/ios/app:AppWithConfigurableAttrs_private_angled_hmap"],
  data = ["//tests/ios/app:AppWithConfigurableAttrs_data"],
  sdk_frameworks = [],
  weak_sdk_frameworks = [],
  sdk_dylibs = [],
  hdrs = [],
  sdk_includes = [],
  copts = select({"//conditions:default": ["-Werror"], "//tests/ios/app:blue": ["-Werror"], "//tests/ios/app:red": []}) + ["-I$(execpath :AppWithConfigurableAttrs_private_hmap)", "-I$(execpath :AppWithConfigurableAttrs_public_hmap)", "-I$(execpath :AppWithConfigurableAttrs_private_angled_hmap)", "-iquote$(execpath :AppWithConfigurableAttrs_private_hmap)", "-fmodules", "-fmodule-name=AppWithConfigurableAttrs", "-I.", "-index-store-path", "$(GENDIR)/tests/ios/app/rules_ios_objc_library_AppWithConfigurableAttrs_objc.indexstore"],
  srcs = ["//tests/ios/app:App/main.m"],
  non_arc_srcs = [],
  pch = "//rules/library:common.pch",
)
```

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -1,6 +1,7 @@
 load("@bazel_skylib//lib:types.bzl", "types")
 load("@build_bazel_rules_apple//apple:ios.bzl", rules_apple_ios_application = "ios_application")
 load("//rules:library.bzl", "apple_library", "write_file")
+load("//rules/library:xcconfig.bzl", "build_setting_name")
 
 _IOS_APPLICATION_KWARGS = [
     "bundle_id",
@@ -56,7 +57,16 @@ def ios_application(name, apple_library = apple_library, **kwargs):
         apple_library: The macro used to package sources into a library.
         **kwargs: Arguments passed to the apple_library and ios_application rules as appropriate.
     """
-    infoplists = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
+
+    infoplists_by_build_setting = kwargs.pop("infoplists_by_build_setting", {})
+    for (build_setting, plists) in infoplists_by_build_setting.items():
+        name_suffix = build_setting_name(build_setting)
+        infoplists_by_build_setting[build_setting] = write_info_plists_if_needed(name = "%s.%s" % (name, name_suffix), plists = plists)
+
+    infoplists_by_build_setting["//conditions:default"] = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
+
+    infoplists = select(infoplists_by_build_setting)
+
     application_kwargs = {arg: kwargs.pop(arg) for arg in _IOS_APPLICATION_KWARGS if arg in kwargs}
     library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": application_kwargs.get("minimum_os_version")}, **kwargs)
 
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -63,7 +63,8 @@ def ios_application(name, apple_library = apple_library, **kwargs):
         name_suffix = build_setting_name(build_setting)
         infoplists_by_build_setting[build_setting] = write_info_plists_if_needed(name = "%s.%s" % (name, name_suffix), plists = plists)
 
-    infoplists_by_build_setting["//conditions:default"] = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
+    default_infoplists = infoplists_by_build_setting.get("//conditions:default", kwargs.pop("infoplists", []))
+    infoplists_by_build_setting["//conditions:default"] = write_info_plists_if_needed(name = name, plists = default_infoplists)
 
     infoplists = select(infoplists_by_build_setting)
 
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -48,13 +48,20 @@ def write_info_plists_if_needed(name, plists):
 
     return already_written_plists + written_plists
 
-def ios_application(name, apple_library = apple_library, **kwargs):
+def ios_application(name, apple_library = apple_library, infoplists_by_build_setting = {}, **kwargs):
     """
     Builds and packages an iOS application.
 
     Args:
         name: The name of the iOS application.
         apple_library: The macro used to package sources into a library.
+        infoplists_by_build_setting: A dictionary of infoplists grouped by bazel build setting.
+
+                                     Each value is applied if the respective bazel build setting
+                                     is resolved during the analysis phase.
+
+                                     If '//conditions:default' is not set the value in 'infoplists'
+                                     is set as default.
         **kwargs: Arguments passed to the apple_library and ios_application rules as appropriate.
     """
 
--- a/rules/library/xcconfig.bzl
+++ b/rules/library/xcconfig.bzl
@@ -177,7 +177,7 @@ def _add_copts_from_option(xcspec, name, option, value, value_escaper, xcconfigs
         for arg in new
     ]
 
-def settings_from_xcconfig(xcconfig):
+def copts_from_xcconfig(xcconfig):
     objc_copts = []
     cc_copts = []
     swift_copts = []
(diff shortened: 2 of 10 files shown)


rules/app.bzl
```diff
@@ -56,7 +57,16 @@ def ios_application(name, apple_library = apple_library, **kwargs):
...
         **kwargs: Arguments passed to the apple_library and ios_application rules as appropriate.
     """
-    infoplists = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
+
+    infoplists_by_build_setting = kwargs.pop("infoplists_by_build_setting", {})
+    for (build_setting, plists) in infoplists_by_build_setting.items():
+        name_suffix = build_setting_name(build_setting)
+        infoplists_by_build_setting[build_setting] = write_info_plists_if_needed(name = "%s.%s" % (name, name_suffix), plists = plists)
+
+    infoplists_by_build_setting["//conditions:default"] = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
```
segiddins, line 66: will this stomp a value that's explicitly passed in for `"//conditions:default"`? If so, should we handle that case?

rules/library/xcconfig.bzl
```diff
@@ -236,3 +236,93 @@ def settings_from_xcconfig(xcconfig):
         ibtool_copts = ibtool_copts,
         linkopts = linkopts,
     )
+
+def build_setting_name(build_setting):
+    """Returns the name of a given bazel build setting from the fully-qualified label
```
segiddins, line 241: why is this necessary? We should always be able to use a full label to refer to any target?

rules/library/xcconfig.bzl
```diff
@@ -236,3 +236,93 @@ def settings_from_xcconfig(xcconfig):
...
+        xcconfig_by_build_setting: A dictionary where the keys are build settings names and
+                                   the values are the respective dictionaries of Xcode build settings
+    Returns:
+        Struct with different copts behind 'select()' statements
+    """
+    xcconfig_with_defaults = xcconfig
+
+    # Adding default values if necessary
+    for (xc_build_setting, value) in fetch_default_xcconfig.items():
+        if not xc_build_setting in xcconfig:
```
segiddins, line 282: what should we do if the setting _is_ in `xcconfig` ?

segiddins (approved):
Have a few small questions, but this looks really good!

rules/library/xcconfig.bzl
```diff
@@ -177,7 +177,7 @@ def _add_copts_from_option(xcspec, name, option, value, value_escaper, xcconfigs
         for arg in new
     ]
 
-def settings_from_xcconfig(xcconfig):
+def copts_from_xcconfig(xcconfig):
```
gyfelton, line 180: curious about why renaming to copts, since it is not just about copts?

gyfelton (commented):
aside from Sam's comments

rules/library/xcconfig.bzl
```diff
@@ -236,3 +236,68 @@ def settings_from_xcconfig(xcconfig):
...
+    objc_copts_by_build_setting = {"//conditions:default": copts_with_defaults.objc_copts}
+    cc_copts_by_build_setting = {"//conditions:default": copts_with_defaults.cc_copts}
+    swift_copts_by_build_setting = {"//conditions:default": copts_with_defaults.swift_copts}
+    momc_copts_by_build_setting = {"//conditions:default": copts_with_defaults.momc_copts}
+    mapc_copts_by_build_setting = {"//conditions:default": copts_with_defaults.mapc_copts}
+    ibtool_copts_by_build_setting = {"//conditions:default": copts_with_defaults.ibtool_copts}
+    linkopts_by_build_setting = {"//conditions:default": copts_with_defaults.linkopts}
+
+    for (build_setting, build_setting_xcconfig) in xcconfig_by_build_setting.items():
+        merged_xcconfig = dict(xcconfig)
```
congt, line 297: Do we want to put xcconfigs common to all build_settings in the select method? Guess probably doesn't matter.

rules/app.bzl
```diff
@@ -56,7 +57,16 @@ def ios_application(name, apple_library = apple_library, **kwargs):
         apple_library: The macro used to package sources into a library.
         **kwargs: Arguments passed to the apple_library and ios_application rules as appropriate.
     """
-    infoplists = write_info_plists_if_needed(name = name, plists = kwargs.pop("infoplists", []))
+
+    infoplists_by_build_setting = kwargs.pop("infoplists_by_build_setting", {})
```
congt, line 61: Could we update the app_doc.md for this argument? It's specific to ios_application, and not supported in rules_apple/ios_application.

rules/app.bzl
congt, line 66: So when a target has both `infoplists_by_build_setting` and `infoplists`, `infoplists` will be used for default condition? Feels a little weird to put default in a separate attribute.

congt (approved):

rules/app.bzl
thiagohmcruz: That's true, since we're accepting any incoming `infoplists_by_build_setting` it makes sense to respect a existing default value if it exists, I'll add the check

rules/library/xcconfig.bzl
thiagohmcruz: This is used to name the files being written [here](https://github.com/bazel-ios/rules_ios/blob/40d41773b30e1b63fa6f87684da25fad54a0f9b3/rules/app.bzl#L64)

Without some kind of unique identifier per build setting rules with the duplicated names were being generated

rules/library/xcconfig.bzl
thiagohmcruz: In that case we want to ignore it unless I'm missing something 🤔 ? The default values should be applied only if they don't exist yet, otherwise `xcconfig` should be respected

rules/library/xcconfig.bzl
thiagohmcruz: I re-named this based on the return value, it's a struct full of copts mostly, I honestly got confused because of the name when I was going through the code for the first time 

But no strong opinions about that, I'm ok with reverting it

rules/app.bzl
thiagohmcruz: Will do 👍

rules/library/xcconfig.bzl
thiagohmcruz: Sorry not sure if I understand the question here, each select needs `xcconfig` received explicitly plus it's own specific xcconfigs plus default values where necessary, this is what is happening next:

```
        merged_xcconfig = dict(xcconfig)

        for (k, v) in build_setting_xcconfig.items():
            merged_xcconfig[k] = v

        for (xc_build_setting, value) in fetch_default_xcconfig.items():
            if not xc_build_setting in merged_xcconfig:
                merged_xcconfig[xc_build_setting] = value
```

If this is not clear maybe I should add some comments there?

rules/app.bzl
thiagohmcruz: Yep, by default the same value as before this change is being used and with the same name `write_info_plists_if_needed(name = name, ...)`, but for each build setting a file with a different name is being written above `write_info_plists_if_needed(name = "%s.%s" % (name, name_suffix), ...)`

If `infoplists_by_build_setting` is not empty it means that the consumer is saying "for the build setting `:blue` use this infoplists array instead of the one in the 'infoplists' attr"

PR #159, opened 2020-11-16
Test

michaeleisel:


meisel/bridge.h
```diff
@@ -0,0 +1 @@
+#import "header.h"
```
segiddins, line 1: ah, I think since this is using quote imports, you might also need to generate an hmap for the target being imported, or else point at the `MyApp_private_hmap` manually

PR #161, opened 2020-11-18
Add "build_with_xcodebuild" attribute to xcodeproj rule

jschear:
This PR is the start of modifying the xcodeproj rule to support generating projects that build with Xcode instead of Bazel.

The resulting Xcode project doesn't successfully build yet; I still need to remove the build script phase that invokes Bazel.

- Adds `build_with_xcodebuild` boolean attr to xcodeproj rule
- Removes the stub build settings from the generated xcode project when `build_with_xcodebuild` is enabled
- Generates a reference project


jschear:
Ah whoops, I meant to open this against a branch in our fork! Sorry for the noise.

PR #164, opened 2020-11-24
Updates apple_library to support the select method for attributes cc_…

congt:
…copts, objc_copts, and swift_copts


segiddins (commented):
Looks good! can we add a test that passes `select`s for those three `copt` attributes to an `apple_framework` target?

congt:
> Looks good! can we add a test that passes `select`s for those three `copt` attributes to an `apple_framework` target?

Done

segiddins (approved):

PR #165, opened 2020-11-25
Add an entry point to xcodeproj rule for overriding project attributes

gyfelton:
### Why this change
Need to provide a way to insert custom attributes for downstream projects
Was thinking about using a `project_overrides` where one can pass `{'attributes': {key: value}, 'options': {key: value}, 'settings' {key: value}}` to cover other fields that we have default values for. But according to Bazel doc https://docs.bazel.build/versions/master/skylark/lib/attr.html#string_dict values for `string_dict` can only be string, so have to break them up into individual fields like the change proposed here.

Please let me know if there is a better way so that we can allow customization done easily for downstream projects.

### Tests
After adding one to a sample project, it indeed gets reflected in the corresponding generated project

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -683,8 +683,11 @@ def _xcodeproj_impl(ctx):
         if _is_current_project_file(f)
     ]
 
+    attributes = {}
+    attributes.update(ctx.attr.project_attributes_overrides)
     xcodeproj_info = struct(
         name = paths.split_extension(project_name)[0],
+        attributes = attributes,
         options = proj_options,
         settings = proj_settings,
         targets = xcodeproj_targets_by_name,
@@ -777,6 +780,7 @@ Tags for configuration:
         "project_name": attr.string(mandatory = False),
         "bazel_path": attr.string(mandatory = False, default = "bazel"),
         "scheme_existing_envvar_overrides": attr.string_dict(allow_empty = True, default = {}, mandatory = False),
+        "project_attributes_overrides": attr.string_dict(allow_empty = True, mandatory = False, default = {}, doc = "Overrides for attributes that can be set at project base level"),
         "generate_schemes_for_product_types": attr.string_list(mandatory = False, allow_empty = True, default = [], doc = """\
 Generate schemes only for the specified product types if this list is not empty.
 Product types must be valid apple product types, e.g. application, bundle.unit-test, framework.
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -778,7 +778,7 @@ Tags for configuration:
         "project_name": attr.string(mandatory = False),
         "bazel_path": attr.string(mandatory = False, default = "bazel"),
         "scheme_existing_envvar_overrides": attr.string_dict(allow_empty = True, default = {}, mandatory = False),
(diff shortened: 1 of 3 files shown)


thiagohmcruz (approved):

thiagohmcruz:
👀 here in case anyone has any ideas, also bumped into this while working on something else. Found more than one discussion similar to this: https://github.com/bazelbuild/bazel/issues/10414

rules/xcodeproj.bzl
```diff
@@ -777,6 +780,7 @@ Tags for configuration:
         "project_name": attr.string(mandatory = False),
         "bazel_path": attr.string(mandatory = False, default = "bazel"),
         "scheme_existing_envvar_overrides": attr.string_dict(allow_empty = True, default = {}, mandatory = False),
+        "project_attributes_overrides": attr.string_dict(allow_empty = True, mandatory = False, default = {}, doc = "Overrides for attributes that can be set at project base level"),
```
segiddins, line 783:
```suggestion
        "project_attributes_overrides": attr.string_dict(allow_empty = True, mandatory = False, default = {}, doc = "Overrides for attributes that can be set at the project base level."),
```

rules/xcodeproj.bzl
```diff
@@ -683,8 +683,11 @@ def _xcodeproj_impl(ctx):
         if _is_current_project_file(f)
     ]
 
+    attributes = {}
+    attributes.update(ctx.attr.project_attributes_overrides)
```
segiddins, line 687: is there any reason why this can't be passed directly into the `struct` initializer below?

segiddins (approved):

segiddins:
> Please let me know if there is a better way so that we can allow customization done easily for downstream projects.

Eventually (in bazel 4 iirc) there's a [module to deserialize JSON](https://docs.bazel.build/versions/master/skylark/lib/json.html#decode), so we could accept a JSON string as the parameter there and serialize an arbitrary data structure to JSON inside a macro which then calls the rule

rules/xcodeproj.bzl
```diff
@@ -683,8 +683,11 @@ def _xcodeproj_impl(ctx):
...
     xcodeproj_info = struct(
         name = paths.split_extension(project_name)[0],
+        attributes = attributes,
```
gyfelton, line 690: do you mean this?
```suggestion
        attributes = ctx.attr.project_attributes_overrides,
```
If so what i have in mind is that the patterm has been have a placeholder at line 686 so we can throw in any default attribute. Also it matches with how things declared for `options` and `settings`.

gyfelton (commented):
> > Please let me know if there is a better way so that we can allow customization done easily for downstream projects.
> 
> Eventually (in bazel 4 iirc) there's a [module to deserialize JSON](https://docs.bazel.build/versions/master/skylark/lib/json.html#decode), so we could accept a JSON string as the parameter there and serialize an arbitrary data structure to JSON inside a macro which then calls the rule

So for now we keep it as is? I just feel with current approach the param list for this rule might explode once we start to need overrides for `options` or `settings` or other configs.

rules/xcodeproj.bzl
gyfelton: since ctx.attr.project_attributes_overrides has a default of `{}`, doing this assignment directly has no harm as well, I will do this after your reply.

segiddins:
> So for now we keep it as is

yeah it works for now, we can always change it / take a more complicated approach later

PR #167, opened 2020-12-02
Add shardcount to ios_unit_test kwargs

dostrander:


segiddins (approved):

congt (approved):

keith:
just curious if any iOS test runners support this?

segiddins:
@keith we (@dostrander) is working on one right now 😉

keith:
what's the thesis around time savings? curious about this vs having many test bundles (but maybe would be helpful for mono test bundles?)

dostrander:
@keith I think having many test bundles would work and give you very similar results, however would put the onus on the individual module owner to divvy up those tests evenly as the module (and the test counts) grow

PR #168, opened 2020-12-04
Xcode 12 support

segiddins:
* Consolidate all xcodeproj testing into a single script

* Run xcodeproj tests on Xcode 12.2

* Set ARCHS instead of VALID_ARCHS so builds in Xcode work on 12.x

* Add support for Xcode projects automatically using the current Xcode

  This means the artifacts from bazel are guaranteed to work with the running Xcode, as they’ll be using the same toolchain


rules/xcodeproj.bzl
```diff
@@ -498,6 +493,8 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
 
         target_settings = {
             "PRODUCT_NAME": target_name,
+            "ARCHS": "$(ARCHS_STANDARD)",
+            "EXCLUDED_ARCHS[sdk=iphonesimulator*]": "arm64",
```
gyfelton, line 497: curious the effect of this on silicon macbook

gyfelton (approved):

rules/xcodeproj.bzl
justinseanmartin: I'd presume we'd need to update this if we need to build for arm64 sims, but we can deal with that later.

justinseanmartin (approved):

congt (approved):

thiagohmcruz (approved):

segiddins:
Hmmm does xcodegen use the current Xcode to determine the SDK or something like that? Any idea how to not get a diff when running the fixtures with different xcode versions @amberdixon ?

Otherwise, this is all gren!

tests/ios/xcodeproj/build.sh
```diff
@@ -6,9 +6,8 @@ cd $(dirname $0)
 xcrun simctl list
 
 export SAMPLE_PROJECT_AND_SCHEME="-project Single-Static-Framework-Project.xcodeproj -scheme ObjcFrameworkTests"
-export SIM_DEVICE_ID=`xcodebuild $SAMPLE_PROJECT_AND_SCHEME -showdestinations | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last"`
+export SIM_DEVICE_ID=$(xcodebuild $SAMPLE_PROJECT_AND_SCHEME -showdestinations -destination "generic/platform=iOS Simulator" | grep "platform:iOS Sim" | head -1 | ruby -e "puts STDIN.read.split(',')[1].split(':').last")
```
amberdixon, line 9: Do you still need to do `grep "platform:iOS Sim"` ?

tests/ios/xcodeproj/pre_build_check.sh
```diff
@@ -2,26 +2,26 @@ set -eu
...
     if [ -f $FULL_PATH ]; then
-        rm $FULL_PATH;
-        echo "Removed file at $FULL_PATH";
+        rm -f $FULL_PATH
+        echo "Removed file at $FULL_PATH"
     fi
 done
 
-export FSP=`grep "FRAMEWORK_SEARCH_PATHS" *.xcodeproj/project.pbxproj | grep -o "bazel-out[^ \\]*"`
+export FSP=$(grep "FRAMEWORK_SEARCH_PATHS" *.xcodeproj/project.pbxproj | grep -o "bazel-out[^ \]*")
```
amberdixon, line 15: Did you remove the double-slash intentionally? Does wrapping this command in `$(` and `)` cause the grep expression to be evaluated differently?

amberdixon (approved):

tests/ios/xcodeproj/build.sh
segiddins: yes

tests/ios/xcodeproj/pre_build_check.sh
segiddins: removed by `shfmt`

amberdixon:
> Hmmm does xcodegen use the current Xcode to determine the SDK or something like that? Any idea how to not get a diff when running the fixtures with different xcode versions @amberdixon ?
> 
> Otherwise, this is all gren!

I don't know the answer to that, but what part of the fixtures look different when you are running with xcode version 12 vs. 11.6?

PR #169, opened 2020-12-06
[proposal] Add the static_framework_generator rule

acecilia:
Hey 👋 

I have a proposal PR here. I understand this may be something not wanted inside this repository: if that is the case, let me know.

**What is included in this PR?**

The `static_framework_generator`. With the introduction of the `apple_framework_packaging` and the `_precompiled_apple_resource_bundle` rules, it is now pretty easy to gather their outputs in order to generate a fully working static framework. This rule does that, and places a copy of the result out of the Bazel cache, so the generated frameworks can be consumed.

My main idea for how to use this rule is (using cocoapods as an example):
1. Build targets with Bazel
2. Use this rule to generate frameworks for them
3. Generate podspec files containing the generated frameworks as vendored_frameworks, so they can be integrated in a cocoapods setup, substituting their source code counterparts (useful to cut build time and indexing time)

There are two major usecases for the above:
1. If you are migrating to Bazel, you need integration with existing build tools: either cocoapods, carthage or swift package manager. All of them accept/support prebuilt frameworks, so you can build with Bazel part of your codebase, and then integrate the resulting frameworks in your existing setup
2. Providing Xcode support. For example: you could have a 100% Bazel setup, and still use cocoapods for project generation (as Bazel Xcode integration is still not 100% afaik). I imagine that the following workflow is interesting:
* Generate prebuilt static frameworks from the bazel cache with this rule
* Generate podspec wrappers for those frameworks
* Use cocoapods to generate the Xcode projects, only keeping in source code the dependencies under development, and using prebuilt static frameworks for the rest.

    If a change on a dependency integrated as a prebuilt framework happens, just re-run this rule so bazel re-builds what is needed and updates the prebuilt frameworks: it should be pretty fast.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -15,7 +15,15 @@ build --test_output=errors # Prints log file output to the console on failure
 
 # By default do not build the tests for sources-with-prebuilt-binaries,
 # because it takes quite some time. They will only run on CI
-build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
+#
+# By default do not run the tests under the static_framework_generator package, 
+# as it requires that tools/static_framework_generator:TargetXIBStaticFrameworkGenerator
+# is run first in order to generate the TargetXIB.framework to be consummed.
+# It will run only on CI
+build --deleted_packages \
+tests/ios/frameworks/sources-with-prebuilt-binaries,\
+tests/ios/static_framework_generator,\
+tools/static_framework_generator
 
 # Enable dbg compilation mode in this repo, so we can test xcodeproj-built
 # binaries contain debug symbol tables.
--- a/rules/apple_patched.bzl
+++ b/rules/apple_patched.bzl
@@ -5,7 +5,7 @@ load(
     apple_dynamic_framework_import_original = "apple_dynamic_framework_import",
     apple_static_framework_import_original = "apple_static_framework_import",
 )
-load("@build_bazel_rules_apple//apple:providers.bzl", "AppleFrameworkImportInfo")
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleFrameworkImportInfo", "AppleResourceInfo")
 load("@build_bazel_rules_swift//swift/internal:providers.bzl", "SwiftUsageInfo")
 
 def apple_dynamic_framework_import(name, **kwargs):
@@ -84,7 +84,7 @@ def _apple_framework_import_modulemap_impl(ctx):
     # is to list here the keys to all of them (you can see the keys for the existing providers of a
     # target by just printing the target)
     # For more information refer to https://groups.google.com/forum/#!topic/bazel-discuss/4KkflTjmUyk
-    other_provider_keys = [AppleFrameworkImportInfo, SwiftUsageInfo, apple_common.AppleDynamicFramework, OutputGroupInfo, DefaultInfo]
+    other_provider_keys = [AppleFrameworkImportInfo, SwiftUsageInfo, apple_common.AppleDynamicFramework, OutputGroupInfo, DefaultInfo, AppleResourceInfo]
     return [new_objc_provider, new_cc_info] + \
            [legacy_target[provider_key] for provider_key in other_provider_keys if provider_key in legacy_target]
 
--- a/rules/static_framework_generator.bzl
+++ b/rules/static_framework_generator.bzl
@@ -1,0 +1,214 @@
+"""This file contains rules to build framework binaries from your podfile or cartfile"""
+
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleResourceInfo")
+
+def static_framework_generator(**kwargs):
+    """
+    A wrapper around static_framework_generator_test which disables the sandbox
+
+    This rule writes its output to the WORKSPACE, which is a location out of the sandbox:
+    in order for bazel to allow that, sandbox must be disabled
+
+    Args:
+        **kwargs: same as the parameters for static_framework_generator_test
+    """
+
+    static_framework_generator_test(
+        tags = ["no-sandbox"],
+        **kwargs
+    )
+
+################
+# Aspect
+################
+
+def _get_resource_targets(deps):
+    return depset(transitive = _get_attr_values_for_name(deps, _TargetInfo, "owned_resource_targets"))
+
+_TargetInfo = provider()
+
+def _get_attr_values_for_name(deps, provider, field):
+    return [
+        getattr(dep[provider], field)
+        for dep in deps
+        if dep and provider in dep
+    ]
+
+def _collect_resources_aspect_impl(target, ctx):
+    resource_targets = []
+    if ctx.rule.kind == "alias":
+        actual = getattr(ctx.rule.attr, "actual")
+        if actual and _TargetInfo in actual:
+            resource_targets += actual[_TargetInfo].resource_targets
+    elif _is_resources_target(ctx):
+        resource_targets.append(target)
+
+    provider = _make_provider(ctx, resource_targets)
+    return [provider]
+
+def _is_top_level_target(ctx):
+    return ctx.rule.kind == "apple_framework_packaging"
+
+def _is_resources_target(ctx):
+    return ctx.rule.kind == "_precompiled_apple_resource_bundle"
+
+def _make_provider(ctx, resource_targets):
+    deps = getattr(ctx.rule.attr, "deps", [])
+    transitive_resource_targets = depset(
+        resource_targets,
+        transitive = _get_attr_values_for_name(deps, _TargetInfo, "transitive_resource_targets"),
+    )
+
+    return _TargetInfo(
+        transitive_resource_targets = depset([]) if _is_top_level_target(ctx) else transitive_resource_targets,
+        owned_resource_targets = transitive_resource_targets if _is_top_level_target(ctx) else depset([]),
+    )
+
+_collect_resources_aspect = aspect(
+    implementation = _collect_resources_aspect_impl,
+    attr_aspects = ["deps", "actual"],
+)
+
+################
+# Rule
+################
+
+def _declare_directory(ctx, framework_name, filename):
+    return ctx.actions.declare_directory("%s/%s/%s" % (ctx.label.name, framework_name, filename))
+
+def _declare_resource_bundles(ctx, target):
+    resource_bundles = []
+    for resource_target in _get_resource_targets([target]).to_list():
+        if AppleResourceInfo in resource_target:
+            for resource_group in resource_target[AppleResourceInfo].unprocessed:
+                resource_files = resource_group[2].to_list()
+                if len(resource_files) == 0:
+                    continue
+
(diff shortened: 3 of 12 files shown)


rules/apple_patched.bzl
```diff
@@ -84,7 +84,7 @@ def _apple_framework_import_modulemap_impl(ctx):
     # is to list here the keys to all of them (you can see the keys for the existing providers of a
     # target by just printing the target)
     # For more information refer to https://groups.google.com/forum/#!topic/bazel-discuss/4KkflTjmUyk
-    other_provider_keys = [AppleFrameworkImportInfo, SwiftUsageInfo, apple_common.AppleDynamicFramework, OutputGroupInfo, DefaultInfo]
+    other_provider_keys = [AppleFrameworkImportInfo, SwiftUsageInfo, apple_common.AppleDynamicFramework, OutputGroupInfo, DefaultInfo, AppleResourceInfo]
```
acecilia, line 87: This is just a quick fix: the AppleResourceInfo was not being propagated correctly

acecilia:
Any feedback? 🙏 @segiddins @ob

.bazelrc
```diff
@@ -15,7 +15,15 @@ build --test_output=errors # Prints log file output to the console on failure
 
 # By default do not build the tests for sources-with-prebuilt-binaries,
 # because it takes quite some time. They will only run on CI
-build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
+#
+# By default do not run the tests under the static_framework_generator package, 
+# as it requires that tools/static_framework_generator:TargetXIBStaticFrameworkGenerator
```
segiddins, line 20: why can't it be built as a dependency?

rules/static_framework_generator.bzl
```diff
@@ -0,0 +1,214 @@
+"""This file contains rules to build framework binaries from your podfile or cartfile"""
+
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleResourceInfo")
+
+def static_framework_generator(**kwargs):
```
segiddins, line 5: I dont think naming the rule `_generator` makes much sense to me. it's implicit that rules generate outputs

rules/static_framework_generator.bzl
```diff
@@ -0,0 +1,214 @@
...
+            runfiles = ctx.runfiles(files = frameworks),
+        ),
+    ]
+
+# Making below rule public in order to properly generate its documentation until the
+# https://github.com/bazelbuild/stardoc/issues/27 issue is resolved
+static_framework_generator_test = rule(
+    implementation = _static_framework_generator_impl,
+    doc = """\
+This generator moves the frameworks artifacts produced by Bazel to a location out of the cache.
```
segiddins, line 177: how is this any different from `bazel build && cp bazel-bin/...` ? I guess this description doesn't explain to my why this rule needs to exist

rules/static_framework_generator.bzl
```diff
@@ -0,0 +1,214 @@
...
+        ),
+    },
+    # Ideally this rule would not be a test: it would be an executable to be run with `bazel run`.
+    # The problem with `bazel run` is that you can only execute one target per `bazel run` invocation.
+    # See: https://github.com/bazelbuild/bazel/issues/10855
+    #
+    # A workaround is to mark this rule as a test, because `bazel test` allows to execute multiple
+    # test targets in paralel. Running `bazel test` once with multiple targets is much faster than
+    # running `bazel run` with one target multiple times
+    test = True,
```
segiddins, line 213: why can't it be a normal build rule, and the user manually copy a zip or directory or something out of `bazel-bin` ?

segiddins (commented):
I'm not seeing how this fits in with the rest of the rules in here -- it doesnt seem to either use or be used by any of the framework rules?

rules/static_framework_generator.bzl
acecilia: I can change the name. I prefer to avoid naming it `ios_static_framework` so the name does not clash with the existing [`ios_static_framework`](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_static_framework).

Any proposals?

rules/static_framework_generator.bzl
acecilia: When running `bazelisk build --apple_platform_type=ios --deleted_packages='' tests/ios/static_framework_generator:TargetXIB` nothing gets created under `bazel-bin`. What gets created is:
* `bazel-out/applebin_ios-ios_x86_64-dbg-ST-78e822f89cdf/bin/tests/ios/static_framework_generator/TargetXIB/TargetXIB.framework`
* I would expect to see the resource bundles also created somewhere, but I cant find them anywhere. Seems like they are not generated when doing `build`

When you `cp` from inside `bazel-bin` or `bazel-out`, it is difficult to reliably tell where the `*.framework` and the resource bundles are located inside `bazel-bin` or `bazel-out`.

This rule solves those issues, serving a fully working framework in a location specified by the consumer

rules/static_framework_generator.bzl
acecilia: The reason is to avoid `the user manually copy a zip or directory or something out of bazel-bin`, and instead do it directly with the rule. 

My understanding is that in order to copy files out of the sandbox, the `DefaultInfo` provider needs to return an `executable` to copy them, and only `executable` or `test` rules can do that. Is that correct, or is there other way?

.bazelrc
acecilia: The rule `tools/static_framework_generator:TargetXIBStaticFrameworkGenerator` generates a framework at the location `tests/ios/static_framework_generator/bazel-generated`.

I decided to test the rule by consuming the framework and using it in a test target (`tests/ios/static_framework_generator:TargetXIBTestsLib`), . To consume the framework I am using `apple_static_framework_import` (`tests/ios/static_framework_generator:TargetXIBStaticFramework`).

The rule `apple_static_framework_import` requires that the framework is present on disk at analysis time. That is why it cant be built as a dependency, and instead it has to be done in two steps:

1. Generate the framework by running `tools/static_framework_generator:TargetXIBStaticFrameworkGenerator`
2. Test that the generated framework can be consumed without issues

I am not aware of a rule similar to `apple_static_framework_import` that does not require the framework to be ready at analysis time.

acecilia:
@segiddins thanks for the feedback :)

It is using [`apple_framework_packaging`](https://github.com/bazel-ios/rules_ios/pull/169/files#diff-d149ced8cb238f11e4dd818f157eeb4c2a201679e420d134c18b7ebfb05f54d9R50) and [`_precompiled_apple_resource_bundle`](https://github.com/bazel-ios/rules_ios/pull/169/files#diff-d149ced8cb238f11e4dd818f157eeb4c2a201679e420d134c18b7ebfb05f54d9R53).

It is possible to consider this rule very similar to [`ios_static_framework`](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_static_framework), but using `apple_framework_packaging` and `_precompiled_apple_resource_bundle` under the hood

ob:
Thanks @acecilia for the PR and @segiddins for the review!

While I'm not opposed to having a packaging rule here, internally we separate our rulesets between building rules and packaging rules. I could see us creating something like `bazel-io/rules_ios_packaging` that maybe incorporated not only frameworks but eventually allowed you to publish SPM packages, Pods, or whatever... what do you guys think?

acecilia:
@ob sounds good, my only concern is that having these rules in a different repository would require:

* Versioning and keeping both repos in sync
* Duplicating the repository setup of `rules_ios` into the new repo. I mean the github work workflow, testing setup and scripts...

What problems do you see happening when keeping everything under the same repository?

rules/static_framework_generator.bzl
```diff
@@ -0,0 +1,214 @@
...
+        ctx.actions.run_shell(
+            outputs = [framework],
+            inputs = existing_framework_files + resource_bundles,
+            command = 'rsync -a "$1/" "$3"; for resource in $2; do rsync -a "$resource" "$3/Resources"; done',
+            arguments = [existing_framework_path, resource_paths, framework.path],
+        )
+        frameworks.append(framework)
+    return frameworks
+
+def _make_destination(ctx):
```
jerrymarino, line 142: Perhaps check if the `destination` is relative VS absolute?

rules/static_framework_generator.bzl
```diff
@@ -0,0 +1,214 @@
...
+It is intended to be used as a tool to ease compatibility with other build systems.
+
+For example, it would be possible to integrate with cocoapods as follows:
+1- Build targets with Bazel
+2- Use this rule to generate frameworks for them
+3- Generate podspec files containing the generated frameworks as vendored_frameworks, so they can be 
+   integrated in a cocoapods setup, substituting their source code counterparts (useful to cut build time)
+""",
+    attrs = {
+        "targets": attr.label_list(
```
jerrymarino, line 187: Does this this cause a reinstall per target? Any reason why we shouldn't have a per target installer?

jerrymarino (commented):
I like the ideal of using Bazel to quickly and incrementally deploy/release/install build artifacts to an external directory which would have to be an input to Bazel 👍 

One question I thought of is there any reason that this should be specific to iOS frameworks or iOS in general? It'd be cool to use such a rule for frameworks and other use cases:
- build mac applications w/bazel and then deploying them to `/usr/local/bin/bazel_out`
- build a dynamic framework but have Xcode link it from `/usr/local/lib`
- build a `sh_binary` and drop it into `/usr/local/bin`
- build an iOS application and run it in the simulator from Xcode from ~/DerivedData 

I like the idea of be general provider that returns the top level artifacts that are `Installable` or `Deployable` as opposed to `_static_framework_generator_impl` that knows the details of the static framework. The other thing that might be useful is to making this more incremental in bazel: e.g., why not have a per output "installer" which accepts all the inputs.

rules/static_framework_generator.bzl
jerrymarino: For the most part, bazel doesn't allow describing no relative outputs in a build rule that isn't in `bazel-out`. To @segiddins point it might be possible to achieve the result of this program in an ad-hoc program that looks at something like `BEP` or even use `BES` to sync build outputs to an external location e.g. for Xcode, this is how xchammer writes bazel artifacts into DERIVED_DATA for Xcode to pick up and then run

acecilia:
Okey, thank you for the feedback. I will close this PR now and apply changes according to the comments. I will try to open other PR in the future, when I get it more polished 🙌

rules/static_framework_generator.bzl
acecilia: Here I chose an array to allow for both options: you can declare the rule once per target (getting one installer per target), or you can declare one rule for multiple targets (one installer for all)

rules/static_framework_generator.bzl
acecilia: Ill try to rework this to place the outputs either in bazel-out or bazel-bin

PR #171, opened 2020-12-09
Supports custom bazel build options for different configs

congt:
It allows xcode to use different build options for debug and release builds.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -652,11 +652,19 @@ def _xcodeproj_impl(ctx):
     proj_settings_debug = {
         "GCC_PREPROCESSOR_DEFINITIONS": "DEBUG",
         "SWIFT_ACTIVE_COMPILATION_CONDITIONS": "DEBUG",
+        "BAZEL_CUSTOM_BUILD_OPTIONS": ctx.attr.custom_build_options_by_config.get("Debug", []),
     }
+
+    # For release config only:
+    proj_settings_release = {
+        "BAZEL_CUSTOM_BUILD_OPTIONS": ctx.attr.custom_build_options_by_config.get("Release", []),
+    }
+
     proj_settings = {
         "base": proj_settings_base,
         "configs": {
             "Debug": proj_settings_debug,
+            "Release": proj_settings_release,
         },
     }
 
@@ -793,6 +801,10 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "installer": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:installer"), cfg = "host"),
         "build_wrapper": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:build-wrapper"), cfg = "host"),
         "additional_files": attr.label_list(allow_files = True, allow_empty = True, default = [], mandatory = False),
+        "custom_build_options_by_config": attr.string_list_dict(allow_empty = True, default = {}, mandatory = False, doc = """\
+The custom build options for different build configs. The key corresponds to the build configuration. Currently we only support `Debug` or `Release`.
+The value is a list of custom Bazel build options for the build configuration.
+"""),
     },
     executable = True,
 )
--- a/tests/ios/xcodeproj/BUILD.bazel
+++ b/tests/ios/xcodeproj/BUILD.bazel
@@ -4,6 +4,13 @@ xcodeproj(
     name = "Single-Static-Framework-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    custom_build_options_by_config = {
+        "Release": [
+            "--define=BUILD_TYPE=release",
+            "--compilation_mode=dbg",
+        ],
+        "Debug": ["--compilation_mode=dbg"],
+    },
     generate_schemes_for_product_types = [
         "framework.static",
         "bundle.unit-test",
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
@@ -313,6 +313,9 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				BAZEL_BUILD_EXEC = "$BAZEL_STUBS_DIR/build-wrapper";
+				BAZEL_CUSTOM_BUILD_OPTIONS = (
+					"--compilation_mode=dbg",
+				);
 				BAZEL_EXECUTION_LOG_ENABLED = 0;
 				BAZEL_INSTALLER = $BAZEL_INSTALLERS_DIR/installer;
 				BAZEL_INSTALLERS_DIR = $PROJECT_FILE_PATH/bazelinstallers;
@@ -385,6 +388,10 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				BAZEL_BUILD_EXEC = "$BAZEL_STUBS_DIR/build-wrapper";
+				BAZEL_CUSTOM_BUILD_OPTIONS = (
+					"--define=BUILD_TYPE=release",
+					"--compilation_mode=dbg",
+				);
 				BAZEL_EXECUTION_LOG_ENABLED = 0;
 				BAZEL_INSTALLER = $BAZEL_INSTALLERS_DIR/installer;
 				BAZEL_INSTALLERS_DIR = $PROJECT_FILE_PATH/bazelinstallers;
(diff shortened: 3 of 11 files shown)


tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -4,6 +4,13 @@ xcodeproj(
     name = "Single-Static-Framework-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    custom_build_options_by_config = {
```
thiagohmcruz, line 7: I wonder if some options should always be there for `debug`/`release`? Is there a use case where `-c opt` for `release` should not be there for example?

Also, maybe the attr name should have the word `bazel` in it to be explicit about what custom options can be added here? Just a thought.

congt: >> some options should always be there for debug/release 
Probably, but it's up to the developer to decide what thy want. The library probably should be agnostic to that.

>> maybe the attr name should have the word bazel
Thought about that, but decided against that because it's a bazel rule. No need to duplicate it again.

rules/xcodeproj.bzl
```diff
@@ -793,6 +801,9 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "installer": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:installer"), cfg = "host"),
         "build_wrapper": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:build-wrapper"), cfg = "host"),
         "additional_files": attr.label_list(allow_files = True, allow_empty = True, default = [], mandatory = False),
+        "custom_build_options_by_config": attr.string_list_dict(allow_empty = True, default = {}, mandatory = False, doc = """\
+The custom build options for different build configs. The key must be `Debug` or `Release`. The value is a list of Bazel build options.
```
gyfelton, line 805: I was wondering if there are any other possible value... but checked xcodegen and they also only explicitly support release and debug 👍

tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -4,6 +4,13 @@ xcodeproj(
...
+        "Release": [
+            "-c",
```
gyfelton, line 9: are these Bazel build options? or this can also accept generic compiler flags? From the doc I assume it means each string must be in a format of `--flag=value`

rules/xcodeproj.bzl
congt: We plan to support release variants, in which case each key will correspond to a release variant.

tests/ios/xcodeproj/BUILD.bazel
congt: `-c opt` is same as `--compliation_mode=opt`

rules/xcodeproj.bzl
```diff
@@ -652,11 +652,19 @@ def _xcodeproj_impl(ctx):
     proj_settings_debug = {
         "GCC_PREPROCESSOR_DEFINITIONS": "DEBUG",
         "SWIFT_ACTIVE_COMPILATION_CONDITIONS": "DEBUG",
+        "BAZEL_CUSTOM_BUILD_OPTIONS": ctx.attr.custom_build_options_by_config.get("Debug", []),
     }
+
+    # For release config only:
```
segiddins, line 658: are the paths and compiler options we're passing for debug config also valid for release?

congt: I think so. I was able to build ver complicate app for release in xcode.

tests/ios/xcodeproj/BUILD.bazel
gyfelton: then why not just match this with debug one?

congt: Wanna test the case that there are multiple build options. I guess I can add another build option.

gyfelton: i see! yup totally makes sense, maybe a comment or more documentation inside this line https://github.com/bazel-ios/rules_ios/pull/171/files#diff-5a136d348b6b1fd482be6ac7e567d20f52b1e808552bcede896900fc65bda48bR805 will help preserve this knowledge

congt: Updated

jerrymarino (commented):
I've added an inline comment regarding having arbitrary build configuration. Any way to push as much of this in a `.bazelrc` as possible? I think it will make matching the CI / CLI options identical in Xcode a bit easier and *require* config normalization

Here is a possible example:
```
xcode_project(
    build_configurations = ["Debug", "Release", "Friday"]
)
```
Then Xcode invokes bazel as:
```
bazel build --define rules_ios.build_configuration=$BAZEL_XCODE_CONFIG
```
Then in your `.bazelrc`
```
Friday: --compilation_mode opt --define FRIDAY=true
```

Then I can make an easy Friday build apart from Xcode in my CI or terminal:
```
bazel build --define rules_ios.build_configuration=Friday
```

jerrymarino:
I think in the contrived example it'd need to pass `config` but it's the same basic gist ( passing a value and matching on that in `.bazelrc`

rules/xcodeproj.bzl
```diff
@@ -652,11 +652,19 @@ def _xcodeproj_impl(ctx):
...
+    proj_settings_release = {
+        "BAZEL_CUSTOM_BUILD_OPTIONS": ctx.attr.custom_build_options_by_config.get("Release", []),
+    }
+
     proj_settings = {
```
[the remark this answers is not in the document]
congt: xcodeproj currently only generate debug and release builds. It will require some work to support arbitrary configs in xcodeproj. @thiagohmcruz will work on that.

congt:
> I've added an inline comment regarding having arbitrary build configuration. Any way to push as much of this in a `.bazelrc` as possible? I think it will make matching the CI / CLI options identical in Xcode a bit easier and _require_ config normalization
> 
> Here is a possible example:
> 
> ```
> xcode_project(
>     build_configurations = ["Debug", "Release", "Friday"]
> )
> ```
> 
> Then Xcode invokes bazel as:
> 
> ```
> bazel build --define rules_ios.build_configuration=$BAZEL_XCODE_CONFIG
> ```
> 
> Then in your `.bazelrc`
> 
> ```
> Friday: --compilation_mode opt --define FRIDAY=true
> ```
> 
> Then I can make an easy Friday build apart from Xcode in my CI or terminal:
> 
> ```
> bazel build --define rules_ios.build_configuration=Friday
> ```

@jerrymarino My original plan is to use `--config` to group build options in .bazelrc file and pass them here.

In the .bazelrc file, we have
```
build:release --compilation_mode opt
build:beta --compilation_mode fastbuild
```

The xcodeproj target will look something like
```
xcodeproj {
  ...
  custom_build_options_by_config = {
    "Release": "--config=release",
    "Beta": "--config=beta"
  }
  ...
}
```
Then Xcode invokes bazel is:
```
bazel build  $BAZEL_CUSTOM_BUILD_OPTION
```
where BAZEL_CUSTOM_BUILD_OPTION is --config=release for release.

The commandline build is also

```
bazel build  --config=release
```

Your proposal simplifies the xcodeproj rule so that we don't need to pass the build options in. But the `--define rules_ios.build_configuration` part concerns me a bit. It's a little implicit and might be a little hard for developers to figure that out.

congt:
> I think in the contrived example it'd need to pass `config` but it's the same basic gist ( passing a value and matching on that in `.bazelrc`

Just saw this comment. It will be better to use `config`. There is still a minor issue: how should we deal with the default config, i.e., `Debug`. It will be a little weird to add an empty Debug config in the .bazelrc file.

jerrymarino:
Yeah my comment is more around _selecting_ a config from VS effectively creating one and the formar will basically enforce people to normalize and make it a bit easier to manage / sync CI with Xcode.

We might need to do a bit of research to determine how a default config could that'd fit into the picture. There might be new features that allow this ( I'm not sure ). In the worst case we could always just require a config to be present for each possibility.

amberdixon (approved):
I am a fan of Jerry's suggestion in the comments, though not sure either how we would specify what configuration value is default .. can we force the developer to specify a value for this configuration and not have a default?

The changes in this PR are acceptable to me as written, so I wouldn't be opposed to merging this and revisiting the configuration question later either. I'll leave it to you to decide!

tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -4,6 +4,13 @@ xcodeproj(
     name = "Single-Static-Framework-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    custom_build_options_by_config = {
+        "Release": [
+            "--define=BUILD_TYPE=release",
+            "--compilation_mode=dbg",
```
gyfelton, line 10: should we be explicit about why this is set for release? (adding a comment here? cos to this can be unexpected)

congt:
Closing in favor for https://github.com/bazel-ios/rules_ios/pull/183

PR #172, opened 2020-12-11
Use an analysis test to validate -swift-version flag

segiddins:
This way, we don't spend time building the framework targets to indirectly do validation


tests/ios/swift-version/swift_version_tests.bzl
```diff
@@ -0,0 +1,45 @@
+load("//rules:framework.bzl", "apple_framework")
+load("@build_bazel_rules_swift//test/rules:action_command_line_test.bzl", "action_command_line_test")
+
+def apple_framework_swift_version_test(*, name, srcs = [":empty.swift"], given, expected, not_expected = []):
+    framework_name = name + "_" + given.replace(".", "_")
```
jerrymarino, line 5: Any reason we _need_ to replace `.` with `_` here?

tests/ios/swift-version/swift_version_tests.bzl
```diff
@@ -0,0 +1,45 @@
...
+            "-swift-version {}".format(expected),
+        ],
+        not_expected_argv = not_expected,
+        tags = [name],
+        target_under_test = swift_library,
+    )
+
+def apple_framework_swift_version_test_suite(name = "apple_framework_swift_version"):
+    apple_framework_swift_version_test(name = name, given = "4", expected = "4")
+    apple_framework_swift_version_test(name = name, given = "4.0", expected = "4")
```
jerrymarino, line 26: So the `swift_version` of this lib is takes an "Xcode SWIFT_VESRION": e.g. a swift version with trailing zeros then putting the `swift` `swift_version` into the CLI. Should we move this into rules_swift? Nice feature BTW :)

jerrymarino (approved):

tests/ios/swift-version/swift_version_tests.bzl
segiddins: i dont think it has a place in `rules_swift` because `swift_library` doesn't have a separate attr for the swift version, it just has `copts`

tests/ios/swift-version/swift_version_tests.bzl
segiddins: I dont think so, but IMO the underscores make for nicer target names

congt (approved):

justinseanmartin (approved):
Great setup for being able to add more of these analysis tests in the future!

PR #173, opened 2020-12-12
Configures TARGETED_DEVICE_FAMILY for targets in xcodeproj

congt:
Xcode depends on TARGETED_DEVICE_FAMILY to configure the target device types.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -1,4 +1,5 @@
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo")
+load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_support")
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//rules:hmap.bzl", "HeaderMapInfo")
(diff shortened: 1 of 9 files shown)


rules/xcodeproj.bzl
```diff
@@ -236,7 +260,7 @@ def _xcodeproj_aspect_impl(target, ctx):
 
 _xcodeproj_aspect = aspect(
     implementation = _xcodeproj_aspect_impl,
-    attr_aspects = ["deps", "actual", "tests", "infoplists", "entitlements", "resources", "test_host"],
+    attr_aspects = ["deps", "actual", "tests", "infoplists", "entitlements", "resources", "test_host", "families"],
```
segiddins, line 263: given that `families` isn't a `label{,list}` attribute, i dont think we need to propagate through it?

rules/xcodeproj.bzl
```diff
@@ -63,6 +63,29 @@ def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
                 hmap_paths.append(relative_path)
     return hmap_paths
 
+def _targeted_device_family(ctx):
```
segiddins, line 66: does it potentially make sense to use `_ui_device_family_plist_value` from https://github.com/bazelbuild/rules_apple/blob/37a2e525fa048cb9b7bf2ce776804f622c5a4262/apple/internal/platform_support.bzl#L58-L70 ?

rules/xcodeproj.bzl
```diff
@@ -63,6 +63,29 @@ def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
...
+    """ The targeted device family of the rule
+
+    Args:
+        ctx: same as what is passed into aspect impl
+    Returns:
+        the targeted device family in the format recognized by xcode
+    """
+    targeted_device_family = None
+    for device_family in getattr(ctx.rule.attr, "families", []):
```
segiddins, line 75: I think this loop could be simplified as

```
mapping = {"iphone": "1", "ipad":"2"}
",".joined([mapping[family] if family in mapping else fail() for family in getattr(ctx.rule.attr, "families", [])])
```

or something more funtional like that

rules/xcodeproj.bzl
congt: Thanks for the pointer!

segiddins (approved):

gyfelton (commented):
niceeeee

PR #174, opened 2020-12-16
Allow test host apps up to iOS 14.3

segiddins:


ndizazzo (approved):

jerrymarino (approved):

jerrymarino:
If this is setting `minimum_os_version`, what are the benefits of making a min version for each OS? e.g. a test host with a min version of `9.4` should also work with `11.0`

segiddins:
Consistency with CocoaPods, which generates an app host with the same minimum OS version as the test bundle

PR #175, opened 2020-12-16
Bump rules_apple and fix deprecated `direct_swiftmodules`

BalestraPatrick:
I'm testing the `xcodeproj` rule in our project and since we're on a newer version of `rules_apple` that removed the already deprecated `direct_swiftmodules`, the `xcodeproj` rule would fail. I fixed this and tested it in a Swift project of ours, and it worked.

I understand that `rules_apple` is creating releases every now and then, so I'm not sure if you'd like to follow that release cadence, or you would be fine with following a commit more at HEAD.

I'm happy to hear your thoughts here.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -164,7 +164,7 @@ def _apple_framework_packaging_impl(ctx):
                 # only thing is the generated module map -- we don't want it
                 continue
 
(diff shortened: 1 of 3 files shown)


segiddins (approved):
Thanks! I'm OK with either tracking releases or HEAD, this is the first time this issue has come up

PR #177, opened 2020-12-17
Handle entitlements_validation if set

thiagohmcruz:
If `entitlements_validation` is set pop it to avoid passing it down to `apple_library` where it shouldn't be used.

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -1,5 +1,6 @@
 load("@bazel_skylib//lib:types.bzl", "types")
 load("@build_bazel_rules_apple//apple:ios.bzl", rules_apple_ios_application = "ios_application")
+load("@build_bazel_rules_apple//apple:common.bzl", "entitlements_validation_mode")
 load("//rules:library.bzl", "apple_library", "write_file")
 load("//rules/library:xcconfig.bzl", "build_setting_name")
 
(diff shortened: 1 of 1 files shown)


congt (approved):

rules/app.bzl
```diff
@@ -76,6 +77,9 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
 
     infoplists = select(infoplists_by_build_setting)
 
+    # Keep original default value 'entitlements_validation_mode.loose' if not set
+    entitlements_validation = kwargs.pop("entitlements_validation", entitlements_validation_mode.loose)
```
segiddins, line 81: should this instead go in `_IOS_APPLICATION_KWARGS` ?

thiagohmcruz: Yep! My bad!

segiddins (approved):

rules/app.bzl
```diff
@@ -10,6 +10,7 @@ _IOS_APPLICATION_KWARGS = [
     "test_host",
     "families",
     "entitlements",
+    "entitlements_validation",
```
congt, line 13: Do you still want to set the default value if it's not set?

segiddins: the default won't be overridden if nothing is passed in

PR #180, opened 2021-01-05
Don't include objc_library generated module maps in packaged frameworks

### LINKED ISSUE #179: Seeing symbol redefinition issues when using a bridging header
Hello There,

I am running into issues with duplicate symbols when using a `swift_objc_bridging_header` in an `ios_application` rule. I have a branch here that showcases the issue when running `//tests/ios/unit-test/test-imports-app:TestImports-Unit-Tests`

https://github.com/bazel-ios/rules_ios/commit/01df46d225e11af057e0496e1739150a9265b265

Any help is appreciated!

segiddins:
Fixes #179, by no longer leaking an incorrect module map


rules/framework.bzl
```diff
@@ -172,6 +172,10 @@ def _apple_framework_packaging_impl(ctx):
 
             # collect modulemaps
             for modulemap in dep[apple_common.Objc].direct_module_maps:
+                if modulemap.owner == dep.label:
```
gyfelton, line 175: to make sure I understand this correctly, for direct module maps of a given obj-c dependency, if the module map is generated by the dep itself, including it will cause redefinition error, so skip them at this point is the way to fix this error

gyfelton (approved):
💯

tinder-maxwellelliott:
Could we expand the test setup to cover more complex situations i.e. Unit Tests with bridging headers that import headers from the tested target? I have made the changes in this commit https://github.com/bazel-ios/rules_ios/commit/1bd0996fbd4ed5393790fa3cf2684884878d9dc4

thiagohmcruz (approved):

tinder-maxwellelliott:
Actually I can break this when using a bridging header with unit tests, see changes here https://github.com/bazel-ios/rules_ios/pull/181

rules/framework.bzl
segiddins: yes

amberdixon (approved):

tinder-maxwellelliott:
Looping back, my import style in my bridging header was breaking this. It would be great if we could also merge in the test updates here https://github.com/bazel-ios/rules_ios/pull/181

PR #183, opened 2021-01-06
Add ability to match bazelrc and Xcode build configurations

thiagohmcruz:
(See discussion in #171) Adds the ability to pass build configurations to the `xcodeproj` rule.

It's now possible to specify `configs` in the `xcodeproj` rule. Each entry will be created as an Xcode build configuration and append a `--config=$CONFIGURATION` to the underlying bazel invocation.

The `xcodegen` tool requires at least one `debug` and `release` configs so they're added by default if not present. We could simply not set this option in `xcodegen` but then a user would have to know which build setting to change in Xcode to change to a different build configuration. So IMO it's a good trade off having to give `xcodegen` what it needs (without any additional underlying consequence) to create a better UX.

**OLD DESCRIPTION FOR REFERENCE ONLY BELOW**:

The idea is that one can set `build_configurations_debug` and `build_configurations_release` to generate `debug`/`release` build configurations and optionally set `build_configurations_debug_bazelrc` and `build_configurations_release_bazelrc` to map those to the respective configs in a `.bazelrc` file.

So, this would simply create build configurations for each type:
```
xcodeproj(
    ...,
    build_configurations_debug = [
        "MyDebugConfig",
    ],
    build_configurations_release = [
        "MyReleaseConfig",
    ],
    ...,
)
```
while this would also append `--config=my_release_config` to the `bazel build` invocation if `MyReleaseConfig` is selected:
```
xcodeproj(
    ...,
    build_configurations_debug = [
        "MyDebugConfig",
    ],
    build_configurations_release = [
        "MyReleaseConfig",
    ],
    build_configurations_release_bazelrc = {
        "MyReleaseConfig": "my_release_config",
    },
    ...,
)
```

Note that `Debug` and `Release` will be used if set and added by default if not.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -623,7 +623,7 @@ def _xcodeproj_impl(ctx):
 
     proj_options = {
         "createIntermediateGroups": True,
-        "defaultConfig": "debug",
+        "defaultConfig": "Debug",
         "groupSortPosition": "none",
         "settingPresets": "none",
     }
@@ -676,7 +676,7 @@ def _xcodeproj_impl(ctx):
     proj_settings = {
         "base": proj_settings_base,
         "configs": {
-            "debug": proj_settings_debug,
+            "Debug": proj_settings_debug,
         },
     }
 
@@ -700,13 +700,11 @@ def _xcodeproj_impl(ctx):
     # The 'xcodegen' tool requires at least one build configuration
     # of each type 'debug' and 'release'. Add those and set all the others to 'none'
     #
-    # Note that the consumer can still set 'debug' and 'release' in 'ctx.attr.configs'
-    # and take advantage of the configs in the .bazelrc file. This is only a way to
-    # be consistent with what 'xcodegen' expects, no additional settings are being added
-    # since 'proj_settings' above has only common settings
+    # Note that the consumer can still set 'Debug' and 'Release' in 'ctx.attr.configs'
+    # and take advantage of the configs in the .bazelrc file.
     xcodeproj_info_configs = {k: "none" for k in ctx.attr.configs}
-    xcodeproj_info_configs["debug"] = "debug"
-    xcodeproj_info_configs["release"] = "release"
+    xcodeproj_info_configs["Debug"] = "debug"
+    xcodeproj_info_configs["Release"] = "release"
 
     xcodeproj_info = struct(
         name = paths.split_extension(project_name)[0],
@@ -806,7 +804,7 @@ Tags for configuration:
         be appended to the underlying bazel invocation. Effectively allowing the configs in the .bazelrc file
         to control how Xcode builds each build configuration.
 
-        If not present 'debug' and 'release' Xcode build configurations will be created by default without
+        If not present the 'Debug' and 'Release' Xcode build configurations will be created by default without
         appending any additional bazel invocation flags.
         """),
         "default_config": attr.string(mandatory = False),
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -807,7 +807,6 @@ Tags for configuration:
         If not present the 'Debug' and 'Release' Xcode build configurations will be created by default without
         appending any additional bazel invocation flags.
         """),
-        "default_config": attr.string(mandatory = False),
         "deps": attr.label_list(mandatory = True, allow_empty = False, providers = [], aspects = [_xcodeproj_aspect]),
         "include_transitive_targets": attr.bool(default = False, mandatory = False),
         "project_name": attr.string(mandatory = False),
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -623,7 +623,7 @@ def _xcodeproj_impl(ctx):
 
     proj_options = {
         "createIntermediateGroups": True,
-        "defaultConfig": "Debug",
+        "defaultConfig": "debug",
         "groupSortPosition": "none",
         "settingPresets": "none",
     }
@@ -640,6 +640,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_INSTALLERS_DIR": "$PROJECT_FILE_PATH/bazelinstallers",
         "BAZEL_INSTALLER": "$BAZEL_INSTALLERS_DIR/%s" % ctx.executable.installer.basename,
         "BAZEL_EXECUTION_LOG_ENABLED": False,
+        "BAZEL_CONFIGS": ctx.attr.configs,
     })
 
     # Stubbding main executable used by xcode so no actual building happening on Xcode side
@@ -671,10 +672,11 @@ def _xcodeproj_impl(ctx):
         "GCC_PREPROCESSOR_DEFINITIONS": "DEBUG",
         "SWIFT_ACTIVE_COMPILATION_CONDITIONS": "DEBUG",
     }
+
     proj_settings = {
         "base": proj_settings_base,
         "configs": {
-            "Debug": proj_settings_debug,
+            "debug": proj_settings_debug,
         },
     }
 
@@ -695,11 +697,23 @@ def _xcodeproj_impl(ctx):
         if _is_current_project_file(f)
     ]
 
+    # The 'xcodegen' tool requires at least one build configuration
+    # of each type 'debug' and 'release'. Add those and set all the others to 'none'
+    #
+    # Note that the consumer can still set 'debug' and 'release' in 'ctx.attr.configs'
+    # and take advantage of the configs in the .bazelrc file. This is only a way to
+    # be consistent with what 'xcodegen' expects, no additional settings are being added
+    # since 'proj_settings' above has only common settings
+    xcodeproj_info_configs = {k: "none" for k in ctx.attr.configs}
+    xcodeproj_info_configs["debug"] = "debug"
+    xcodeproj_info_configs["release"] = "release"
+
     xcodeproj_info = struct(
         name = paths.split_extension(project_name)[0],
         attributes = ctx.attr.project_attributes_overrides,
         options = proj_options,
         settings = proj_settings,
+        configs = xcodeproj_info_configs,
         targets = xcodeproj_targets_by_name,
         schemes = xcodeproj_schemes_by_name,
         fileGroups = project_file_groups,
@@ -785,6 +799,17 @@ Tags for configuration:
     xcodeproj-ignore-as-target: Add this to a rule declaration so that this rule will not generates a scheme for this target
 """,
     attrs = {
+        "configs": attr.string_list(mandatory = False, default = [], doc = """
+        List of bazel configs present in the .bazelrc file that can be used to build targets.
+
+        A Xcode build configuration will be created for each entry and a '--config=$CONFIGURATION' will
+        be appended to the underlying bazel invocation. Effectively allowing the configs in the .bazelrc file
+        to control how Xcode builds each build configuration.
+
+        If not present 'debug' and 'release' Xcode build configurations will be created by default without
+        appending any additional bazel invocation flags.
+        """),
+        "default_config": attr.string(mandatory = False),
         "deps": attr.label_list(mandatory = True, allow_empty = False, providers = [], aspects = [_xcodeproj_aspect]),
         "include_transitive_targets": attr.bool(default = False, mandatory = False),
         "project_name": attr.string(mandatory = False),
--- a/tests/ios/xcodeproj/BUILD.bazel
+++ b/tests/ios/xcodeproj/BUILD.bazel
@@ -4,6 +4,8 @@ xcodeproj(
     name = "Single-Static-Framework-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    # Not that 'configs' must hold names for configs present in your .bazelrc file
+    # in order to build with the respective Xcode build configurations active
     configs = [
         "bar",
         "foo",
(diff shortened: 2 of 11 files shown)


rules/xcodeproj.bzl
```diff
@@ -785,6 +858,38 @@ Tags for configuration:
     xcodeproj-ignore-as-target: Add this to a rule declaration so that this rule will not generates a scheme for this target
 """,
     attrs = {
+        "build_configurations_debug": attr.string_list(
+            mandatory = False,
+            default = [],
+            doc = "List of names for build configurations of type 'debug' to generate. If 'Debug' is not present it will be added.",
+        ),
+        "build_configurations_debug_bazelrc": attr.string_dict(
```
congt, line 866: Could we directly use the build_configurations_debug/build_configurations_release values in build_wrapper and get rid of build_configurations_debug_bazelrc and build_configurations_release_bazelrc?

thiagohmcruz: This would force people to create configs for all specified configurations in the `.bazelrc` file, including `Debug` for example that, from a user perspective, should 'just work'.

Unless there's another way to conditionally append `--config <config name>` to the invocation that happens in the build wrapper?

congt: Yeah, I raised the concern in the original [PR](https://github.com/bazel-ios/rules_ios/pull/171). Here is @jerrymarino 's comment:
```
We might need to do a bit of research to determine how a default config could that'd fit into the picture. There might be new features that allow this ( I'm not sure ). In the worst case we could always just require a config to be present for each possibility.
```
And I didn't find any better way with the default config.

thiagohmcruz: @jerrymarino thoughts here? I feel like it's better to add this as an optional thing (if you want to use what is in your `.bazelrc` then just set `build_configurations_debug/release_bazelrc`, otherwise the rule just creates the build configuration as a "pure" 'debug'/'release' one)

jerrymarino: I'd consider using an API that's as minimal as possible and ideally the user could provide all of this data as a mapping:
```
{
    "Debug": "--config my_debug",
    "ReleaseEnterprise" : "--config my_release_enterprise",
    "ReleaseAppstore" : "--config my_release_appstore",
}
```

Then inside the rule add `Debug=""` and `Release=""` if they're empty.

rules/xcodeproj.bzl
```diff
@@ -695,10 +753,25 @@ def _xcodeproj_impl(ctx):
         if _is_current_project_file(f)
     ]
 
+    # It's necessary to tell xcodegen what is the type ('debug'/'release')
```
jerrymarino, line 756: I'd suggest to keep it agnostic of `Debug` and `Release` if possible. Perhaps we can designate them all as `debug` inside of XcodeGen?

Reason being there's a lot more permutations to build

rules/xcodeproj.bzl
```diff
@@ -785,6 +858,38 @@ Tags for configuration:
...
+            default = {},
+            doc = """
+        Dictionary where each key is the name of a build configuration in 'build_configurations_debug'
+        and the 'value' is the name of the respective configuration in the .bazelrc file.
+
+        Set this attribute if you want the generated Xcode project to append '--config value' when
+        building for each specified build configuration
+        """,
+        ),
+        "build_configurations_release": attr.string_list(
```
jerrymarino, line 877: What do you think about using a dict and taking it's keys as the configs instead of putting the keys in both places?

Also, if possible I'd try to keep away from `Debug` and `Release` idioms to reduce the possible build permutations and allow any config. I'm not sure what this implicates inside of XcodeGen but maybe internally the rule can just use debug and remove any defaults - I'm not sure we'd want XcodeGen defaults anyhow.

rules/xcodeproj.bzl
jerrymarino: Perhaps there's a simpler API - What do you think about a mapping where the user can provide configs they'd like
```
 {
   Basic: "--config my_cofig",
   AppStore: "--config appstore"
}
```
Then insert `Debug = ""` and `Release = ""` if it's required by XcodeGen.

rules/xcodeproj.bzl
thiagohmcruz: I see, I guess we can just set them all to `none` in xcodegen? https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#configs

rules/xcodeproj.bzl
thiagohmcruz: I think xcodegen accepts other values different than `debug`/`release` and it simply doesn't apply any value by default (learned that today tbh).

I'll see if I can remove the `debug`/`release` idiom here, just call it `build_configurations` and see how to separate things inside

rules/xcodeproj.bzl
thiagohmcruz: My only concern with accepting any string here `Basic: "--config my_cofig",` is that this will ultimately being passed to the `build-wrapper.sh` so anything in this string will be passed as a `$BAZEL_RULES_IOS_OPTIONS` in there.

This means that this attribute is really `build_options_by_configuration` or something like that right? We're not just specifying the build configurations that should be created. 

I used your idea [here](https://github.com/bazel-ios/rules_ios/pull/171#pullrequestreview-550484114) as a reference, that's why I went in this path

jerrymarino:
@thiagohmcruz last night I thought of a way that we can simplify the API down to an array of configs as we discussed in https://github.com/bazel-ios/rules_ios/pull/171 .


The user _specifies_ an array of configs strings by name - each `config` strings represent the name of configuration inside of a .bazelrc. This has a few benefits, and is quite a simple / flexible API:
1. there is no management of CLI flags in Bazel. This has the benefit if making it easier to _do the same thing as Xcode_ on the CI, or _do the same thing as CI_ in Xcode. This also reduces confusion for people who want to do a appstore build inside of Xcode identically to the CI, as they can see the identical builds ( e.g. a build with --config appstore )
2. there is no additional flags added to Bazel invocations
3. because of 2. there wouldn't be contention between `rules_ios` defaults, and the users CLI builds. e.g. as today it passes `local_debug_options`, which change the arguments of `swift_library`. With this design and no defaults, that flag would have to go in the Bazelrc

### xcode_project rule API

These configs are defined inside of the Bazelrc. The name `config` is 100% representative of the idiom and CLI argument in bazel: the `--config` flag
```md
xcode_project(
    configs = ["debug", "dogfood"]
)
```

### Behavior
Xcode provides a *some way to build a target for each configuration. Then, when invokes bazel it passes the configuration 
```
bazel build --config $BUILD_CONFIGURATION
```
*_perhaps it's a BUILD_CONFIGURATION - available for each target and scheme_


### Bazel rc example

```
# .bazelrc
build: debug
build: dogfood --config debug --define Dogfood=true
```


### Default behavior
The largest question in https://github.com/bazel-ios/rules_ios/pull/171  was how to handle defaults: when the user doesn't pass `configs`. There are few ways to deal with this but at the core we just don't pass `--config` to the Bazel invocation.
Possible permutations
1. Have a single `Default` or `Bazel` configuration 
2. Use `Debug` and `Release`

thiagohmcruz:
> @thiagohmcruz last night I thought of a way that we can simplify the API down to an array of configs as we discussed in #171 .
> 
> The user _specifies_ an array of configs strings by name - each `config` strings represent the name of configuration inside of a .bazelrc. This has a few benefits, and is quite a simple / flexible API:
> 
> 1. there is no management of CLI flags in Bazel. This has the benefit if making it easier to _do the same thing as Xcode_ on the CI, or _do the same thing as CI_ in Xcode. This also reduces confusion for people who want to do a appstore build inside of Xcode identically to the CI, as they can see the identical builds ( e.g. a build with --config appstore )
> 2. there is no additional flags added to Bazel invocations
> 3. because of 2. there wouldn't be contention between `rules_ios` defaults, and the users CLI builds. e.g. as today it passes `local_debug_options`, which change the arguments of `swift_library`. With this design and no defaults, that flag would have to go in the Bazelrc
> 
> ### xcode_project rule API
> These configs are defined inside of the Bazelrc. The name `config` is 100% representative of the idiom and CLI argument in bazel: the `--config` flag
> 
> ```lisp
> xcode_project(
>     configs = ["debug", "dogfood"]
> )
> ```
> 
> ### Behavior
> Xcode provides a *some way to build a target for each configuration. Then, when invokes bazel it passes the configuration
> 
> ```
> bazel build --config $BUILD_CONFIGURATION
> ```
> 
> *_perhaps it's a BUILD_CONFIGURATION - available for each target and scheme_
> 
> ### Bazel rc example
> ```
> # .bazelrc
> build: debug
> build: dogfood --config debug --define Dogfood=true
> ```
> 
> ### Default behavior
> The largest question in #171 was how to handle defaults: when the user doesn't pass `configs`. There are few ways to deal with this but at the core we just don't pass `--config` to the Bazel invocation.
> Possible permutations
> 
> 1. Have a single `Default` or `Bazel` configuration
> 2. Use `Debug` and `Release`

Awesome! Thx for taking the time to think about this, I'm working on the changes and will ping you as soon as it's ready!

thiagohmcruz:
@jerrymarino made some changes based on your suggestions, let me know your thoughts. Updated the description here and added some comments for context.

thiagohmcruz:
Missed a few things in the tests. Working on it.

rules/xcodeproj.bzl
```diff
@@ -785,6 +790,17 @@ Tags for configuration:
...
+        List of bazel configs present in the .bazelrc file that can be used to build targets.
+
+        A Xcode build configuration will be created for each entry and a '--config=$CONFIGURATION' will
+        be appended to the underlying bazel invocation. Effectively allowing the configs in the .bazelrc file
+        to control how Xcode builds each build configuration.
+
+        If not present 'debug' and 'release' Xcode build configurations will be created by default without
+        appending any additional bazel invocation flags.
+        """),
+        "default_config": attr.string(mandatory = False),
```
congt, line 803: nit: default to `Debug`?

congt (approved):

rules/xcodeproj.bzl
amberdixon: Maybe i'm mistaken, but i'm not seeing the default_config getting used anywhere.

Maybe you could remove this default_config attr and just do:
```
"configs": attr.string_list(mandatory = False, default = ["Debug", "Release"], doc ...
```

tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -4,6 +4,10 @@ xcodeproj(
     name = "Single-Static-Framework-Project",
     testonly = True,
     bazel_path = "bazelisk",
+    configs = [
+        "bar",
```
amberdixon, line 8: nit: Does the bazel build work for these configs? It looks like `bar` and `foo` bazel configurations are not defined anywhere, so i would imagine `bazel build --configuration=bar` would fail. you probably don't need to fix this, but someone might see it in the future and be confused.

amberdixon (approved):

rules/xcodeproj.bzl
thiagohmcruz: You're right! My bad, this is a leftover that shouldn't be here. Also, `configs` here should hold only configs present in a `.bazelrc` file, `["Debug", "Release"]` are created by default as Xcode build configurations only in case the user decided to not specify.

The reason for that is because in order to tell `xcodegen` bellow to create build configurations for each `configs` I have to pass the expected values to `configs` here:

```
//xcodeproj.bzl:709
    xcodeproj_info = struct(
        name = paths.split_extension(project_name)[0],
        attributes = ctx.attr.project_attributes_overrides,
        options = proj_options,
        settings = proj_settings,
        configs = xcodeproj_info_configs,
        targets = xcodeproj_targets_by_name,
        schemes = xcodeproj_schemes_by_name,
        fileGroups = project_file_groups,
    )
```
when I do that, `xcodegen` throws an error if `configs` does not receive at least of build configuration of each type `debug`/`release` (lowercased, this is the type, not the name) so `["Debug", "Release"]` receive those types and the user defined `configs` are set to type `none`.

We have two paths here:

1. The `xcodeproj` API exposes something similar to what `xcodegen` expects (making it more complex/configurable)
2. We keep the `xcodeproj` api minimal for now, and evolve it if necessary (meaning receiving just the list of names and assuming things behind the scenes)

For an initial iteration I'd suggest sticking with `2.` but I'm happy to implement it differently if folks think it's worth it atm.

Sorry for the long explanation @amberdixon I just wanted to give you the context here!

tests/ios/xcodeproj/BUILD.bazel
thiagohmcruz: You're right! Even though this is not being matched at the end it makes sense to be explicit about where those names should be coming from in a real scenario. Thx!

rules/xcodeproj.bzl
amberdixon: I think you can keep the implementation code the same (which automatically sets Debug -> debug and Release -> release in the configs dictionary.) You could arguably change the defaults passed into the configs attr, so that anyone reading the rule definition will know that these defaults are gonna get applied (even if they don't know the exact bazel configuration that they map to.)

thiagohmcruz:
FYI I was able to use a dummy setup to generate a project with only the build configurations that I want. So no `Debug`/`Release` required by default. I'm investigating why, this PR can be simplified once I'm able to do the same here.

thiagohmcruz:
> FYI I was able to use a dummy setup to generate a project with only the build configurations that I want. So no `Debug`/`Release` required by default. I'm investigating why, this PR can be simplified once I'm able to do the same here.

Never mind, it is indeed [required](https://github.com/yonaskolb/XcodeGen/blob/master/Sources/ProjectSpec/SpecValidation.swift#L111-L118), I'll keep these changes then

Issue #186, opened 2021-01-08
Multi architecture support / ios_multi_cpus


jerrymarino:
Hello! I was curious about the status of using `--ios_multi_cpus` in `rules_ios`'s `apple_framework`. I've tried to build the test application with this feature and ended up getting the following error:
```
$ bazel build -s tests/ios/app/App --apple_platform_type=ios --ios_minimum_os=10.2  --ios_multi_cpus=i386,x86_64
ERROR: file 'tests/ios/app/_objs/OnlySources_objc/arc/OnlySources.o' is generated by these conflicting actions:
Label: //tests/ios/app:OnlySources_objc
RuleClass: objc_library rule
Configuration: cbe9f9128c0d84741485fbbdcf191f7fb188abe6501fd879b855daa84346ecf6, 4fa98f0e6e62218748f252ceda045824522286926b99978a6209345974b7a871
Mnemonic: ObjcCompile
Action key: 798966fd4a85d658616d1369ac7cecf95af97d72654776cdfe3b1b02d1cde402
Progress message: Compiling tests/ios/app/OnlySources/OnlySources.m
PrimaryInput: File:[/Users/jmarino/Development/rules_ios[source]]tests/ios/app/OnlySources/OnlySources.m
PrimaryOutput: File:[[<execution_root>]bazel-out/applebin_ios-ios_i386-dbg-ST-e4791b898f40/bin]tests/ios/app/_objs/OnlySources_objc/arc/OnlySources.o
Owner information: ConfiguredTargetKey{label=//tests/ios/app:OnlySources_objc, config=BuildConfigurationValue.Key[cbe9f9128c0d84741485fbbdcf191f7fb188abe6501fd879b855daa84346ecf6]}, ConfiguredTargetKey{label=//tests/ios/app:OnlySources_objc, config=BuildConfigurationValue.Key[4fa98f0e6e62218748f252ceda045824522286926b99978a6209345974b7a871]}
MandatoryInputs: are equal
Outputs: are equal
ERROR: com.google.devtools.build.lib.skyframe.ArtifactConflictFinder$ConflictException: com.google.devtools.build.lib.actions.MutableActionGraph$ActionConflictException: for tests/ios/app/_objs/OnlySources_objc/arc/OnlySources.o, previous action: action 'Compiling tests/ios/app/OnlySources/OnlySources.m', attempted action: action 'Compiling tests/ios/app/OnlySources/OnlySources.m'
INFO: Elapsed time: 0.062s
[...]

PR #188, opened 2021-01-12
[cache][multi-arch] use apple_common.multi_arch_split

### LINKED ISSUE #184: Redundant `swift_library` builds when using apple_framework and ios_application
When using `rules_ios`, I noticed that there's redundant different builds of a `swift_library`.

```
apple_framework(name = "fw", ..)
ios_application(name="app", deps=["fw"])
```

When running the build, there will be 2 swiftmodules built ( 1 per the app, and 1 per the framework )
```
bazel build app fw
```

Additionally it builds 2x of all the transitive swift libraries and also causes issues with cache hits.

jerrymarino:
This PR fixes #184

Today, there's redundant builds of a `swift_library` with `rules_ios` framework
bundle and `rules_apple` bundles like a an app or test.

Given:
```
apple_framework(name = "fw", ..)
ios_application(name="app", deps=["fw"])
```

When running the build, there will be 2 swiftmodules built ( 1 per the app, and 1 per the framework )
```
bazel build app fw
```
Consequently, it builds 2x of all the transitive swift libraries and also
causes issues with cache hits.

The root cause is that multi platform support isn't implemented in rules apple.
To partition up actions, we give the deps the transition
`apple_common.multi_arch_split`. This implicates a different configuration

Note that there is more changes to handle `ios_multi_cpus`
end to end #186

Testing - it doesn't seem possible to test this with an analysis test
with skylib given the way Bazel transtively computes the output
directory, e.g. for a `target_under_test`. When #186 is done adding a
CLI test with `ios_multi_cpus` will assert that it can build in this
way.


tests/ios/app/BUILD.bazel
```diff
@@ -280,3 +281,12 @@ ios_application(
         ":SwiftLib",
     ],
 )
+
+genrule(
+    name = "TopLevelBundles",
+    cmd = "file $(SRCS) && touch $(OUTS)",
```
amberdixon, line 287: Will `file` still succeed even if $SRCS don't exist? On my machine, I think `file` returns 0, even if i pass it nonsense filenames.

How does this genrule test the changes in this PR? That's not clear to me. Thanks!

amberdixon (approved):
The approach makes sense to me, but left an open question about the new test introduced.

tests/ios/app/BUILD.bazel
jerrymarino: This isn't testing anything inside of the PR and `file` is added there mainly to get the `$(SRCS)` on the CLI invocation. It exercises the redundant build path with a single target - we could add assertion on it with #186.

rules/framework.bzl
```diff
@@ -407,6 +408,19 @@ the framework as a dependency.""",
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
             doc = "Needed to allow this rule to have an incoming edge configuration transition.",
         ),
+        "platform_type": attr.string(
+            mandatory = False,
+            doc =
+                """Internal - currently rules_ios uses the dict `platforms`
```
segiddins, line 414: worth validating that either the value is not set or that its set to a platform in the `platforms` dict?

tests/ios/app/BUILD.bazel
segiddins: please add a comment for the target explaining what it's exercising?

segiddins (commented):
> Testing - it doesn't seem possible to test this with an analysis test
with skylib given the way Bazel transtively computes the output
directory, e.g. for a target_under_test. When #186 is done adding a
CLI test with ios_multi_cpus will assert that it can build in this
way.

I think we can use analysis tests to force into whatever config we want, and take multiple targets as deps and check that the subdir of bazel-out that their outputs go into matches?

rules/framework.bzl
```diff
@@ -337,6 +337,7 @@ apple_framework_packaging = rule(
         ),
         "deps": attr.label_list(
             mandatory = True,
+            cfg = apple_common.multi_arch_split,
```
thiagohmcruz, line 340: From the [docs](https://docs.bazel.build/versions/master/skylark/lib/apple_common.html)

```
Use of this transition requires that the 'platform_type' and 'minimum_os_version' string attributes are defined and mandatory on the rule.
```
I see that these attributes are not mandatory here, is that intentional?

tests/ios/app/BUILD.bazel
thiagohmcruz: +1 also spent some time trying to understand the need for this `genrule`

thiagohmcruz (approved):

rules/framework.bzl
jerrymarino: Yep, it's mandatory that these are available on the rule. We might need to have different defaults than what Bazel is applying in some situations, will look more into it

rules/framework.bzl
jerrymarino: We should have this covered here:
https://github.com/bazel-ios/rules_ios/pull/188/files#diff-627e46d92a0f2032c906c4cc6a941fa44d4a85e2f89e83f3700812f66781a9eaR68

rules/framework.bzl
thiagohmcruz: I see, I misunderstood the docs then. I thought that the "mandatory" attr for them should be "True" but they just need to be present then.

jerrymarino:
> I think we can use analysis tests to force into whatever config we want, and take multiple targets as deps and check that the subdir of bazel-out that their outputs go into matches?

Do you have any suggestion on how to use an analysis test to force a config? I couldn't figure out a way to do this without a decently complex aspect, which might be more trouble than its worth. Anyhow, there's a few other issues I've gotta work out with the CI on this still.

segiddins:
@jerrymarino https://github.com/bazelbuild/bazel-skylib/blob/master/docs/unittest_doc.md#analysistest.make-config_settings should allow changing config settings used for analysis of the target under test

congt (approved):

tests/ios/app/BUILD.bazel
jerrymarino: For now, maybe it's easier to just add this along with a test case 🙌 , I'm going to look more into using an aspect in addition to an analysis test to see if we can get an assertion on this

rules/BUILD.bazel
```diff
@@ -41,3 +41,35 @@ config_setting(
         ":local_debug_options_enabled": "True",
     },
 )
+
+config_setting(
```
segiddins, line 45: there are already config settings in rules/apple_platform/BUILD.bazel you should be able to use

rules/framework.bzl
```diff
@@ -30,6 +30,13 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         transitive_deps = library.transitive_deps,
         deps = library.lib_names,
         platforms = library.platforms,
+        platform_type = select({
+            "@build_bazel_rules_ios//rules:macos_build" : "macos",
+            "@build_bazel_rules_ios//rules:ios_build" : "ios",
+            "@build_bazel_rules_ios//rules:tvos_build" : "tvos",
+            "@build_bazel_rules_ios//rules:watchos_build" : "watchos",
+            "//conditions:default": "ios"
```
segiddins, line 38: should this default to macos, since that's bazel's default when no explicit platform type has been specified?

tests/ios/unit-test/test-imports-app/BUILD.bazel
```diff
@@ -20,7 +20,7 @@ ios_application(
         "main.m",
     ],
     bundle_id = "com.example.TestImports-App",
-    families = ["ipad"],
+    families = ["ipad", "iphone"],
```
segiddins, line 23: why this change? it's good to test that we correctly handle the non-default set of families

segiddins:
@jerrymarino it looks like this no longer has anything to do with apple_common.multi_arch_split, did you force-push the wrong branch?

rules/framework.bzl
jerrymarino: I've gotta test it out some more, but I suspect we can allow bazel to set this by it's own means by passing an empty value `""`

jerrymarino:
@segiddins thanks for digging into the bit on analysis test config - this is great. Once I've fixed the misc edges on this I suspect I can come up with an aspect to surface out all `swift_library` outputs and make assertions on the output root in the test. Likely would be a followup to keep this one somewhat straight forward.

rules/BUILD.bazel
jerrymarino: ✅

tests/ios/unit-test/test-imports-app/BUILD.bazel
jerrymarino: ✅  Removed in the latest rev - seemed related to [default test runner selection changes](https://github.com/bazel-ios/rules_ios/pull/188#discussion_r557606756)

PR #196, opened 2021-01-15
Pass through apple_split_cpu

jerrymarino:
This transtion when used in conjunction with rules_apple's transition
causes `apple_split_cpu` to be 100% removed. This PR propagates it to
ensure that `swift_library`'s have the same configuration from the CLI.

This is the last known instance of something that looks like
bazelbuild/bazel#12171 that I could find

Testing - it's not clear how to exactly test this. The analysis test
mentioned in (#188) will have some benefit but this is mainly an issue
with CLI invocations.

For now, please test on the CLI
```
bazel clean
bazel build -s tests/ios/app:AppWithSelectableCopts tests/ios/app:SwiftLib --apple_platform_type=ios
find bazel-out/ -name \*.swiftmodule
```


segiddins:
> Testing - it's not clear how to exactly test this. The analysis test
mentioned in (#188) will have some benefit but this is mainly an issue
with CLI invocations.

Doing an analysis test with the apple_platform_type set in the build config should allow testing here, and I think we should try to add more formal test cases to ensure we don't regress going forward

jerrymarino:
I've been researching ways to test this while coming up with a test for #188. The main delta here is that this one is _only_ happening on the CLI and analysis test as is has that ability to accept a `target_under_test` 🤔

jerrymarino:
cc @amberdixon @segiddins would you mind giving this one a review?

segiddins (approved):

tests/ios/app/BUILD.bazel
```diff
@@ -280,3 +281,5 @@ ios_application(
         ":SwiftLib",
     ],
 )
+
+make_tests()
```
gyfelton, line 285: is it that we can add more tests under this method in future?

jerrymarino: Mainly a target can't be instantiated from a `.bzl` file, and I went off the convention of putting the analysis test definitions adjacent to the impl in the `.bzl` file to keep the BUILD file minimal. Also a nice place to put more instances of `_identical_outputs_test` e.g. for app VS test transitive dep consistency.

tests/ios/app/analysis-tests.bzl
```diff
@@ -0,0 +1,100 @@
...
+    expect_failure = False,
+    attrs = {
+        "deps": attr.label_list(
+            aspects = [_collect_transitive_outputs],
+        ),
+    },
+)
+
+def make_tests():
+    # This test asserts that transitive dependencies have identical outputs for
```
thiagohmcruz, line 67: To @gyfelton's question, also curious to know how we'll add more tests here. Should `make_tests` hold all of them?  If that's the case (nit) should the test description be moved into `_identical_outputs_test `?

jerrymarino: I added a: https://github.com/bazel-ios/rules_ios/pull/196#discussion_r560465686. If you follow the pattern I setup you can add a test for apps / test bundles like:

```
make_tests():

..... 
    _identical_outputs_test(
        name = "test_TestHostEquivilance",
        target_under_test = ":TestAppWithInfoPlist",

        # These inputs *must* be passed seperately, in order to
        # have different transitions applied by Skyframe.
        deps = [":TestAppWithSelectableCopts", ":AppWithSelectableCopts"],
    )
```

but `_identical_outputs_test` cannot be extended as it uses Bazel's analyiss test `analysistest.make`

PR #199, opened 2021-01-22
ObjcProvider compile info => CcCompilationContext

dflems:
https://github.com/bazelbuild/bazel/issues/10674

I'm sort of stumbling through this due to lack of familiarity, but giving it a shot. Comments inline.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -273,12 +273,13 @@ def _apple_framework_packaging_impl(ctx):
     objc_provider_fields = {
         "providers": [dep[apple_common.Objc] for dep in ctx.attr.transitive_deps],
     }
+    compilation_context_fields = {}
 
     if framework_root:
-        objc_provider_fields["framework_search_paths"] = depset(
-            direct = [framework_root],
+        compilation_context_fields["framework_includes"] = depset(
+            direct = [paths.dirname(framework_root)],
         )
-    _add_to_dict_if_present(objc_provider_fields, "header", depset(
+    _add_to_dict_if_present(compilation_context_fields, "headers", depset(
         direct = header_out + private_header_out + modulemap_out,
     ))
     _add_to_dict_if_present(objc_provider_fields, "module_map", depset(
@@ -294,8 +295,6 @@ def _apple_framework_packaging_impl(ctx):
         "multi_arch_linked_binaries",
         "multi_arch_dynamic_libraries",
         "source",
-        "define",
-        "include",
         "link_inputs",
         "linkopt",
         "library",
@@ -305,6 +304,17 @@ def _apple_framework_packaging_impl(ctx):
             transitive = [getattr(dep[apple_common.Objc], key) for dep in ctx.attr.deps],
         )
         _add_to_dict_if_present(objc_provider_fields, key, set)
+
+    for key in ["defines", "includes"]:
+        collected = []
+        for dep in ctx.attr.deps:
+            if CcInfo in dep:
+                collected.append(getattr(dep[CcInfo].compilation_context, key))
+        set = depset(
+            direct = [],
+            transitive = collected,
+        )
+        _add_to_dict_if_present(compilation_context_fields, key, set)
 
     # gather swift info fields
     swift_info_fields = {
@@ -340,7 +350,11 @@ def _apple_framework_packaging_impl(ctx):
     # Eventually we need to remove any reference to objc provider
     # and use CcInfo instead, see this issue for more details: https://github.com/bazelbuild/bazel/issues/10674
     objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
-    cc_info_provider = CcInfo(compilation_context = objc_provider.compilation_context)
+    cc_info_provider = CcInfo(
+        compilation_context = cc_common.create_compilation_context(
+            **compilation_context_fields
+        ),
+    )
     return [
         objc_provider,
         cc_common.merge_cc_infos(direct_cc_infos = [cc_info_provider], cc_infos = [dep[CcInfo] for dep in ctx.attr.transitive_deps if CcInfo in dep]),
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -305,17 +305,6 @@ def _apple_framework_packaging_impl(ctx):
         )
         _add_to_dict_if_present(objc_provider_fields, key, set)
 
-    for key in ["defines", "includes"]:
-        collected = []
-        for dep in ctx.attr.deps:
-            if CcInfo in dep:
-                collected.append(getattr(dep[CcInfo].compilation_context, key))
-        set = depset(
-            direct = [],
-            transitive = collected,
-        )
-        _add_to_dict_if_present(compilation_context_fields, key, set)
-
     # gather swift info fields
     swift_info_fields = {
         "swift_infos": [dep[SwiftInfo] for dep in ctx.attr.transitive_deps if SwiftInfo in dep],
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -211,8 +211,9 @@ def _xcodeproj_aspect_impl(target, ctx):
 
         swift_module_paths = []
         if SwiftInfo in target:
-            swift_defines.append(depset(target[SwiftInfo].direct_defines))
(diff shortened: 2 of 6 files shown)


rules/framework.bzl
```diff
@@ -273,6 +272,17 @@ def _apple_framework_packaging_impl(ctx):
...
+    for key in ["defines", "includes"]:
+        collected = []
+        for dep in ctx.attr.deps:
+            if CcInfo in dep:
+                collected.append(getattr(dep[CcInfo].compilation_context, key))
+        set = depset(
+            direct = [],
+            transitive = collected,
+        )
+        _add_to_dict_if_present(compilation_context_fields, key, set)
```
dflems, line 284: `define` and `include` was removed from `ObjCProvider`. This block of code is based on the block above it that aggregates transitive attributes. I'm not sure if this part is necessary or if the merging of `CcInfo`s below will handle this?

rules/xcodeproj.bzl
```diff
@@ -209,8 +209,9 @@ def _xcodeproj_aspect_impl(target, ctx):
 
         swift_module_paths = []
         if SwiftInfo in target:
-            swift_defines.append(depset(target[SwiftInfo].direct_defines))
-            swift_defines.append(target[SwiftInfo].transitive_defines)
+            swift_info = target[SwiftInfo]
+            swift_defines.append(depset(_collect_swift_defines(swift_info.direct_modules)))
+            swift_defines.append(depset(_collect_swift_defines(swift_info.transitive_modules.to_list())))
```
dflems, line 214: `direct_defines` and `transitive_defines` were removed from `rules_swift`. I lifted this approach for collecting defines from Tulsi.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
```diff
@@ -341,7 +341,7 @@
...
+				FRAMEWORK_SEARCH_PATHS = "$(PLATFORM_DIR)/Developer/Library/Frameworks \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/frameworks/objc/ObjcFrameworkTestLib\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/frameworks/objc/ObjcFramework\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/frameworks/objc/ObjcFrameworkTestLib\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/frameworks/objc/ObjcFramework\"";
```
dflems, line 344: same search paths, just in a slightly different order. not sure if it matters.

jerrymarino: This is an unrelated observation - I think these are included 2x.

jerrymarino:
@dflems nice PR 🥳 will try to give this a review in more depth soon.

rules/xcodeproj.bzl
jerrymarino: So this gives xcode the same invocation as the swift invocation in Bazel? I wonder how it works with `build --features=swift.split_derived_files_generation`, if that is picked up in `transitive_modules` or in any way.

dflems: I think this is more or less functionally identical to how it was before. `rules_swift` just removed the attributes so now you can only collect them yourself (ex: https://github.com/bazelbuild/tulsi/commit/dc84e3943bf2b11da37e30127e3fa830e293d296)

rules/framework.bzl
dflems: The tests seem to pass either way, unless there isn't anything in the tests exercising objc include dirs or defines.

ob: I think this will be handled by the merging of `CcInfo` below. We can just remove the two keys.

rules/xcodeproj.bzl
ob: lgtm.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
ob: Yeah, but they were being included 2x before so we should probably track it in an issue. 
I would hope the order doesn't matter... I think this is just a result of how we are collecting the search paths in the changed code.

ob (approved):
Nice! Thanks for the fix @dflems

dflems:
@ob: removed the redundant application of defines/includes

rules/xcodeproj.bzl
jerrymarino: For sure, that's probably orthogonal to this! I was just curious if this code caused it to pick it up 2x in any way, it looks like the Xcode argument's aren't matching Bazels.

jerrymarino:
Is there any reason it'd be useful to keep around the old code path if people are still running 3.x.x?

dflems:
@jerrymarino: Maybe. Looks like `--incompatible_objc_provider_remove_compile_info` was introduced in 3.4.0. What's the earliest version that this repo supports right now?

segiddins:
Realistically, we're not trying to support multiple bazel versions here. 4.0 and any 3.x is fine

rules/framework.bzl
```diff
@@ -307,7 +306,11 @@ def _apple_framework_packaging_impl(ctx):
     # Eventually we need to remove any reference to objc provider
     # and use CcInfo instead, see this issue for more details: https://github.com/bazelbuild/bazel/issues/10674
     objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
-    cc_info_provider = CcInfo(compilation_context = objc_provider.compilation_context)
+    cc_info_provider = CcInfo(
```
segiddins, line 309: do we need to do something to include the transitive compilation contexts / ccinfos ?

dflems: @segiddins I _think_ the `merge_cc_infos` below does this?

segiddins: 🤦🏻  indeed it does. my bad for not reading more carefully

segiddins (approved):

dflems:
@segiddins @jerrymarino anything blocking this?

rules/framework.bzl
```diff
@@ -294,8 +295,6 @@ def _apple_framework_packaging_impl(ctx):
         "multi_arch_linked_binaries",
         "multi_arch_dynamic_libraries",
         "source",
-        "define",
-        "include",
```
jerrymarino, line 298: QQ: looks like does `framework_includes` takes care of `include` - what about `define`?

jerrymarino (commented):
@dflems this LGTM for the most part, left a quick question inline regrading objc defines If that's fixed in some other way we can pop this in, I wasn't sure if any testing here covered those defines.

jerrymarino:
Also thanks again for helping maintain this, it's much appreciated!

jerrymarino (approved):

rules/framework.bzl
dflems: AFAIK `define` and `include` were migrated to the cc compilation context in 4.0.0:

* `3.7.0` docs: https://docs.bazel.build/versions/3.7.0/skylark/lib/ObjcProvider.html
* `4.0.0` docs: https://docs.bazel.build/versions/4.0.0/skylark/lib/ObjcProvider.html (`define` and `include` removed)

`--incompatible_objc_provider_remove_compile_info` is enabled by default in 4.0.0+. I don't know if we need to have something special in here for people who are still on earlier versions and don't have it set.

ob:
Should we merge this? This is needed for Bazel 4.0 right?

thiagohmcruz:
@jerrymarino ☝️

PR #201, opened 2021-01-26
Set custom path for lldb init file by target and drop Xcode 11

thiagohmcruz:
So consumers (on Xcode 12) don't have to worry about adding `command source ~/.lldbinit-source-map` to their `~/.lldbinit` file anymore.

Also drop Xcode 11. Meaning:
* Use Xcode 12.2 for `Build and Test`, `Check Starlark and Docs` in GH workflow
* Remove `xcodeproj` test using 11.6
* Drop `_LocalDebugOptions`, `local_debug_options` and `local_debug_options_enabled`
* Remove comment about Xcode 11 from `README`
* `carthage` was broken on Xcode 12, added the workaround suggested [here](https://github.com/Carthage/Carthage/blob/master/Documentation/Xcode12Workaround.md)

The path to the new LLDB init file can be found in each scheme action's info tab:

scheme_run_action


tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -10,6 +10,7 @@ xcodeproj(
         "bar",
         "foo",
     ],
+    custom_lldbinit = "fake_lldninit_path",
```
thiagohmcruz, line 13: Do we care about pointing to a real file here?

rules/xcodeproj.bzl
```diff
@@ -806,6 +808,7 @@ Tags for configuration:
         If not present the 'Debug' and 'Release' Xcode build configurations will be created by default without
         appending any additional bazel invocation flags.
         """),
+        "custom_lldbinit": attr.string(mandatory = False, default = "", doc = "Custom path to the .lldbinit file"),
```
segiddins, line 811: why does this need to be customizable? rules_ios is already writing the settings https://github.com/bazel-ios/rules_ios/blob/c8287ff5f1bd9458f075b79849103bedaa5526e4/tools/xcodeproj_shims/install.sh#L103, could the scheme & that file coordinate on a path (e.g. `$CONFIGURATION_TEMP_DIR/$TARGET_NAME.lldbinit`) so it "just works" out of the box?

thiagohmcruz: That makes sense, went on a "natural" route of the user might want to specify this. But in reality having it "just work" is a better approach IMO.

Will push some changes with this in mind!

amberdixon: You can look for references to `bazelinstallers` here. We have some installers (like lldb-settings.sh, swiftmodules.sh, etc.) You can create a new one called `lldbinit` and place it either inside or next to the same bazelinstallers directory. And then the scheme_action_details can reference either the `$PROJECT_FILE_PATH/bazelinstallers/lldbinit` or `$PROJECT_FILE_PATH/lldbinit`.

amberdixon: Note that you can generate the lldbinit file from a template, in the same way we generate the `print_json_leaf_nodes` file and several others ... we basically just copy these files from our xcodeproj_shims into the newly created xcode project directory.

thiagohmcruz: 🙏

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
```diff
@@ -13,11 +13,7 @@ set -euo pipefail
 #
 # The platform settings one points the current working directory (cwd) to workspace root.
 # This is needed for features relying on lldb remote debugging, such as `oso_prefix_is_pwd`.
-#
-# NOTE: In order to use this, add this line to `~/.lldbinit`:
-#
-#     command source ~/.lldbinit-source-map
-cat <<-END > ~/.lldbinit-source-map
+cat > $BAZEL_LLDB_INIT_FILE <<-END
```
thiagohmcruz, line 16: Reason for this change:

```
3.6.6 Here Documents
...
No parameter and variable expansion, command substitution, arithmetic expansion, or filename expansion is performed on word.
```

See https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Here-Documents

rules/xcodeproj.bzl
```diff
@@ -589,6 +592,13 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         for arg in commandline_args_tuple:
             scheme_action_details["commandLineArguments"][arg] = True
 
+        # From https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes
+        #
+        # You can now specify a path for the LLDB init file to use in a Run and Test action.
+        # Configure this path in the Info tab of a scheme’s Run or Test action.
+        # The path can contain a build settings macro such as ${SRCROOT}, so the file can be part of a project. (38677796) (FB5425738)
+        scheme_action_details["customLLDBInit"] = lldbinit_file
```
gyfelton, line 600: should we be explicit that this only supports xcode 12 and above? aka if one uses xcode 11, all these settings won't take effect?

gyfelton (commented):
LGTM to me, just one comment, don't believe i have enough context to approve this tho :)

rules/xcodeproj.bzl
```diff
@@ -540,6 +541,8 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
             ["-D%s" % d for d in target_info.cc_defines.to_list()],
         )
 
+        target_settings["BAZEL_LLDB_INIT_FILE"] = lldbinit_file
```
gyfelton, line 544: so this file is where to put content of `lldb-init-source-map` in, thx @thiagohmcruz !

thiagohmcruz: Yep! Setting it to an env var here so the shell scripts can read it, same path goes to Xcode's generate scheme action.

gyfelton (approved):

rules/BUILD.bazel
```diff
@@ -11,34 +8,3 @@ genrule(
     outs = ["empty.swift"],
     cmd = "touch $(OUTS)",
 )
-
-# A dummy target that enables serialize-debugging-options but only in local development,
-# so that swiftmodule paths will refer to paths on the local machine and LLDB will work
-# in Xcode versions older than 11.5.
-# See details here: https://github.com/ios-bazel-users/ios-bazel-users/blob/master/DebuggableRemoteSwift.md
-swift_library(
-    name = "_LocalDebugOptions",
```
amberdixon, line 20: thank you for removing this!!

amberdixon (approved):
Can you add a note to the README that we only support xcode 12? thanks!

rules/xcodeproj.bzl
amberdixon: it seems like we are intentionally dropping xcode11 support .. I am OK with that, but suggested Thiago update the README accordingly.

PR #205, opened 2021-02-03
Symlink headers and modulemap instead of copy in framework_packaging

thiagohmcruz:
So "jump to definition" jumps to the file under the generated Xcode project instead of the one in `bazel-out`

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -46,6 +46,39 @@ def _find_framework_dir(outputs):
         return prefix + ".framework"
     return None
 
+def _framework_packaging_symlink_headers(ctx, inputs, outputs):
+    inputs_by_basename = {input.basename: input for input in inputs}
+
+    # If this check is true it means that multiple inputs have the same 'basename',
+    # an additional check is done to see if that was caused by 'action_inputs' containing
+    # two different paths to the same file
+    #
+    # In that case fails with a msg listing the differences found
+    if len(inputs_by_basename) < len(inputs):
+        inputs_by_basename_paths = [x.path for x in inputs_by_basename.values()]
+        inputs_with_duplicated_basename = [x for x in inputs if not x.path in inputs_by_basename_paths]
+        if len(inputs_with_duplicated_basename) > 0:
+            fail("""
+                [Error] Multiple files with the same name exists.\n
+                See below for the list of paths found for each basename:\n
+                {}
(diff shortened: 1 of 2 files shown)


rules/framework.bzl
```diff
@@ -63,13 +64,28 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
...
-        arguments = [args],
-        inputs = action_inputs,
-        outputs = outputs,
-        mnemonic = "PackagingFramework%s" % action.title().replace("_", ""),
-    )
+
+    if action in actions_to_symlink:
+        for input in inputs:
+            input_basename = input.basename
+            output = [x for x in outputs if x.basename == input_basename]
```
segiddins, line 71: this is going to lead to `O(n^2)` runtime, would it be possible to make either a list of tuples or a dict of input file -> output file?

thiagohmcruz: That makes sense! Will see how I can improve the time complexity here

amberdixon: Just curious, what is an example of a basename?

segiddins:
Would we also want to do the same for the modulemap?

thiagohmcruz:
@amberdixon I'm trying to ensure the same check that existed before in the python script happens in the rule. I can go back to commits history to see when duplicates happened and that check was necessary (cc @segiddins in case you know the answer to that 🙏 ).

Also @segiddins iterating differently now, some additional checks happen only if an error msg is necessary but otherwise I'm just mapping things in a dict now

One more thing, I can add similar logic to symlink the modulemap but I'd appreciate any ideas on how to validate if the symlinks are doing what they're supposed to do? In other works, from this list
```
    actions = {
        "binary":
            lambda args: _merge_binaries(args.framework_root, args.framework_name, args.inputs),
        "modulemap":
            lambda args: _copy_modulemap(args.framework_root, args.input()),
        "swiftmodule":
            lambda args: _cp(args.input(), args.output()),
        "swiftdoc":
            lambda args: _cp(args.input(), args.output()),
        "clean":
            lambda args: _clean(args.framework_root, args.input(), args.output()),
    }
```
what would be the criteria to symlink or copy files?

rules/framework.bzl
thiagohmcruz: Posted a new comment with answers to these 👇

segiddins (approved):

Issue #202, opened 2021-02-01
Sporadic issue with duplicate Enums/Classes/etc in remote cached CI builds 


tinder-maxwellelliott:
Hello there,

First and foremost thank you for this wonderful library! I am running into an intermittent issue in my CI builds (I cannot reproduce this locally) where I am getting duplicate Enums/Class errors like so:

```
08:52:54 /private/var/tmp/_bazel_ios-node-2/c567ed985a5d3e1ea0400bb5101861f6/execroot/__main__/./Projects/SomeApp/SomeApp/APPLog.h:12:36: error: redefinition of 'APPLogLevel'
08:52:54 typedef NS_CLOSED_ENUM(NSUInteger, APPLogLevel) {
08:52:54                                    ^
08:52:54 /private/var/tmp/_bazel_ios-node-2/c567ed985a5d3e1ea0400bb5101861f6/execroot/__main__/bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-fastbuild-ST-c928d317f2e9/bin/Projects/SomeApp/SomeAppApp_framework_unlinked/SomeApp.framework/Headers/APPLog.h:12:36: note: previous definition is here
08:52:54 typedef NS_CLOSED_ENUM(NSUInteger, APPLogLevel) {
```

Is there some kind of suggested setting for remote cached CI builds to use to prevent this kind of errors? Thanks

PR #209, opened 2021-02-12
Ensures the deletion of stale files in non-sandboxed mode

congt:
# Problem 

We previously symlink both headers and module_map which don't depend on the clean action to delete stale files.
When compiling objc files depending on the current framework, the framework might have stale header files, which will cause the compilation failure.

# The fix

Instead of symlinking module_map, we run the framework_packaging action for the module_map (which copies the file). It depends on the clean action and will trigger it.
As module_map is part of the the objc_provider, the framework_packaging action must be triggered before compiling downstream objc files.

# Other solutions tried but failed

1. Instead of using actions.symlink, I tried to manually create the symlink in framework_packaging.py. But build fails with error as bazel failed to cache the dangling symlink.
2. Added the binary to the objc_provider as static_framework_file. It also failed.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -70,15 +70,6 @@ def _framework_packaging_symlink_headers(ctx, inputs, outputs):
     for (output, input) in output_input_dict.items():
         ctx.actions.symlink(output = output, target_file = input)
 
-def _framework_packaging_symlink_modulemap(ctx, inputs, outputs):
-    if len(inputs) != 1 or len(outputs) != 1:
-        fail("""
-        Multiple .modulemap files found, double check expected inputs and outputs:\n
-        inputs: {}\n
-        outputs: {}
-        """.format([x.path for x in inputs], [x.path for x in outputs]))
-    ctx.actions.symlink(output = outputs[0], target_file = inputs[0])
-
 def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
     if not inputs:
         return []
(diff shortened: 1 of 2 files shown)


segiddins (approved):

rules/framework.bzl
```diff
@@ -245,6 +234,10 @@ def _apple_framework_packaging_impl(ctx):
     binary_out = _framework_packaging(ctx, "binary", binary_in, binary_out, framework_manifest)
     header_out = _framework_packaging(ctx, "header", header_in, header_out, framework_manifest)
     private_header_out = _framework_packaging(ctx, "private_header", private_header_in, private_header_out, framework_manifest)
+
+    # Instead of creating a symlink of the modulemap, we need to copy it to modulemap_out.
+    # It's a hacky fix to guarantee running the clean action before compiling objc files depending on this framework in non-sandboxed mode.
+    # Otherwise, stale header files under framework_root will cause compilation failure in non-sandboxed mode.
```
thiagohmcruz, line 240: Thx for adding the details in here 🙏

thiagohmcruz (approved):

justinseanmartin:
Is there any way we can write a test that will validate this is working as expected and doesn't regress in the future?

congt:
> Is there any way we can write a test that will validate this is working as expected and doesn't regress in the future?

The problem is caused by stale header file from previous commit. I don't know any way to test that. I think the real future-proof fix is to fix the debugging issue in sandbox and turn on sandbox.

segiddins:
The biggest issue with sandboxing is actually the performance hit -- until the cc toolchain & rules_swift get explicit module build support, I don't think sandboxing is really feasible for large repos

jerrymarino (approved):

tinder-maxwellelliott:
Is this related to #202?

congt:
> Is this related to #202?

The error message is quite different, e.g., `<module-includes>:1:1: error: umbrella header for module 'CardReaderSupport' does not include header 'HQR12CardReaderContext.h' [-Werror,-Wincomplete-umbrella]` I guess they're caused by different issues?

PR #212, opened 2021-02-23
Writeup doc on Index While Building

jerrymarino:
Overall architecture of design for "Index While Building" in Bazel

- https://github.com/lyft/index-import/pull/53
- https://github.com/bazelbuild/rules_swift/pull/567
- https://github.com/bazelbuild/rules_swift/issues/561

### DIFF
--- a/docs/index_while_building.md
+++ b/docs/index_while_building.md
@@ -93,9 +93,9 @@ xcbuild ( e.g. XCBuildKit ) protocol messages, in xcconfig by means of the
 
 ### Preliminary
 
-- [x] Profile and determine impact of current state of IWB
-    compilation https://github.com/bazelbuild/rules_swift/issues/561
-    consumption of index in Xcode https://github.com/bazel-ios/rules_ios/issues/203
+- [x] Profile and determine impact of current state of IWB.
+- [x] Swift compilation https://github.com/bazelbuild/rules_swift/issues/561
+- [x] Consumption of index in Xcode https://github.com/bazel-ios/rules_ios/issues/203
 
 ### Determine direction forward with prototypes - spiking
 
@@ -110,7 +110,7 @@ There are a couple avenues here:
 now: it doesn't work with remote caching
 
 2. Disable "Index while building" and correspondingly index-import
-Need to consider the user facing impact
+Need to consider the user facing impact: e.g. faster build times, can't index out-of-focus impl files.
 
 ### Index while building V2 - first pass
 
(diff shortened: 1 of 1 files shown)


docs/index_while_building.md
```diff
@@ -0,0 +1,102 @@
...
+now: it doesn't work with remote caching
+
+2. Disable "Index while building" and correspondingly index-import
+Need to consider the user facing impact  MDX-3735
+
+
+## Land index while building V2 - first pass
+
+1. Propose changes to `rules_swift` for compilation
+2. Implement an aspect to quickly pull remotely built indexes
```
amberdixon, line 97: In rules_ios, we currently look at BEP output to determine which indexstores were modified: https://github.com/bazel-ios/rules_ios/blob/master/tools/xcodeproj_shims/installers/indexstores.sh#L9

Introduced by this commit https://github.com/bazel-ios/rules_ios/commit/609d9e5a68cdcb64bf61a326906b096790d63489

docs/index_while_building.md
```diff
@@ -0,0 +1,102 @@
...
+
+Like the current implementation, index while building V2 piggy backs on the
+`-index-store-path` feature in clang and swift. However, in V2 of index while
+building, `swift` and `clang` compilers use a global index cache internally to
+preserve performance. Finally, to integrate with remote caching, actions copy
+data into `bazel-out`.
+
+### rules_swift workers
+
+Workers are extended to use a global index internally. Then, it writes records
```
jerrymarino, line 34: This is a minor nuance worth clarifying - this component doesn't actually write to the cache: it imports relevant units and records to a tree artifact output file in to Bazel so that it can write to the tree artifact to caches.

docs/index_while_building.md
```diff
@@ -0,0 +1,102 @@
...
+`swift` and `clang` write index data based on the status of what resides in
+index-store-path. When using a transient per-swift-library index, it writes O(
+M imports * N libs) indexer data and slows down compilation significantly: to
+the tune of 300% slow down and 6GB+ index data in benchmarks. `rules_swift`
+likes to use per `swift_library` data in order to remote cache indexes.
+`rules_ios` uses a per `apple_library` index to reduce `index-import` wallclock
+time.  Adding "Index while building" to Bazel needs special consideration to
+both preserve performance characteristics of the original architecture of a
+global index and interoperate with Bazel's remote execution and caching.
+
```
jerrymarino, line 23: It might be useful to spell out how this design saves so much bandwidth on the cache VS current and the special sauce in https://github.com/lyft/index-import/pull/53 to computing _what_ units and records to cache.

docs/index_while_building.md
jerrymarino: Initially we might able to use that that with the current rev of https://github.com/bazelbuild/rules_swift/pull/567 without more effort 👍 .

Longer term the aspect can import indexes while we're waiting for deps to compile - VS waiting until the end. If we go with a BEP reader that is running during the build like in xcbuildkit we might be able tp piggy back on that https://github.com/jerrymarino/xcbuildkit/blob/master/Examples/BazelBuildService/main.swift#L19

segiddins (approved):

congt (approved):

PR #213, opened 2021-02-24
Pass FRAMEWORK_SEARCH_PATHS to target.swift-extra-clang-flags

thiagohmcruz:
In some cases `lldb` does not work properly with swift files in mixed swift/objc projects (empty console after command, lldb errors when evaluating expressions at breakpoints, etc).

Enabled the `lldb` logs in such a project and noticed that `SwiftASTContext` would fail to load modules for different reasons, noticed a pattern on the errors and it was always related to not being able to load `modulemap` or `swiftmodules` or `bridging headers`, the kind of thing that usually it's located under a `X.framework` folder.

We're already setting the values from `FRAMEWORK_SEARCH_PATHS` to `target.swift-framework-search-paths` so this shouldn't be happening, then noticed that `SwiftASTContext` apparently uses what is in `target.swift-extra-clang-flags` (see [this](https://github.com/apple/swift-lldb/blob/stable/source/Symbol/SwiftASTContext.cpp#L1556)) and decided to pass the values in `FRAMEWORK_SEARCH_PATHS` to `target.swift-extra-clang-flags` as well.

This indeed fixed the issues I was seeing

To avoid regressions this PR also adds an additional check in `tests/ios/xcodeproj/post_build_check.sh` to ensure all `FRAMEWORK_SEARCH_PATHS` are going to be available to lldb (meaning they should be in `BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS`)


justinseanmartin:
Also found this article that seems like it is potentially related to the same sort of root cause, though leads to different failure message:
https://github.com/steipete/steipete.com/blob/6478fb444b913366fb6c55608ff5470917fe86d7/_posts/2020-06-04-couldnt-irgen-expression.md

rules/xcodeproj.bzl
```diff
@@ -542,8 +542,19 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
...
+
         target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(
-            ["-D%s" % d for d in target_info.cc_defines.to_list()],
+            ["-D%s" % d for d in target_info.cc_defines.to_list()] +
+            # Passing framework search paths via -F flag so
+            # LLDB has the necessary information to create the ASTs
+            #
+            # Fixes an issue where the debugger output was empty in Xcode
+            # when debugging .swift files
+            ["-F%s" % f for f in fmw_search_paths_lldb],
```
segiddins, line 557: should we quote the search paths, since they can have spaces?

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
```diff
@@ -318,7 +318,7 @@
 				BAZEL_BUILD_TARGET_LABEL = "tests/ios/frameworks/objc:ObjcFrameworkTestLib";
 				BAZEL_BUILD_TARGET_WORKSPACE = build_bazel_rules_ios;
 				BAZEL_LLDB_INIT_FILE = $CONFIGURATION_TEMP_DIR/ObjcFrameworkTestLib.lldbinit;
-				BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS = "";
+				BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS = "-F$(PLATFORM_DIR)/Developer/Library/Frameworks -F\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/frameworks/objc/ObjcFramework";
```
segiddins, line 321: the escaped opening quote is unmatched?

rules/xcodeproj.bzl
thiagohmcruz: Makes sense, I'll do that 👍

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj
thiagohmcruz: Missed that 🤔 , I'll take a look

rules/xcodeproj.bzl
```diff
@@ -542,8 +542,19 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = " ".join(
             ["\"%s\"" % d for d in defines_without_equal_sign],
         )
+
+        # Without `strip` here the paths won't be created
+        # correctly in the LLLDB init file
+        fmw_search_paths_lldb = target_settings["FRAMEWORK_SEARCH_PATHS"].strip("\"").split(" ")
```
segiddins, line 548: insteaof reading this out of the settings and having to split the string, should we get the raw array of the search paths?

thiagohmcruz: Yep, as I was bringing back the quoted paths I thought of the same thing. Working on exactly his now 👍

segiddins (approved):

PR #216, opened 2021-02-25
Use file to 'Pass FRAMEWORK_SEARCH_PATHS to target.swift-extra-clang-flags' instead

thiagohmcruz:
Same as #213 but passing it via a file.

In big projects Xcode will complain about the size of the build setting holding the list of `FRAMEWORK_SEARCH_PATHS`. So this PR uses a file to pass that to LLDB.


thiagohmcruz:
The docs check is failing with:

```
Run brew install buildifier
  brew install buildifier
  shell: /bin/bash -e {0}
Error: buildifier has been disabled because it does not build!
Error: Process completed with exit code 1.
```

Never seen this before 🤔

amberdixon (approved):

segiddins (approved):

justinseanmartin:
> buildifier has been disabled because it does not build!

I rekicked the build and it failed a second time in the same way, so seems like something we're going to need to fix.

tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -21,10 +21,39 @@ settings set target.sdk-path $SDKROOT
...
+  # passing it directly from file to file
+  #
+  # Manually solving 'BAZEL_WORKSPACE_ROOT' and 'PLATFORM_DIR' for now 
+  # until it's possible to pass bigger build settings to Xcode
+  #
+  # For context see: https://github.com/bazel-ios/rules_ios/pull/216
+  FSPS_WITH_ROOT_RESOLVED=${FSPS//"\$BAZEL_WORKSPACE_ROOT"/$BAZEL_WORKSPACE_ROOT}
+  FSPS_WITH_ROOT_RESOLVED=${FSPS_WITH_ROOT_RESOLVED//"\$(PLATFORM_DIR)"/$PLATFORM_DIR}
+  LLDB_SWIFT_EXTRA_CLANG_FLAGS+=($FSPS_WITH_ROOT_RESOLVED)
+fi
```
amberdixon, line 50: What if instead of this approach, you do the following:

```
for [[ fsp in $FRAMEWORK_SEARCH_PATHS ]]; do
 $LLDB_SWIFT_EXTRA_CLANG_FLAGS += " -F$fsp"
done
```

Would that work? That way you could get rid of the `$BAZEL_LLDB_FRAMEWORK_SEARCH_PATHS_FILE` and `BAZEL_LLDB_FRAMEWORK_SEARCH_PATHS` altogether.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
```diff
@@ -21,10 +21,39 @@ settings set target.sdk-path $SDKROOT
...
+  FSPS_WITH_ROOT_RESOLVED=${FSPS_WITH_ROOT_RESOLVED//"\$(PLATFORM_DIR)"/$PLATFORM_DIR}
+  LLDB_SWIFT_EXTRA_CLANG_FLAGS+=($FSPS_WITH_ROOT_RESOLVED)
+fi
+# Set all swift-extra-clang-flags if 'LLDB_SWIFT_EXTRA_CLANG_FLAGS'
+# is not empty
+if [[ ${#LLDB_SWIFT_EXTRA_CLANG_FLAGS[@]} -ne 0 ]]
 then
   cat >> $BAZEL_LLDB_INIT_FILE <<-END
-settings set -- target.swift-extra-clang-flags $BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS
+settings set -- target.swift-extra-clang-flags ${LLDB_SWIFT_EXTRA_CLANG_FLAGS[@]}
```
jerrymarino, line 56: I was pondering Tulsi/XCHammer doesn't need this. I expect it is because we already wrote the search paths the xcode project.

rules/xcodeproj.bzl
```diff
@@ -544,14 +545,24 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
...
+        # Passing framework search paths via -F flag so
+        # LLDB has the necessary information to create the ASTs
+        #
+        # Fixes an issue where the debugger output was empty in Xcode
+        # when debugging .swift files
+        #
+        # Writing to a file and loading this in 'tools/xcodeproj_shims/installers/lldb-settings.sh'
+        # because for big projects Xcode will complain about the size of 'BAZEL_LLDB_FRAMEWORK_SEARCH_PATHS_FILE'
+        lldb_framework_search_paths_filename = "%s_BAZEL_LLDB_FRAMEWORK_SEARCH_PATHS_FILE" % (target_name)
+        lldb_framework_search_paths_file = ctx.actions.declare_file(lldb_framework_search_paths_filename)
```
jerrymarino, line 558: Does this work when debugging on a device? I don't think you're able to give it both ARM and x86_64 search paths in this way.

jerrymarino (commented):
This seems fine to me but left a question for on-device usage.

jerrymarino (dismissed):
Left comment on device builds, it'd be good to get some tests on this too.

tools/xcodeproj_shims/installers/lldb-settings.sh
thiagohmcruz: You're right! This could work @amberdixon, just one thing concerns me 

We need to assert that all `FRAMEWORK_SEARCH_PATHS` are being passed to LLDB, I'd have to change the check in [post_build_check](https://github.com/bazel-ios/rules_ios/blob/master/tests/ios/xcodeproj/post_build_check.sh) to find the generated `.lldbinit` files and check if it's content contains `FRAMEWORK_SEARCH_PATHS` being set to the correct flag (`swift-extra-clang-flags`)

Atm when Xcode builds and the installer is called we put these files under `CONFIGURATION_TEMP_DIR`, are these generated during the tests here in `rules_ios` so I can find the `.lldbinit` files and figure out a way to inspect `swift-extra-clang-flags`?

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
thiagohmcruz: I'd definitely need some time kind of exploring the project generated by other tools to answer that question, not sure how lldb would work in this particular scenario

rules/xcodeproj.bzl
thiagohmcruz: Good call, I'll test that now

amberdixon: We probably have a similar issue right now (even without this change) where autocomplete might not work if you build for device, but not for simulator, because the framework search paths may be pointing to the simulator search paths.

jerrymarino: Ah yeah makes sense to me. Overall, I'd be curious if adding `apple_framework` in XCHammer gets if device builds working. There's also new build system, extensions, device testing, and other issues it may unlock once `apple_framework` is up and running.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
jerrymarino: I'm curious too. If `FRAMEWORK_SEARCH_PATHS` - if it's an argv problem then maybe you'll need the args file with said build.

That being said, perhaps pointing tools at a single directory would improve performance because the tools won't have so many directories to search - e.g. a single "Built products directory" or `/System/Library/Frameworks` exemplify this line of thinking.

rules/xcodeproj.bzl
thiagohmcruz: Confirmed that building for device has the same behaviour before and after this PR. Build succeeds with an error tip in Xcode saying
```
Could not find module 'X' for target 'arm64-apple-ios'; found: x86_64
```
So I won't fix this as part of this PR

jerrymarino: Ok so it was broken before because it gave it the search paths for a simulator. I'd suggest confirming this issue won't happen if you use the right system dir, I wouldn't be surprised that it isn't hitting the error because it first fails on the system module but haven't tested it.

PR #217, opened 2021-02-25
Implement 'first pass' of index while building V2

jerrymarino:
This concludes the first step of mitigating perf issues and first PR in
the series for index while building - docs/index_while_building.md.

Hinging on the feature, swift.index_while_building_v2, it moves
rules_ios to use a global index and pulls in the swift PR
https://github.com/bazelbuild/rules_swift/pull/567, and index-import PR
https://github.com/lyft/index-import/pull/53 .

In hopes of mitigating performance problems of processing the global
index in this way it flips the `-incremental` bit.
Longer term we will not be processing a global index, per
docs/index_while_building.md so this is temporary measure.

This PR implements line items from 'Index while building V2 - first
pass' in the roadmap for this feature defined in
docs/index_while_building.md.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -26,3 +26,7 @@ build --strategy=SwiftCompile=worker
 
 # This flips index_while_building_v2 on
 build --features swift.index_while_building_v2
+
+# Default WMO enabled. As a followup rules_ios should have variant with and
+# without this enabled.
+build --swiftcopt -whole-module-optimization
--- a/.bazelrc
+++ b/.bazelrc
@@ -21,6 +21,8 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # binaries contain debug symbol tables.
 build --compilation_mode=dbg
 
+# Use 'worker' strategy for swift compilation
+build --strategy=SwiftCompile=worker
+
 # This flips index_while_building_v2 on
 build --features swift.index_while_building_v2
-
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,3 +20,7 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # Enable dbg compilation mode in this repo, so we can test xcodeproj-built
 # binaries contain debug symbol tables.
 build --compilation_mode=dbg
+
+# This flips index_while_building_v2 on
+build --features swift.index_while_building_v2
+
--- a/.bazelrc
+++ b/.bazelrc
@@ -24,5 +24,6 @@ build --compilation_mode=dbg
 # Use 'worker' strategy for swift compilation
 build --strategy=SwiftCompile=worker
 
-# This flips index_while_building_v2 on
+# This flips index_while_building_v2 - see docs/index_while_building.md for a
+# detailed summary
 build --features swift.index_while_building_v2
--- a/.bazelrc
+++ b/.bazelrc
@@ -26,7 +26,3 @@ build --strategy=SwiftCompile=worker
 
 # This flips index_while_building_v2 on
 build --features swift.index_while_building_v2
-
-# Default WMO enabled. As a followup rules_ios should have variant with and
-# without this enabled.
-build --swiftcopt -whole-module-optimization
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,3 +4,10 @@ sh_binary(
     name = "buildifier",
     srcs = ["@buildifier.mac//file"],
 )
+
+config_setting(
+    name = "index_while_building_v2",
+    values = {
+        "features": "swift.index_while_building_v2",
+    },
+)
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,5 +1,7 @@
 # Pull buildifer.mac as an http_file, then depend on the file group to make an
 # executable
+load("@build_bazel_rules_swift//swift/internal:feature_names.bzl", "SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2")
+
 sh_binary(
     name = "buildifier",
     srcs = ["@buildifier.mac//file"],
@@ -8,6 +10,6 @@ sh_binary(
 config_setting(
     name = "index_while_building_v2",
     values = {
-        "features": "swift.index_while_building_v2",
+        "features": SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2,
     },
 )
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -664,17 +664,31 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         lib_names.append(cpp_libname)
 
     additional_objc_copts.append("-I.")
+    index_while_building_objc_copts = select({
+        "@build_bazel_rules_ios//:index_while_building_v2": [
+            # Note: this won't work work for remote caching yet. It uses a
+            # _different_ global index for objc than so that the BEP grep in
+            # rules_ios picks this up.
+            # Checkout the task roadmap for future improvements:
+            # Docs/index_while_building.md
+            "-index-store-path",
+            "bazel-out/rules_ios_global_index_store.indexstore",
+        ],
+        "//conditions:default": [
+            "-index-store-path",
+            "$(GENDIR)/{package}/rules_ios_objc_library_{libname}.indexstore".format(
+                package = native.package_name(),
+                libname = objc_libname,
+            ),
+        ],
+    })
 
-    additional_objc_copts.extend(("-index-store-path", "$(GENDIR)/{package}/rules_ios_objc_library_{libname}.indexstore".format(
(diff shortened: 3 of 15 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+config_setting(
+    name = "index_while_building_v2",
+    values = {
+        "features": "swift.index_while_building_v2",
```
segiddins, line 4: load the feature name from rules_swift instead of hard-coding the string?

rules/repositories.bzl
```diff
@@ -53,13 +53,15 @@ def rules_ios_dependencies():
         sha256 = "1fff3fa1e565111a8f678b4698792101844f57b2e78c5e374431d0ebe97f6b6c",
     )
 
+    # Note: this is a branch of the rules_swift PR
+    # https://github.com/bazelbuild/rules_swift/pull/567
```
congt, line 57: Do you plan to merge this PR after merging the rules_swift PR?

congt (approved):

rules/repositories.bzl
jerrymarino: @congt we ended up deciding to to use a short lived fork of these dependencies to let this bake and reap the benefits while upstream refactoring is worked out.

BUILD.bazel
jerrymarino: 👍  hope we can do this, I've gotta look to see if those symbols are exported

rules/repositories.bzl
justinseanmartin: I'm so sorry, I think there was some miscommunication in that meeting on Wednesday. I talked with @segiddins and @amberdixon and we'd heard different things, so my apologies for not ensuring we were all on the same page about next steps.

I'm *super* excited to get this landed and the immediate impact it will have. Is there an impediment to upstreaming the changes, or is it the work to address PR feedback and gathering consensus on the approach? Are we concerned there will be pushback overall?

In general, we've taken the approach of attempting to upstream changes first, and only resorting to forking if we can't get the changes landed. If we hit a wall upstreaming the changes, then we should reevaluate. My perspective is that it is harder to unfork things as time goes on, and there isn't as much incentive to unfork, which is more likely to lead to a fork becoming long lived (which has associated maintenance cost). I'm aware that this means it takes longer to realize the impact of the work. I'm definitely open to other perspectives here, both from you and the rest of the team.

jerrymarino: I agree to getting this upstreamed and want to get it working for all Bazel users. We should try to work with the maintainers of rules_swift to get it landed. That being said, the PR has been open at rules_swift since last Friday which should be upstream able. I also mentioned the short lived fork in our plan to move this forward https://github.com/bazel-ios/rules_ios/pull/212/files#diff-26cd539e3b611f02c1703c8fce4656a533b33dc08a4479ab5bfa7523260fa292R123 

@segiddins can you list the required changes to land this? There's been a lot of ideas shared and back and forth on the PR

jerrymarino: @justinseanmartin Here is what I propose for the next steps: 
1. keep the temp fork of rules_swift/index-import until everyone feels good about landing the PR ( and weeks of testing internally )
2. land the PR in rules_ios that points at the forks of index-import and rules_swift
3. point repos at HEAD of rules_ios after the PR lands

Overall we might want to have a larger discussion with the community around landing PRs and features into rules_swift behind feature flags. It's a great way to share ideas. Not everyone has access to work that way ATM.

justinseanmartin: Overall, we're in agreement with the commitment to actively work to get  unforked in the near term (weeks not months), taking an active role in reaching out and collaborating with `rules_swift` and `index-import` maintainers.

jerrymarino: @justinseanmartin SGTM -  I'm fully in support to getting this un-forked in some form 👍 

Longer term I hope we can get a flag to handle this in swift / clang: that'd be the ideal state for this specific issue. The main reason we need to the helper binary ( e.g. my PR to `index-import` ) in `rules_swift` is that we can't tell the compiler to index only impl files. We'll also need to wait for that release cycle if we go with the swift/clang approach, that is if we could get it landed in there.

.bazelrc
```diff
@@ -20,3 +20,13 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
...
+
+# Use 'worker' strategy for swift compilation
+build --strategy=SwiftCompile=worker
+
+# This flips index_while_building_v2 on
+build --features swift.index_while_building_v2
+
+# Default WMO enabled. As a followup rules_ios should have variant with and
+# without this enabled.
+build --swiftcopt -whole-module-optimization
```
amberdixon, line 32: what is the reason for enabling whole-module-optimization? Was this to address another bug that was exhibited?

rules/library.bzl
```diff
@@ -664,17 +664,31 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
     additional_objc_copts.append("-I.")
+    index_while_building_objc_copts = select({
+        "@build_bazel_rules_ios//:index_while_building_v2": [
+            # Note: this won't work work for remote caching yet. It uses a
+            # _different_ global index for objc than so that the BEP grep in
+            # rules_ios picks this up.
+            # Checkout the task roadmap for future improvements:
+            # Docs/index_while_building.md
+            "-index-store-path",
+            "bazel-out/rules_ios_global_index_store.indexstore",
```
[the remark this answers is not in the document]
amberdixon: > additionally, does the bumped version of index-import alleviate the issues we saw of re-mapping being done per-indexstore as opopsed to per-unit? that was the original motivation for separating them out.

It looks like Jerry's passing the `-incremental` flag to index-import which *should* alleviate the problems we saw with using a global objc indexstore.

Jerry, could you make sure to test out a build which compiles a lot of obj-c files? Ensure that index-import runs fast on rebuilds of those projects (because on rebuilds, index-import should not be moving or remapping any unit files at all.)

amberdixon (approved):

rules/library.bzl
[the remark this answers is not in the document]
jerrymarino: It should point at `bazel-out` to make sure there is a singe index store.

@amberdixon until we implement `Local optimization - sans index-import`, it still has to remap https://github.com/bazel-ios/rules_ios/blob/master/docs/index_while_building.md#local-optimization---sans-index-import

We should have dropped atleast an order of magnitude of index data it needs to process with this PR though

rules/library.bzl
```diff
@@ -664,17 +664,31 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         lib_names.append(cpp_libname)
 
     additional_objc_copts.append("-I.")
+    index_while_building_objc_copts = select({
+        "@build_bazel_rules_ios//:index_while_building_v2": [
+            # Note: this won't work work for remote caching yet. It uses a
+            # _different_ global index for objc than so that the BEP grep in
```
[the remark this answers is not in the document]
jerrymarino: Needs more work for clang - I've outlined the entire solution to index while building V2 here: https://github.com/bazel-ios/rules_ios/blob/master/docs/index_while_building.md#objc-remote-cache-support

.bazelrc
```diff
@@ -20,3 +20,7 @@ build --deleted_packages tests/ios/frameworks/sources-with-prebuilt-binaries
 # Enable dbg compilation mode in this repo, so we can test xcodeproj-built
 # binaries contain debug symbol tables.
 build --compilation_mode=dbg
+
+# This flips index_while_building_v2 on
```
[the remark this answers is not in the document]
jerrymarino: Good point, will change to the design TDLR https://github.com/bazel-ios/rules_ios/blob/master/docs/index_while_building.md#index-while-building-v2-design

.bazelrc
jerrymarino: Yeah we use WMO as default, need to send a patch for the non WMO case. ATM if you set `iwbv2` then you've gotta use WMO until we fix that one

rules/library.bzl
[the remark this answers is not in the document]
jerrymarino: This is probably another change that we'll want for indexstore size: https://github.com/bazelbuild/rules_swift/issues/575 - I can propose that upstream or we can add it into rules_ios

.bazelrc
jerrymarino: @amberdixon fixed here - moves it to conditionally use the non transient output for non WMO: https://github.com/bazelbuild/rules_swift/pull/567/commits/ceaacfc34dba1575e8555db50ddb4009772bf9c8 - I can bump the sha of this.

jerrymarino:
Seems good for now, for a followup we might want to consider moving the `rules_swift` dep into the workspace so that consumers have to opt in to getting our fork of `rules_swift` as opposed to having it loaded in `repositories.bzl`

PR #219, opened 2021-02-26
Pull buildifier with Bazel

jerrymarino:
Replace the brew dep with Bazel and then pull the version we want.


segiddins (approved):

thiagohmcruz (approved):
Thx for fixing this! 🙏

jerrymarino:
You've got it, should be easier to run on an exact shasum now too

PR #222, opened 2021-03-04
Conditionally pass DEBUG flag to LLDB

thiagohmcruz:
Can be considered a continuation of #218, now that LLDB is aware of the `FRAMEWORK_SEARCH_PATHS` some use cases might have to be handled case-by-case.

This PR is an example, an objc interface was hidden behind a `#if DEBUG` statement and debugging a `.swift` file that depends on it was failing to find it.

With this change we'll respect the `Debug` config and pass the respective flag to LLDB. Moving forward once we evolve how different build configurations are set/handled we have to remember to take that into consideration in how to pass flags to LLDB. Added comments in two different places about this to ensure it's being called out explicitly.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -825,6 +825,12 @@ Tags for configuration:
     xcodeproj-ignore-as-target: Add this to a rule declaration so that this rule will not generates a scheme for this target
 """,
     attrs = {
+        # Important! The `Debug` is being used to conditionally pass
+        # flags to LLDB, so if this behaviour ever changes that needs
+        # to be considered otherwise debugging swift files will stop working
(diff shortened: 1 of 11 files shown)


segiddins (approved):
Seems fine as a stopgap, but ideally the xcodeproj aspect should be gathering the list of defines, since there could be more than just `DEBUG`/`NDEBUG` that are being set

thiagohmcruz:
> Seems fine as a stopgap, but ideally the xcodeproj aspect should be gathering the list of defines, since there could be more than just `DEBUG`/`NDEBUG` that are being set

Agree, this will put things at a usable state. Will ensure that the ideal implementation is being tracked!

PR #224, opened 2021-03-05
Filter out DerivedData paths before passing FRAMEWORK_SEARCH_PATHS to LLDB

thiagohmcruz:
Continuation of the debugger fixes.

After generating an Xcode project using the `xcodeproj` rule, if different targets are manually built in Xcode it will create paths under DerivedData that can cause LLDB to fail to load it's ASTs since the correct headers and symlinks are being created under `bazel-out` instead. 

If DerivedData is one of the framework search paths for LLDB it will search there and fail before finding the correct paths under `bazel-out` so this PR is explicitly filtering out those paths only for the `target.swift-extra-clang-flags` lldb flag.

### DIFF
--- a/tools/xcodeproj_shims/installers/lldb-settings.sh
+++ b/tools/xcodeproj_shims/installers/lldb-settings.sh
@@ -35,10 +35,16 @@ fi
 # See original approach to fix this for more details
 # https://github.com/bazel-ios/rules_ios/pull/213
 for fsp in $FRAMEWORK_SEARCH_PATHS; do
-  # Note that these paths will come quoted
-  # and with the Xcode env vars already resolved
-  # so we can just pass them without any modification
-  LLDB_SWIFT_EXTRA_CLANG_FLAGS+=(" -F$fsp")
+  # We don't want LLDB so look for things under DerivedData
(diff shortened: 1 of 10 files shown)


tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -35,10 +35,16 @@ fi
 # See original approach to fix this for more details
 # https://github.com/bazel-ios/rules_ios/pull/213
 for fsp in $FRAMEWORK_SEARCH_PATHS; do
-  # Note that these paths will come quoted
-  # and with the Xcode env vars already resolved
-  # so we can just pass them without any modification
-  LLDB_SWIFT_EXTRA_CLANG_FLAGS+=(" -F$fsp")
+  # We don't want LLDB so look for things under DerivedData
```
justinseanmartin, line 38:
```suggestion
  # We don't want LLDB to look for things under DerivedData
```

justinseanmartin (approved):

amberdixon (approved):

amberdixon:
Spoke to Thiago offline, he's OK with me landing this change.

PR #229, opened 2021-03-10
Point Xcode at LD stub

### LINKED ISSUE #228: [xcodeproj] project generator is building tests with Xcode
Hello! I've noticed that the Xcode project is building some tests with Xcode in my project. Is it expected to build the tests with Xcode or Bazel? cc @thiagohmcruz @amberdixon

jerrymarino:
When building a C++ file, setting LD in Xcode doesn't stub out the
linker, it's a different codepath.

This is a simple way to point at the stub by relying on a public clang
flag `-fuse-ld` Ideally we can replace the full build system without
stubbing out tools e.g. with something like XCBuildKit longer term.

Fixes #228

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -671,7 +671,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_CONFIGS": ctx.attr.configs,
     })
 
-    # Stubbding main executable used by xcode so no actual building happening on Xcode side
+    # Stubbing compiler, linker executables used by xcode so no actual building happening on Xcode side
     proj_settings_base.update({
         "CC": "$BAZEL_STUBS_DIR/clang-stub",
         "CXX": "$CC",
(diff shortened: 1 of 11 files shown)


justinseanmartin (approved):

amberdixon (approved):
thanks Jerry!

thiagohmcruz (approved):

keith:
Seems like we should report this to the apple folks too. Is there a LDCPP maybe?

jerrymarino:
You've got it! I'm glad we found a quick fix for it 🥳 - longer term I'm curious to evaluate using XCHammer's [xcode_project rule](https://github.com/pinterest/xchammer#bazel-build-xcode-projects) as another easy way to fix it.

+1 to a feedback! I'd expect it to use `CXX`, as it runs `clang++` - I couldn't find docs or info about other linker build settings than `LD`

jerrymarino:
It looks like the CI is now flaking on `Build and Test` - any suggestions on how to fix this - if it's just flaky today but the tests work locally will land and file an issue to investigate the CI problem.

justinseanmartin:
> It looks like the CI is now flaking on `Build and Test` - any suggestions on how to fix this - if it's just flaky today but the tests work locally will land and file an issue to investigate the CI problem.

Looks like it passed on rerun. I don't know if there is a way to see the historical results, but without that there isn't enough context to know why it failed the first time.

jerrymarino:
Maybe they can give us elastic search on the CI logs 🤔

keith:
I filed FB9056166 about this

keith:
turns out LDPLUSPLUS exists

PR #231, opened 2021-03-30
The defines field should be populated from the deps compilation_context

amberdixon:
According to https://github.com/bazelbuild/bazel/issues/10674, the old `define` field on the ObjcProvider is now the `defines` field on CompilationContext.

Related to https://github.com/bazel-ios/rules_ios/pull/199/files


segiddins (approved):
Would be good to add a test that covers this, but not a blocker

ob (approved):
Good catch!

PR #237, opened 2021-05-04
Bump to Bazel 4.0

jerrymarino:
This PR bumps rules_ios to Bazel 4.0 and includes misc changes. Many
have been addressed in other PRs.

Note:
it also rebases the rules_swift fork to the latest revision and pushed
tag for it. We hope to eventually upstream the changes for IWBV2, which
would be a followup.


WORKSPACE
```diff
@@ -12,6 +12,14 @@ load(
     "apple_rules_dependencies",
 )
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_python",
+    url = "https://github.com/bazelbuild/rules_python/releases/download/0.2.0/rules_python-0.2.0.tar.gz",
+    sha256 = "778197e26c5fbeb07ac2a2c5ae405b30f6cb7ad1f5510ea6fdac03bded96cc6f",
+)
```
thiagohmcruz, line 21: Should this be added behind a `_maybe` in `rules_ios_dependencies()` [here](https://github.com/bazel-ios/rules_ios/blob/master/rules/repositories.bzl)?

WORKSPACE
```diff
@@ -21,19 +29,21 @@ load(
...
     "@build_bazel_apple_support//lib:repositories.bzl",
     "apple_support_dependencies",
 )
 
 apple_support_dependencies()
 
-load(
-    "@com_google_protobuf//:protobuf_deps.bzl",
-    "protobuf_deps",
-)
```
thiagohmcruz, line 34: Mind clarifying why this is being removed?

rules/library.bzl
```diff
@@ -617,6 +617,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             name = swift_libname,
             module_name = module_name,
             generated_header_name = generated_swift_header_name,
+            generates_header = True,
```
thiagohmcruz, line 620: For other reviewers found why this is necessary in these [release notes](https://github.com/bazelbuild/rules_swift/releases/tag/0.20.0).

thiagohmcruz (commented):
🙏 just a few questions for now, I believe we need more 👀 here before landing.

WORKSPACE
jerrymarino: This is for rules swift and now loaded through the new "extras" function

WORKSPACE
jerrymarino: Yeah that's a good call, I'm not sure we _need_ this exact sha ATM

jerrymarino: @thiagohmcruz actually this is rules_ios's workspace, this is only used for rules_ios. I don't think this is needed anymore.

thiagohmcruz: I see! Well I'm always in favor of deleting unused code haha

segiddins (approved):

PR #239, opened 2021-05-10
Add iOS 12.4 to list of versions for test_host_app

adrianod1as:
So it can matches the version used by a project I'm working on, providing [consistency](https://github.com/bazel-ios/rules_ios/pull/174#issuecomment-745721580) to it.

Please let me know if I have to change anything else so it may works. Since I'm still learning Bazel, I just based my changes on the commits in this file.


thiagohmcruz:
Similar to: https://github.com/bazel-ios/rules_ios/pull/221

If the tests pass it should be safe to land. I'll keep an eye on it. Thx @adrianod1as!

PR #241, opened 2021-05-17
Enable test action associations in schemes

erikkerber:
A convention many projects utilize is to have test
targets run as a test action for a scheme where the build
target is not itself a test target, including the
default project generated in Xcode.

MyFramework <-- framework
MyFrameworkTests <-- bundle.unittests

This hopes to enable adding test targets and environment
variables to schemes created for any build target.

### DIFF
--- a/rules/additional_scheme_info.bzl
+++ b/rules/additional_scheme_info.bzl
@@ -1,0 +1,54 @@
+""" A rule for adding additional information such as test targets and test environments to a generated scheme """
+
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo")
+
+AdditionalSchemeInfo = provider(
+    """
+    Information mapping the scheme for a build target with additional information
+    such as test targets to add to a test action.
+    """,
+    fields = {
+        "build_target": "The build target/scheme to add the additional scheme info to.",
+        "test_action_targets": "A list of test actions to add to the scheme for the scheme_build_target.",
+    },
+)
+
+def _additional_scheme_info_impl(ctx):
+    test_action_targets = [
+        struct(
+            name = test_target[AppleBundleInfo].bundle_name,
+            environment_variables = ctx.attr.test_environment,
+        )
+        for test_target in ctx.attr.test_action_targets
+    ]
+
+    return [
+        AdditionalSchemeInfo(
+            build_target = ctx.attr.build_target[AppleBundleInfo].bundle_name,
+            test_action_targets = test_action_targets,
+        ),
+    ]
+
+additional_scheme_info = rule(
+    implementation = _additional_scheme_info_impl,
+    doc = "Pairs a build target with one or more test targets. Returns list of AdditionalSchemeInfo.",
+    attrs = {
+        "build_target": attr.label(
+            mandatory = True,
+            doc = "The build target to attach the additional scheme info to.",
+            providers = [AppleBundleInfo],
+        ),
+        "test_action_targets": attr.label_list(
+            mandatory = False,
+            doc = "A list of test targets that should be appended to the build_target",
+            allow_empty = True,
+            providers = [AppleBundleInfo],
+        ),
+        "test_environment": attr.string_dict(
+            mandatory = False,
+            allow_empty = True,
+            doc = "Extra environment variables to pass during tests",
+            default = {},
+        ),
+    },
+)
--- a/rules/additional_scheme_info.bzl
+++ b/rules/additional_scheme_info.bzl
@@ -40,14 +40,17 @@ additional_scheme_info = rule(
         ),
         "test_action_targets": attr.label_list(
             mandatory = False,
-            doc = "A list of test targets that should be appended to the build_target",
+            doc = "A list of test targets that should be appended to the build_target.",
             allow_empty = True,
             providers = [AppleBundleInfo],
         ),
         "test_environment": attr.string_dict(
             mandatory = False,
             allow_empty = True,
-            doc = "Extra environment variables to pass during tests",
+            doc = """
+Extra environment variables to pass during tests. This will overwrite 
+any value provided in scheme_existing_envvar_overrides.
+            """,
             default = {},
         ),
     },
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -5,6 +5,7 @@ load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//rules:hmap.bzl", "HeaderMapInfo")
+load("//rules:additional_scheme_info.bzl", "AdditionalSchemeInfo")
 
 def _get_attr_values_for_name(deps, provider, field):
     return [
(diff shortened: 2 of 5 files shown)


rules/additional_scheme_info.bzl
```diff
@@ -0,0 +1,56 @@
...
+    test_action_targets = [
+        struct(
+            name = test_target[AppleBundleInfo].bundle_name,
+            environment_variables = ctx.attr.test_environment,
+        )
+        for test_target in ctx.attr.test_bundle_targets
+    ]
+
+    xcode_scheme = struct(
+        scheme_build_action_target = ctx.attr.build_target[AppleBundleInfo].bundle_name,
```
jschear, line 24: nit: I think it might be a bit simpler for these to be fields on the `AdditionalSchemeInfo` provider itself.

rules/xcodeproj.bzl
```diff
@@ -640,12 +641,34 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
             "run": scheme_action_details,
         }
 
+        scheme_infos = _get_attr_values_for_name(ctx.attr.additional_scheme_infos, AdditionalSchemeInfo, "additional_scheme_info")
```
jschear, line 644: If you go with the suggestion above, this would be:
`scheme_infos = [target[AdditionalSchemeInfo] for target in ctx.attr.additional_scheme_infos]`

rules/additional_scheme_info.bzl
erikkerber: Yup, yup, totally.

And done.

erikkerber:
@amberdixon @justinseanmartin Hello 👋🏼 

No rush on (anyone) looking this over, but mind giving the workflow approval a nudge?

justinseanmartin:
> No rush on (anyone) looking this over, but mind giving the workflow approval a nudge?

Looks like someone else nudged and got green builds. I will ask folks to take a look and give feedback or approvals.

rules/additional_scheme_info.bzl
```diff
@@ -0,0 +1,54 @@
...
+    return [
+        AdditionalSchemeInfo(
+            build_target = ctx.attr.build_target[AppleBundleInfo].bundle_name,
+            test_action_targets = test_action_targets,
+        ),
+    ]
+
+additional_scheme_info = rule(
+    implementation = _additional_scheme_info_impl,
+    doc = "Pairs a build target with one or more test targets. Returns list of AdditionalSchemeInfo.",
```
gyfelton, line 34: Since the name is generic and not only applies to tests, we should make the doc generic too. here is my suggestion:
```suggestion
    doc = "Pairs an existing build target with more info such as test targets. Returns a list of AdditionalSchemeInfo.",
```

I tested the case where `test_action_targets ` (and `test_environment `) are missing as attr and the project generation is fine, which is expected since they are optional fields. In future I do see this as a place to add more attr for other kinds of specification for a build target.

rules/additional_scheme_info.bzl
```diff
@@ -0,0 +1,54 @@
...
+            doc = "The build target to attach the additional scheme info to.",
+            providers = [AppleBundleInfo],
+        ),
+        "test_action_targets": attr.label_list(
+            mandatory = False,
+            doc = "A list of test targets that should be appended to the build_target",
+            allow_empty = True,
+            providers = [AppleBundleInfo],
+        ),
+        "test_environment": attr.string_dict(
```
gyfelton, line 47: In my test this overwrites `scheme_existing_envvar_overrides` just for the build target if there is an existing env var set for the test target(SO i added ` env = {  "ENV1": "original",    },` under `ios_unit_test(    name = "ObjcFrameworkTests",... ` and `scheme_existing_envvar_overrides = {"ENV1": "someoverridevalue"},` under ` xcodeproj(    name = "Single-Static-Framework-Project", ...` I think this is expected, so just add clarification that this overrides everything will be sufficient

gyfelton (approved):
Thanks for the contribution and it works as expected!

jerrymarino:
I was wondering what you thought of automatically doing this as sensible default VS having additional code and API: adding relevant test actions to related top level tests. Given a repo with thousands of tests and apps it might be a lot of work and targets to setup a scheme for each if it's passed into the Xcode project. Here the user will need to create one of these for each transitive target to pass into the Xcodeproj rule.

FWIW this is something that XCHammer auto handles for apps: auto adds test hosts to schemes.  We can probably fork it to add that support for `apple_framework_packagaing`. Otherwise, we could try to add that logic into rules_ios. I don't know how hard it'd be to implement the logic in starlark or if that's possible.

The thing I was thinking is that'd allow us to define scheme metadata as part of `apple_library` macro and attach to targets. 

```
apple_library(.. 
   # define lib
   # define test  
   # define test scheme metatdata
   # define app scheme metatdata
   # define profiling scheme metatdata
) 
```

The upside to this is we can localize this metadata in the lib/framework build file of the framework/app/test in question, as opposed to having to put them all in one place.

rules/additional_scheme_info.bzl
erikkerber: 👍🏼 I had noticed that attribute after I had done this (the goal was test action mapping, but custom env vars per scheme are important for us as well).

I'll add notes to the docs.

erikkerber:
@jerrymarino Agree it leads to quite a bit of boilerplate. I believe we'll ultimately help alleviate that with a macro + conventions that minimize how much we need to write and scale. (Similar to to what you're proposing I suppose but with a project-specific macro). 

Regarding adding metadata to apple_library – I suppose it's a question of how much rules_ios wants to promote separation of Xcode and core rules_ios rules out of the box. I admit I'm about 1 month new into rules_ios so I don't quite have a strong opinion to loosely hold 😀

jerrymarino:
@erikkerber makes sense to me! Also thanks again for sending over the PR 👍 . I was mainly curious about how I'd use this feature for my app which has ton of code. On first look, I assumed it'd be hard for me to use this feature as-is, but I was wondering if you had any suggestions:

Say, I've got the directory tree:

```sh
Company/Language/Product/AppA/BUILD
Company/Language/Product/AppB/BUILD
Company/Language/Product/App..[N]/BUILD
Company/Language/FrameworkA/BUILD
Company/Language/FrameworkB/BUILD
Company/Language/Framework..[N]/BUILD
BUILD # <- all projects are required here
```

I assumed my BUILD file would have to look something like this:

```py
additional_scheme_info(name="SchemeInfoFor_Company_Language_FrameworkA", env: {"var_for_a",..}, arguments = [].., profile_env_vars = []..)
additional_scheme_info(name="SchemeInfoFor_Company_Language_FrameworkB", env: {"var_for_b",..}, arguments = [].., profile_env_vars = []..)
additional_schem_einfo(name="SchemeInfoFor_Company_Language_Framework..[n]", env: {"var_for_..[n]",..}, arguments = [].., profile_env_vars = []..)

xcode_proj(name="some", additional_scheme_info= ["InfoFor_Company_language_frameworkA", ...])
```

jerrymarino (approved):

erikkerber:
Currently our implementation _is_ that brute-force :joy:. A way I envisioned making it a little better:

Rough pseudo-starlark:

**something.bzl**
```
def module(name):
  apple_framework(name = name, srcs = glob(["Sources/**"]))
  unit_test_srcs = glob(["Tests/**])
  if unit_test_srcs:
    ios_unit_tests(name = "{}Tests".format(name), srcs = unit_test_srcs)
    additional_scheme_info(name = "{}Scheme".format(name)
```

**Apples/BUILD.bazel**

```
module()
```

**BUILD.bazel**

```
xcode_proj(name="some", deps = ["//Apples"], additional_scheme_info=["//Apples:ApplesScheme"])
```

To your point there is no way around ultimately itemizing all of the scheme labels in the xcodeproj declaration. I'm very open to the idea that another approach may be a better direction for rules_ios, if this approach isn't super palatable for all users.

gyfelton:
@erikkerber Let me know when u think it's ready to merge since u mentioned about updating notes etc.

erikkerber:
@gyfelton Done!

gyfelton:
Merged, thanks for the contribution!

PR #245, opened 2021-05-28
Allow importing modulemaps from vendored frameworks when building with sandboxing enabled

segiddins:


thii:
FYI if you pass `--define=apple.incompatible.objc_framework_propagate_modulemap=true` to the build then the modulemaps will be included in the inputs.

https://github.com/bazelbuild/rules_apple/blob/9a152a84f5ca9ebaeae7bcb244b729e7d3a0d750/apple/internal/apple_framework_import.bzl#L95-L108

gyfelton:
> FYI if you pass `--define=apple.incompatible.objc_framework_propagate_modulemap=true` to the build then the modulemaps will be included in the inputs.
> 
> https://github.com/bazelbuild/rules_apple/blob/9a152a84f5ca9ebaeae7bcb244b729e7d3a0d750/apple/internal/apple_framework_import.bzl#L95-L108

Does it mean the patched version this PR uses (https://github.com/bazel-ios/rules_ios/pull/58) is not needed? I am looking back at the patch to make sure I understand the change correctly

gyfelton:
@segiddins can provide more context about how this might fix the errors we saw related to `protocol not found`? (if this is the error you are trying to fix)

segiddins:
I think @thii nailed it. Goal was to include modulemaps in compile inputs for {Swift,Objc}Compile

PR #246, opened 2021-06-02
Bump rules_swift to HEAD

jerrymarino:
Now it works without worker.

- Add the updated flag for IWBV2
- Remove index-import as it's now baked in

- [ ] hold until the PR at rules_swift merges
https://github.com/bazelbuild/rules_swift/pull/567


justinseanmartin (approved):

thiagohmcruz (approved):

jerrymarino:
Also it looks like an old bug in xctestrunner where it'd leave a result bundle and blow up retries

PR #248, opened 2021-06-07
Fix trailing doublequote in swiftmodule installer

erikkerber:
Small thing I noticed when debugging the diagnostics output while noticing swift modules never seemed to be found/copied.

```bash
Start copying swiftmodules at Mon Jun  7 11:47:49 CDT 2021
Found 0 existing SWIFTMODULES
No SWIFTMODULES found
Finish copying swiftmodules at Mon Jun  7 11:47:49 CDT 2021
```

This removes a trailing double-quote from the BEP parsing. 

i.e.
```
 /private/var/tmp/_bazel_ekerber/dd323d2a4ff53820c11ccc0d5aebe791/execroot/myapp/bazel-out/ios-x86_64-min12.2-applebin_ios-ios_x86_64-dbg-ST-fc18f8f88e01/bin/MyApp.swiftmodule"
```

Which would then fail the -f check in **_swiftmodules.sh**

With this change, swift modules are copied:

```
Start copying swiftmodules at Mon Jun  7 11:56:28 CDT 2021
Found 104 existing SWIFTMODULES
```

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/swiftmodules.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/swiftmodules.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 # Copy Bazel build `.swiftmodule` files to `DerivedData`. This is used by Xcode
 # and its indexing.
 echo "Start copying swiftmodules at `date`"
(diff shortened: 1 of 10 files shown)


justinseanmartin (approved):

justinseanmartin:
I think the generated test projects need to be updated also in order to get CI passing.

thiagohmcruz (approved):
As Justin said it looks like CI needs a small adjustment. LGTM!

erikkerber:
Ah I see it now. Thanks 👍🏼

PR #249, opened 2021-06-07
Copy dSYMs for Profile/Instruments support

erikkerber:
Context via iOS-dev-at-scale: https://ios-dev-at-scale.slack.com/archives/CRPRV58UF/p1622743928003000

`xcodeproj/rules_ios` won't copy generated dSYM artifacts to derived data, which are used by Instruments to symbolicate various profiling tools.

This adds a straight-forward find/rsync strategy that will move any dsyms, if they exist, over to Derived Data. For this to have any effect, the user will have to use a build configuration that _generates_ dsyms

i.e.
```
build:xcode_profile --compilation_mode=opt
build:xcode_profile --apple_generate_dsym
```

With this approach, I did not use the same BEP/grep approach used in swiftmodules.sh and indexstores.sh. I found that it created a problem if the app was built outside of Xcode as not all artifacts would appear in the BEP file on all builds – but would love to discuss with somebody if there is something I'm missing.

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/dsyms.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/dsyms.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -euo pipefail
+
+echo "Start copying dSYMs at `date`"
+
+# Typicaly dSYMs would only be copied when profiling (and are required by Instruments).
+# In this case, we aren't sure what confiuration is used for profiling,
+# so just copy them over if they exist.
+
+readonly dsym_input="bazel-bin/$BAZEL_BIN_SUBDIR/${FULL_PRODUCT_NAME}.dSYM"
+readonly dsym_output="$TARGET_BUILD_DIR"
+
+if [[ -z $dsym_input ]]; then
+  echo "No dsyms found, finished at `date`"
+  exit 0
+fi
+
+echo "Found ${#dsym_input[@]} DSYMS"
+
+rsync \
+  --recursive --chmod=u+w --delete -i \
+  "$dsym_input" "$dsym_output"
+
+echo "Finish copying dsyms at `date`"
--- a/tools/xcodeproj_shims/install.sh
+++ b/tools/xcodeproj_shims/install.sh
@@ -106,3 +106,4 @@ fi
 # which XCode will use for indexing. Let's keep those.
 "$BAZEL_INSTALLERS_DIR"/swiftmodules.sh >"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stderr-"$DATE_SUFFIX".log &
 "$BAZEL_INSTALLERS_DIR"/indexstores.sh >"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stderr-"$DATE_SUFFIX".log &
+"$BAZEL_INSTALLERS_DIR"/dsyms.sh >"$BAZEL_DIAGNOSTICS_DIR"/dsyms-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/dsyms-stderr-"$DATE_SUFFIX".log &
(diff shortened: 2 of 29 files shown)


segiddins (approved):

tools/xcodeproj_shims/install.sh
```diff
@@ -106,3 +106,5 @@ fi
 # which XCode will use for indexing. Let's keep those.
 "$BAZEL_INSTALLERS_DIR"/swiftmodules.sh >"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stderr-"$DATE_SUFFIX".log &
 "$BAZEL_INSTALLERS_DIR"/indexstores.sh >"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stderr-"$DATE_SUFFIX".log &
+"$BAZEL_INSTALLERS_DIR"/indexstores.sh >"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stderr-"$DATE_SUFFIX".log &
```
segiddins, line 109: intenionally duplicated?

erikkerber: It was not intentional, and thank you for the eagle eye.

jerrymarino (approved):

tools/xcodeproj_shims/install.sh
```diff
@@ -106,3 +106,4 @@ fi
 # which XCode will use for indexing. Let's keep those.
 "$BAZEL_INSTALLERS_DIR"/swiftmodules.sh >"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/swiftmodules-stderr-"$DATE_SUFFIX".log &
 "$BAZEL_INSTALLERS_DIR"/indexstores.sh >"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/indexstores-stderr-"$DATE_SUFFIX".log &
+"$BAZEL_INSTALLERS_DIR"/dsyms.sh >"$BAZEL_DIAGNOSTICS_DIR"/dsyms-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/dsyms-stderr-"$DATE_SUFFIX".log &
```
jerrymarino, line 109: Any tradeoffs on making this a background job? I was wondering if that can introduce any race conditions in consumers of this: for instance if any program requires these to be present immediately proceeding a build.

I was also curious about how the ordering should stack up against these other jobs. I was wondering if its possible for the profiler to fail due to contention against a long index-import operation?

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/dsyms.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -euo pipefail
+
+echo "Start copying dSYMs at `date`"
+
+# Typicaly dSYMs would only be copied when profiling (and are required by Instruments).
+# In this case, we aren't sure what confiuration is used for profiling,
+# so just copy them over if they exist.
+
+readonly dsym_input="bazel-bin/$BAZEL_BIN_SUBDIR/${FULL_PRODUCT_NAME}.dSYM"
```
jerrymarino, line 11: FWIW XCHammer / Tulsi is reading this out of the BEP: https://github.com/bazelbuild/tulsi/blob/1542ca49297d36c79204f08936eb49ec1af4ac02/src/TulsiGenerator/Scripts/bazel_build.py#L1256

I think with rules_ios this is OK if we can assume this path is guaranteed to match 👍

jerrymarino: I'm not 100% sure if this code is handling all the cases that logic is: e.g. if you've got child bundles or if the app is  partitioned into into dylibs . I don't know how this code would be updated to handle that case.

tools/xcodeproj_shims/install.sh
erikkerber: I admit I interpreted this pattern as organizational in nature and not something that promoted async.

The profiler would only fail if the profiling task started before the dsyms were copied, which is technically possible but in practice nearly impossible (launching Instruments would take orders of magnitude longer).

If you think it's more appropriate to do this work directly in **install.sh**, synchronously, I'm totally able to move it up.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/dsyms.sh
erikkerber: You're right that this doesn't handle dylibs. I might ping you offline if that's alright.

PR #251, opened 2021-06-08
Pattern for managing swift compilation improvements

jerrymarino:
This PR is a pattern to add required features to swift compilation like
AST dumping ( https://github.com/bazelbuild/rules_swift/pull/623 )  and
Index While Building V2 (
https://github.com/bazelbuild/rules_swift/pull/567 ) for rules_ios in a
maintainable way without ditching rules_swift altogether. It does this
in way where we can create our own toolchain and add our own features in
it, and integrate them or replace upstream features quite nicely.

It achieves this by providing:
1. a `swift_library` rule to override attributes
2. a "toolchain" composed of existing toolchains but allows full overriding
   of all attributes and most logic.

The benefit is that we can add add new action configs ( configurators ),
and run improved worker code without maintaining a fork. Longer term we
can propose features and gradually get them upstreamed.

For now, it bases a swift_library rule on the rules_swift rule, which is
Apache compatible. Longer term, we may consider calling
swift_common.compile directly and get rid of the extra rule in our
stack. I suspect if rules_swift uses register_toolchain idiom we may
also be able to get rid of this.


jerrymarino:
For now this PR is based on the rules_swift update branch which contained the changes for index while building V2. We'll need another PR to migrate index while building V2 into the toolchain and appending it's `swift_toolchain_config.action_config`. I suspect we can take a similar approach for https://github.com/bazelbuild/rules_swift/pull/623

jerrymarino:
Taking a different approach for this

PR #255, opened 2021-06-11
Pass HEADER_SEARCH_PATHS to LLDB

thiagohmcruz:
Can be seen as a continuation of https://github.com/bazel-ios/rules_ios/pull/213

Realized that we also need to include the `HEADER_SEARCH_PATHS` otherwise LLDB won't find the `.hmap` files created in `rules_ios`.

One use case that is broken because of this is debugging a `.swift` file in a test host app that depends on a bridging header (`SWIFT_OBJC_BRIDGING_HEADER ` Xcode setting) and that is `#import`-ing some other headers. Those headers will be added to an `.hmap` file and linked in Xcode's `HEADER_SEARCH_PATHS` setting but as of now LLDB is not aware of it.

### DIFF
--- a/tools/xcodeproj_shims/installers/lldb-settings.sh
+++ b/tools/xcodeproj_shims/installers/lldb-settings.sh
@@ -46,6 +46,23 @@ for fsp in $FRAMEWORK_SEARCH_PATHS; do
     LLDB_SWIFT_EXTRA_CLANG_FLAGS+=(" -F$fsp")
   fi
 done
+# So LLDB is aware of all headers that are not visible
+# from $FRAMEWORK_SEARCH_PATHS above
+#
+# Fox example, test hosts that depend on objc -> swift
+# bridging headers that only exists in .hmap files
+for hdr in $HEADER_SEARCH_PATHS; do
+  # We don't want LLDB so look for things under DerivedData
+  # given the fact that all relevant paths are being created
+  # under 'bazel-out'
+  if [[ "$hdr" != *"$BUILT_PRODUCTS_DIR"* ]]
+  then
+    # Note that these paths will come quoted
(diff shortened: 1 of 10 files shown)


tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -46,6 +46,23 @@ for fsp in $FRAMEWORK_SEARCH_PATHS; do
     LLDB_SWIFT_EXTRA_CLANG_FLAGS+=(" -F$fsp")
   fi
 done
+# So LLDB is aware of all headers that are not visible
+# from $FRAMEWORK_SEARCH_PATHS above
+#
+# Fox example, test hosts that depend on objc -> swift
+# bridging headers that only exists in .hmap files
+for hdr in $HEADER_SEARCH_PATHS; do
```
gyfelton, line 54: To me it feel more like we just need to pump all build settings to lldb side... which makes sense because we try to mimic how xcode pass build settings within XCode? 
I feel that we need a readme (at the top of this file?) to indicate what is carried over under what conditions (such as don't look for things under DerivedData), and maybe a list of are the know build settings that might needed to be copied over?

gyfelton (approved):

tools/xcodeproj_shims/installers/lldb-settings.sh
thiagohmcruz: The problem with putting too much info here is that this runs on every build. Had performance problems in the past here specially when building really big apps.

Agree that the end goal is what you described, but I think we should do it iteratively as we see the need for new settings and keep an eye on performance as we go

I plan do add better tests for this soon to catch regressions, will make sure there's some documentation also being added 👍

PR #257, opened 2021-06-14
Allow test_host_app app rules to receive custom attributes

thiagohmcruz:
So a consumer of `rules_ios` can optionally initialize the default test host apps with additional attributes.

One example of a use case enabled by this PR is wanting to `bazel run` a [xcodeproj](https://github.com/bazel-ios/rules_ios/blob/master/rules/xcodeproj.bzl#L857) target with the `--ios_multi_cpus=arm64` flag set so the generated Xcode project has `arm64` paths under `bazel-out` for `HEADER_SEARCH_PATHS` and `FRAMEWORK_SEARCH_PATHS` [allowing one to debug .swift sources in mixed projects on device](https://github.com/bazel-ios/rules_ios/blob/master/tools/xcodeproj_shims/installers/lldb-settings.sh#L37-L55).

At this time the above is not possible because `--ios_multi_cpus=arm64` requires a `provisioning_profile` attribute in `ios_application` so it's not possible to use the current default test host apps for that since it would throw an error re missing `provisioning_profile` attribute. 

The changes in this PR allow a consumer to load the `generate_test_host_apps` symbol and pass in additional attributes without losing all the existing configuration in `rules_ios`.

### DIFF
--- a/rules/test_host_app/test_host_app.bzl
+++ b/rules/test_host_app/test_host_app.bzl
@@ -1,0 +1,40 @@
+"""
+Generates test host ios_application targets for each supported version
+
+A consumer of `rules_ios` can optionally load the `generate_test_host_apps` symbol
+and pass in additional attributes to be set in all `ios_application` rules
+"""
+
+load("//rules:app.bzl", "ios_application")
+
+def generate_test_host_apps(**kwargs):
+    versions = [
+        "9.3",
+        "10.0",
+        "11.0",
+        "12.0",
+        "12.2",
+        "12.4",
+        "13.0",
+        "13.2",
+        "13.6",
+        "14.0",
+        "14.1",
+        "14.2",
+        "14.3",
+    ]
+    for version in versions:
+        ios_application(
+            name = "iOS-{}-AppHost".format(version),
+            srcs = ["main.m"],
+            bundle_id = "com.example.ios-app-host-{}".format(version),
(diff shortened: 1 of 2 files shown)


rules/test_host_app/test_host_app.bzl
```diff
@@ -0,0 +1,40 @@
...
+Generates test host ios_application targets for each supported version
+
+A consumer of `rules_ios` can optionally load the `generate_test_host_apps` symbol
+and pass in additional attributes to be set in all `ios_application` rules
+"""
+
+load("//rules:app.bzl", "ios_application")
+
+def generate_test_host_apps(**kwargs):
+    versions = [
```
gyfelton, line 11: wondering this is the opportunity to allow passing of versions list... (and we have a default list of course) I'd imagine the need for doing so if rules_ios doers not keep up

gyfelton (approved):

rules/test_host_app/test_host_app.bzl
thiagohmcruz: Thought about this too, also the ability to set the other `ios_application` attributes (`entitlements`, `families`, etc) but I think we should address use cases as they feel necessary? Otherwise it's easy to make this more generic than necessary IMO

I think we can keep this as is and come back here in the future to make it more customizable only if there's a use case for it

Issue #225, opened 2021-03-05
Test / mitigate costly module recompilation: inheritance of OS version


jerrymarino:
Add the ability to inherit the OS version from the top level application down the tree. By default it'd be ideal to set the OS version for a single build. There are a few reasons this is useful.

1. performance. As modules are cached with respect to OS version, I get O(N OS versions) module rebuilds. see lower
2. deprecations. If I'm staticlaly linking some code into my app I'd like the compiler to alert me if it's using a deprecated or missing API

It seems like pt 1. might be marginal, however, when scaling to thousands of includes and many OS versions it is costly. Consider the build

```
swift_lib a - deps [b], os version 10
swift_lib b - deps [c], os version 9
swift_lib c - deps [d], os version 8
swift_lib d -deps [e] , os version 7
swift_lib e, os version 6
```

At each step of the build we need to reparse and recompile modules for all transitive imports because the modules for the previous step aren't reused. For specific use cases this is a non trivial cost, depending on how expensive it is to compile a PCM

jerrymarino:
This issue is related to https://github.com/bazelbuild/rules_swift/issues/575 in a way - but solves it from another, and more ideal angle: pinning the OS version downwards, and minimizing the OS version in the repo to the shipping products to reduce variants.

jerrymarino:
In my preliminary benchmark run, this cut ~5-6 mins off clean build time from the baseline, but the larger impact seems to be on specific on incremental build use cases that hit module recompilation at every library along the way.

jerrymarino:
Here is a t-shirt sized summary of the impact:

- Changing a file in the top level module with an empty module cache: 435 seconds
- Changing a file with a hot module cache: 50 seconds.

The same issue occurs with objc.

This is most severe in an incremental build which changes a lot modules. Generally while the build progresses we populate 95% of code into iOS 11 module cache. When it gets to the top level library ( has  iOS 12 as a version ) it has to re-build the module cache for the entire tree. I'm going to work on a proposal to get in version pinning in. 

Consider https://github.com/bazel-ios/rules_ios/pull/258 a WIP for this, but we need to address libraries that don't go through that transition right now.

PR #258, opened 2021-06-15
[HOLD] Allow build level pinning of OS version

jerrymarino:
This fixes module re-use across top level modules by setting a build
level OS version for each platform.

Note: I'm evaluating re-working the transition and making the
default CLI arg take precedent

Relates to #225.


jerrymarino:
Taking a different approach for this

PR #261, opened 2021-06-23
Wrap ios_application dependencies in framework

jerrymarino:
Currently when using this rule set with an application organized around
`apple_framework_packaging` the rules_ios transition is missing from the
app sources. Source files picked up in the `ios_application`a globs are
sent through `apple_library`. This defaults to wrapping the code
apple_framework_packaging to apply the transition.

Alternatives considered: it may be ideal to just remove the "library"
feature from this macro to simplify the implementation and eliminate the
need to partition out kwargs. That may be ideal longer term but will
require changes in other repos.

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -1,5 +1,6 @@
 load("@build_bazel_rules_apple//apple:ios.bzl", rules_apple_ios_application = "ios_application")
 load("//rules:library.bzl", "apple_library")
+load("//rules:framework.bzl", "apple_framework")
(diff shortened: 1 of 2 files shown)


segiddins:
If this is fixing an issue, can you add a test that covers that issue?

jerrymarino:
Thanks for the comment @segiddins , yes there is a bug where the transition applied.

For the test case, I didn't see a test case covering how the transitions are applied to bundles, I'm not going to be tackling this anytime soon.

PR #263, opened 2021-06-25
Add ability to generate a Xcode project with additional LLDB settings

thiagohmcruz:
* So it's easier to add extra LLDB settings without manually giving Xcode path to `.lldbinit` files, use the ones generated by `rules_ios` instead
* Small refactor in `tests/ios/xcodeproj/post_build_check.sh` to not look for search paths if the `.lldbinit` file does not exist (besides adding checks for the feature being introduced in this PR)


tools/xcodeproj_shims/installers/lldb-settings.sh
```diff
@@ -21,6 +21,13 @@ settings set target.sdk-path $SDKROOT
 settings set target.swift-framework-search-paths $FRAMEWORK_SEARCH_PATHS
 END
 
+if [ ! -z ${BAZEL_LLDB_LOG_FILE+x} ]; then
+  if [ -n $BAZEL_LLDB_LOG_FILE ] && [ -n $BAZEL_LLDB_LOG_FILE_TYPE ]; then
+    mkdir -p $(dirname $BAZEL_LLDB_LOG_FILE)
+    echo "log enable -f $BAZEL_LLDB_LOG_FILE lldb $BAZEL_LLDB_LOG_FILE_TYPE" >> $BAZEL_LLDB_INIT_FILE
```
congt, line 27: Will the lldb log still be enabled after we removing these attrs? Is it possible to clean it up?

jerrymarino: How do we turn this off?

rules/xcodeproj.bzl
```diff
@@ -913,6 +915,21 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "additional_bazel_build_options": attr.string_list(default = [], mandatory = False),
         "bazel_execution_log_enabled": attr.bool(default = False, mandatory = False),
         "bazel_profile_enabled": attr.bool(default = False, mandatory = False),
+        "lldb_log_file": attr.string(default = "", mandatory = False, doc = """
```
jerrymarino, line 918: As a user I often have some custom LLDB init code I'd like to run - what do you think about just exposing an API to enable that. In XCHammer generated project this works because I can just add an entry to `$HOME/.lldbinit` - to customize the lldb init for debugging or other features. That might already be possible today in rules_ios.

That seemed beneficial because it didn't change the project generated but still allows debugging: as a user I can just pop a line into the lldbinit!

jerrymarino (commented):
Minor suggestion iniline

tools/xcodeproj_shims/installers/lldb-settings.sh
gyfelton: i thought if user re-generate the project without this attribute, it will be cleaned up?

thiagohmcruz: That's correct, re-generating without this will create a `.lldbinit` file without the `log enable` command. Same for other attributes so that's why I followed this pattern

rules/xcodeproj.bzl
thiagohmcruz: [We create one LLDB configuration file per target](https://github.com/bazel-ios/rules_ios/blob/master/rules/xcodeproj.bzl#L507)

But I guess there's value in making this PR a way to set multiple additional LLDB settings instead of "log enable" only?

If we would follow the path of using `$HOME/.lldbinit` no changes are required in `rules_ios` right? Since Xcode will pick it up. The goal of this PR was to abstract editing scripts and/or LLDB configuration files directly to simplify the user experience.

I think making the changes here generic for any list of LLDB settings instead of "log enable" only is a good middle-ground between:

1. Assuming you know what you're doing, edit `$HOME/.lldbinit` and let Xcode do its thing
2. Let `rules_ios` talk to LLDB by just setting `xcodeproj` attributes for "log enable" only

I'd update this PR to set any list of LLDB settings passed in in all `.lldbinit` files for each target in that case via an `xcodeproj` attribute.

cc @jerrymarino thoughts?

thiagohmcruz:
@gyfelton @jerrymarino this is now more generic instead of handling `log enable` only. Adding the ability to set `additional_lldb_settings` for all targets should enable way more use cases. Let me know your thoughts!

jerrymarino (approved):
Nice @thiagohmcruz  - LGTM

PR #265, opened 2021-06-28
Add ability to specify pre/post actions for build/run/...

gyfelton:
This change gives xcodeproj two more attributes for specifying pre/post actions against build/test etc.
It will run before and after build/test/run and will not spit output to build log, so any additional script ran there should be fast and shall not hang.

In terms of the data structure to pass here, three considerations are given here based on this [discussion](https://github.com/bazelbuild/bazel/blob/e2c91cc330179b638e3b88b2fd5ba4db63e0e1ce/src/main/java/com/google/devtools/build/lib/packages/Type.java#L58-L80) regarding why we should not have a `json style` compliticated data structure as a type for `attr`. One point strikes me is that having a complicated data structure as a type for `attr` makes `query` filtering `attr` a very difficult job. 

Comparison of three types here:
1. `string_list`
Pros: straight forward, it can be a list of scripts but tied to `build` and `pre-action`
Cons:, to extend the feature, in worse case we will need 6+ attr for `pre/post * build/test/run`, right now we can do `pre/post * build` which is 2 attr.

2. `string_dict`
Pros: can be very specific, such as `{type: build, name: 'script name', script: 'some script'}
Cons: Each attribute can only host one script for either pre/post. Note that we cannot put `list` in the dict here or we run into a `json style` data structure (DS).

3. `string_list_dict` (base is a dict and each key is string and value is a list of string)
Pros: most complicated DS we can get, so it gives the ability to specify at least the type and its scripts:
`{'build': [script1, script2, script3...], "run":[script1, script2, script3]}` for pre/post (2 attr). 
Cons: limited to key value means that we cannot specify name for script unless the string array itself has some special format within it. We cannot specify other fields such as [settingsTarget](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#execution-action) . 

Given above, I am choosing `string_list_dict` here and follow the sample DS listed above. We are using this already in codebase and it is at least better than `additional_prebuild_script` where it only accepts a single script string

 
TODO 
- [ ] Fix specs

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -683,12 +683,10 @@ def _add_pre_post_actions(target_name, scheme, key, actions):
         scheme: target scheme to update for
         key: one of preActions or postActions
         actions: original attribute passed in from ctx.attr.additional_pre_actions or ctx.attr.additional_post_actions
-    Returns:
-        nothing, scheme will be updated a the end
     """
-    supproted_keys = ["preActions", "postActions"]
-    if key not in supproted_keys:
-        fail("Key must be one of %s" % supproted_keys)
+    supported_keys = ["preActions", "postActions"]
+    if key not in supported_keys:
+        fail("Key must be one of %s" % supported_keys)
     for action_type in actions:
         if action_type not in scheme:
             break
@@ -939,8 +937,16 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "additional_files": attr.label_list(allow_files = True, allow_empty = True, default = [], mandatory = False),
         "additional_prebuild_script": attr.string(default = "", mandatory = False),  # Note this script will run BEFORE Bazel build script
         "additional_bazel_build_options": attr.string_list(default = [], mandatory = False),
-        "additional_pre_actions": attr.string_list_dict(default = {}, mandatory = False, doc = "Configure a list of pre-actions for build/run etc. For each entry the key is one of build/test/run and value is a list of scripts"),
-        "additional_post_actions": attr.string_list_dict(default = {}, mandatory = False, doc = "Configure a list of post-actions, see additional_pre_actions for details"),
+        "additional_pre_actions": attr.string_list_dict(default = {}, mandatory = False, doc = """
+Configure a list of pre-actions for build/run/test in each scheme generated. 
+For each entry the key is one of build/test/run and value is a list of scripts.
+And it will not surface any error or output through build log.
+        """),
+        "additional_post_actions": attr.string_list_dict(default = {}, mandatory = False, doc = """
+Configure a list of post-actions for build/run/test in each scheme generated. 
+For each entry the key is one of build/test/run and value is a list of scripts.
+And it will not surface any error or output through build log.
+        """),
         "bazel_execution_log_enabled": attr.bool(default = False, mandatory = False),
         "bazel_profile_enabled": attr.bool(default = False, mandatory = False),
     },
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -669,7 +669,34 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
                 ],
             }
 
+        pre_actions_attr = ctx.attr.additional_pre_actions
+        _add_pre_post_actions(xcodeproj_schemes_by_name[target_name], "preActions", pre_actions_attr)
+        post_actions_attr = ctx.attr.additional_post_actions
+        _add_pre_post_actions(xcodeproj_schemes_by_name[target_name], "postActions", post_actions_attr)
     return (xcodeproj_targets_by_name, xcodeproj_schemes_by_name)
+
+def _add_pre_post_actions(scheme, key, actions):
+    """Helper method to populate passed in scheme with pre/post actions. Note their output won't show up in build log.
+
+    Args:
+        scheme: target scheme to update for
+        key: one of preActions or postActions
+        actions: original attribute passed in from ctx.attr.additional_pre_actions or ctx.attr.additional_post_actions
+    Returns:
+        nothing, scheme will be updated a the end
+    """
+    supproted_keys = ["preActions", "postActions"]
+    if key not in supproted_keys:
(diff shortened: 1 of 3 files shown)


rules/xcodeproj.bzl
```diff
@@ -911,6 +938,8 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "additional_files": attr.label_list(allow_files = True, allow_empty = True, default = [], mandatory = False),
         "additional_prebuild_script": attr.string(default = "", mandatory = False),  # Note this script will run BEFORE Bazel build script
         "additional_bazel_build_options": attr.string_list(default = [], mandatory = False),
+        "additional_pre_actions": attr.string_list_dict(default = {}, mandatory = False, doc = "Configure a list of pre-actions for build/run etc. For each entry the key is one of build/test/run and value is a list of scripts"),
+        "additional_post_actions": attr.string_list_dict(default = {}, mandatory = False, doc = "Configure a list of post-actions, see additional_pre_actions for details"),
```
thiagohmcruz, line 942: Nit: I think it's fine to duplicate some of the doc here instead of pointing folks to the pre-actions's 

But up to you!

rules/xcodeproj.bzl
```diff
@@ -669,8 +669,35 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
...
     return (xcodeproj_targets_by_name, xcodeproj_schemes_by_name)
 
+def _add_pre_post_actions(scheme, key, actions):
+    """Helper method to populate passed in scheme with pre/post actions. Note their output won't show up in build log.
+
+    Args:
+        scheme: target scheme to update for
+        key: one of preActions or postActions
+        actions: original attribute passed in from ctx.attr.additional_pre_actions or ctx.attr.additional_post_actions
+    Returns:
```
thiagohmcruz, line 685: Maybe simply remove the "Returns:" field then from the docs?

rules/xcodeproj.bzl
```diff
@@ -669,8 +669,35 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
...
+        nothing, scheme will be updated a the end
+    """
+    supproted_keys = ["preActions", "postActions"]
```
thiagohmcruz, line 688: "supported"

thiagohmcruz (approved):
Is it intentional to not add the ability to set the "title" for each action in the first iteration?

tinder-maxwellelliott:
> Is it intentional to not add the ability to set the "title" for each action in the first iteration?

Would like to see this as well if possible

rules/xcodeproj.bzl
```diff
@@ -911,6 +938,8 @@ https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
         "additional_files": attr.label_list(allow_files = True, allow_empty = True, default = [], mandatory = False),
         "additional_prebuild_script": attr.string(default = "", mandatory = False),  # Note this script will run BEFORE Bazel build script
```
tinder-maxwellelliott, line 939: We should remove this no?

gyfelton: mine is different, this is added at 
Screen Shot 2021-06-28 at 4 06 16 PM

mine is at:
Screen Shot 2021-06-28 at 4 06 36 PM

gyfelton:
> Is it intentional to not add the ability to set the "title" for each action in the first iteration?

@tinder-maxwellelliott @thiagohmcruz With current data structure it is not possible to add title.

As discussed in the PR description, we can only specify a list of scripts for a "target" (build/test, run or profile etc.)

I don't think title is needed because the script itself is often descriptive enough and the title does not really add any value.

What do you folks think?

rules/xcodeproj.bzl
tinder-maxwellelliott: > mine is different, this is added at
> Screen Shot 2021-06-28 at 4 06 16 PM
> 
> mine is at:
> Screen Shot 2021-06-28 at 4 06 36 PM

So I think that we should potentially remove that build phase, I know I added it but I am unsure if that is the right "Bazel" way to do this. I believe developers should really only be using pre/post actions in a Bazel Build, offering this build phase gives the wrong impression about what is going on with this project.

This is just my opinion, trying to reduce visible levers for developers when things go wrong

tinder-maxwellelliott:
> > Is it intentional to not add the ability to set the "title" for each action in the first iteration?
> 
> @tinder-maxwellelliott @thiagohmcruz With current data structure it is not possible to add title.
> 
> As discussed in the PR description, we can only specify a list of scripts for a "target" (build/test, run or profile etc.)
> 
> I don't think title is needed because the script itself is often descriptive enough and the title does not really add any value.
> 
> What do you folks think?

The title is low priority, I think higher priority is making the pre/post the only api we offer to do some unit of work, i.e. removing  `Additional prebuild script`

thiagohmcruz:
> > > Is it intentional to not add the ability to set the "title" for each action in the first iteration?
> 
> > 
> 
> > @tinder-maxwellelliott @thiagohmcruz With current data structure it is not possible to add title.
> 
> > 
> 
> > As discussed in the PR description, we can only specify a list of scripts for a "target" (build/test, run or profile etc.)
> 
> > 
> 
> > I don't think title is needed because the script itself is often descriptive enough and the title does not really add any value.
> 
> > 
> 
> > What do you folks think?
> 
> 
> 
> The title is low priority, I think higher priority is making the pre/post the only api we offer to do some unit of work, i.e. removing  `Additional prebuild script`

@gyfelton +1 to that, my bad I missed the explanation for the title and agree that it's low priority

rules/xcodeproj.bzl
tinder-maxwellelliott: Sounds good. I was trying to understand if others were still using this code path. 

On another note, how are you handling surfacing errors in Xcode from these scripts? I know how to accomplish this in build phases but not in scheme pre/post actions

gyfelton: so right now the only way I can think of is to write to a log to surface errors. Therefore I would be very cautious about when to use it and use pre_build script phase whenever possible.

jerrymarino (approved):

PR #267, opened 2021-06-29
minimum_deployment_os_version / `-target` stopgap

jerrymarino:
It adds the ability to set `minimum_deployment_os_version` and pulls in
the latest sha of `rules_apple`.

It stopgaps target specific evaluation xcconfig issues. When I tried to
build an app with target specifc xcconfigs with settings like
`SWIFT_PLATFORM_TARGET_PREFIX`, I end up getting invalid `-target` flags
in compiler invocations which were incomplete. `rules_ios`'s xcconfig
evaluator is not a full implementation of xcconfig and conflicts with
Bazel's settings here. It doesn't currently evaluate on a per architecture
basis but aligning the 2 may be a longer term exercise.

This PR is related to #225 : making it a bit easier to align with what Xcode is doing

### DIFF
--- a/data/xcspecs.bzl
+++ b/data/xcspecs.bzl
@@ -27,6 +27,26 @@ SETTINGS = {
         "Name": "Apple Clang",
         "OptionConditionFlavors": ["arch", "sdk"],
         "Options": {
+            "CLANG_TARGET_TRIPLE_ARCHS": {
+                "CommandLineArgs": [
+                    "-target",
+                    "$(value)-$(LLVM_TARGET_TRIPLE_VENDOR)-$(LLVM_TARGET_TRIPLE_OS_VERSION)$(LLVM_TARGET_TRIPLE_SUFFIX)",
+                ],
+                "Condition": "com_apple_compilers_llvm_clang_1_0__CLANG_TARGET_TRIPLE_ARCHS__Condition",
+                "DefaultValue": "com_apple_compilers_llvm_clang_1_0__CLANG_TARGET_TRIPLE_ARCHS__DefaultValue",
+                "Type": "StringList",
+            },
+            "CLANG_TARGET_TRIPLE_VARIANTS": {
+                "CommandLineFlag": "-target-variant",
+                "Condition": "com_apple_compilers_llvm_clang_1_0__CLANG_TARGET_TRIPLE_VARIANTS__Condition",
+                "ConditionFlavors": ["arch"],
+                "Type": "StringList",
+            },
+            "arch": {
+                "CommandLineFlag": "-arch",
+                "Condition": "com_apple_compilers_llvm_clang_1_0__arch__Condition",
+                "Type": "String",
+            },
             "CLANG_TOOLCHAIN_FLAGS": {"CommandLineArgs": ["$(value)"], "Type": "StringList"},
             "diagnostic_message_length": {
                 "CommandLinePrefixFlag": "-fmessage-length=",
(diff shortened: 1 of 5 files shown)


data/xcspecs.bzl
```diff
@@ -27,26 +27,6 @@ SETTINGS = {
         "Name": "Apple Clang",
         "OptionConditionFlavors": ["arch", "sdk"],
         "Options": {
-            "CLANG_TARGET_TRIPLE_ARCHS": {
```
gyfelton, line 30: is this removal temporary?
also wondering how other downstream projects will be affected, consider adding @tinder-maxwellelliott as a reviewer for example

rules/app.bzl
```diff
@@ -21,6 +21,7 @@ _IOS_APPLICATION_KWARGS = [
     "strings",
     "alternate_icons",
     "settings_bundle",
+    "minimum_deployment_os_version",
```
gyfelton, line 24: maybe put it aside with minimum_os_version and give a comment to explain their difference?

gyfelton (approved):

data/xcspecs.bzl
jerrymarino: We can add some of these flags back if we end up implementing a comprehensive xcconifg evaluator, which would be more involved and perhaps ideal.

It's unlikely a user would set these, mainly if they did today it'd be broken - I suspect it's a safe change.

rules/app.bzl
jerrymarino: The docs for `ios_application` are at rules_apple https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_application  - Perhaps we can just remove this wrapper longer term.

data/xcspecs.bzl
congt: Instead of manually modifying this file, should we update the xcspec_extractor.py and run `bazel run data_generators:extract_xcspecs`? We can also add some comments there.

justinseanmartin: Agree with Cong's comment. Directly editing codegen seems undesirable.

How much work would it be to handle these fields, rather than sidestepping these values in the xcconfig? That seems like it would be directionally better.

Will this leave the values in the xcconfig applied in the generated project (even if Bazel build settings are fine)? Will that lead to unexpected IDE behavior?

thiagohmcruz: +1 

This will also make this and the other generated file out of sync: https://github.com/bazel-ios/rules_ios/blob/b439882984e3e756f8a54a589b6aa3e7071bb502/data/xcspec_evals.bzl#L3669

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -59,6 +59,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             config_vars = ctx.var,
             device_families = ["iphone", "ipad"],
             explicit_minimum_os = None,
+            explicit_minimum_deployment_os = None,
```
thiagohmcruz, line 62: I assume this is required due to the `rules_apple` bump right?

justinseanmartin: It is [a new feature that landed 2 weeks ago](https://github.com/bazelbuild/rules_apple/pull/1099) that we want to leverage and expose to rules_ios callers, but does effectively exactly what Jerry was looking to do to decouple minimum OS version from from the version used to compile the code.

data/xcspecs.bzl
jerrymarino: Yep, it's a good point these are generated - we can probably pop them off inside of the method that processes these!

PR #268, opened 2021-06-29
[xcodeproj] Support for building/debugging extensions

### LINKED ISSUE #266: App Extension support

erikkerber:
This adds support for building and debugging app extension in Xcode using the xcodeproj rule. Simply enough, this is achieved by:

1. Adding a `com.apple.product-type.app-extension` case to the installer script
2. Collecting the extension names for any bundle in the aspect, which is used to create an app/extension build "pair" in the scheme for the extension.

By including the app and extension build targets in the scheme, the install script will be run once for the app and the extension, and both will be copied to the built products directory.

 Achieving *2* can be done in several ways, and I chose an approach that doesn't require any sort of "look ahead/two pass" technique, but am open to alternatives to my approach. In short – if when generating the schemes for an application target that application also bundles extensions, go ahead and create/modify the scheme _for each extension_, adding itself (the application target) to the build actions. 

The "create if non-exist, mutate if exist" strategy required me to break out scheme creating into a convenience method `_create_scheme_for_target`, so it can be called once for the current target itself, and once for the extension scheme (as needed) per pass.

~Not strictly documented – but the extensions should be passed as a dep to `xcodeproj` in this case. It's possible to _not_ require that with a few modifications, but am interested in others opinions.~ Extensions are picked up with https://github.com/bazel-ios/rules_ios/pull/268/commits/f0911b38f9db59c6df06712da10c8975edb3423e

Resolves #266

### DIFF
--- a/tools/xcodeproj_shims/install.sh
+++ b/tools/xcodeproj_shims/install.sh
@@ -45,6 +45,14 @@ com.apple.product-type.application)
         "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/$TARGET_NAME.zip"
     )
     ;;
+com.apple.product-type.app-extension)
+    input_options=(
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${FULL_PRODUCT_NAME}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${TARGET_NAME}_archive-root/$TARGET_NAME${WRAPPER_SUFFIX:-}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/${FULL_PRODUCT_NAME}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/$TARGET_NAME.zip"
+    )
+    ;;
 *)
     echo "Error: Installing ${TARGET_NAME} of type ${PRODUCT_TYPE} is unsupported" >&2
     exit 1
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -133,6 +133,12 @@ def _xcodeproj_aspect_impl(target, ctx):
             if test_host_target:
                 test_host_appname = test_host_target[_TargetInfo].direct_targets[0].name
 
+        # Collect any extension names associated with this bundle. Schemes are then generated that build both.
+        application_extension_names = []
+        application_extensions = getattr(ctx.rule.attr, "extensions", [])
+        if application_extensions:
+            application_extension_names = [extension[AppleBundleInfo].bundle_name for extension in application_extensions]
+
         framework_includes = depset([], transitive = _get_attr_values_for_name(deps, _SrcsInfo, "framework_includes"))
         info = struct(
             name = bundle_info.bundle_name,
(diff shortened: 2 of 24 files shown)


tools/xcodeproj_shims/install.sh
```diff
@@ -45,6 +45,14 @@ com.apple.product-type.application)
         "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/$TARGET_NAME.zip"
     )
     ;;
+com.apple.product-type.app-extension)
+    input_options=(
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${FULL_PRODUCT_NAME}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${TARGET_NAME}_archive-root/$TARGET_NAME${WRAPPER_SUFFIX:-}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/${FULL_PRODUCT_NAME}"
+        "bazel-bin/$BAZEL_BIN_SUBDIR/${BAZEL_BUILD_TARGET_LABEL#*:}.runfiles/${BAZEL_BUILD_TARGET_WORKSPACE}/${BAZEL_BIN_SUBDIR}/$TARGET_NAME.zip"
```
erikkerber, line 53: In practice I don't know when/if these are relevant locations and were mirrored from the `application` case above, and just put these in place discussion purposes in this PR

jerrymarino:
@erikkerber thanks for adding this I like that you didn't add a wrapper for the extension rule as well. Overall it seems fine, is there any way that you'd be able to add a sample case for this under `tests/ios/app`?

tools/xcodeproj_shims/install.sh
jerrymarino: If this works for your use case it seems like a fine addition to merge in, however it'd be ideal to have a sample/ integration test so we can ensure your use case is working in the future

The other thing I as wondering, is that you mentioned building the app as a dependency of this: if we go that approach, is there a reason to run Bazel 2 times for the extension if that extension inside of the app? If it's already built can't we just point Xcode at the app directory?

The last question I had is around the experience for extensions in general. If I have an app with `10` extensions and I'm iterating on do I need to constantly re-build, re-bundle, and re-link all of them? Does Xcode allow you to only build 1 app at a time or all 10?

jerrymarino (approved):

tools/xcodeproj_shims/install.sh
erikkerber: If the approach seems good, I'll go ahead and make a sample integration and test to show it working.

Re: Running Bazel twice – Though the 2nd "build" will obviously be pretty quick, I'm trying to think of how this would work to skip having to do that. Either:

1. Still build 2 Xcode build targets so each build target could manage its installation, but somehow script in "skipping" the Bazel build itself.
2. Build just 1 Xcode build target, but somehow infer all of the additional built products it needs to install in derived data

Since the install script relies heavily on Xcode build vars and individual target settings for the specific built product for that build target (i.e. `$TARGET_NAME`, `$FULL_PRODUCT_NAME`), I'm not sure what the options are here except `find *.app *.appex` in the bazelbin.

I wouldn't be surprised if there was a clever way to do this, but probably easier for others 😀

erikkerber: > l. Say I have an app with 10 extensions. If I'm iterating on a single one do I need to constantly re-build, re-bundle, and re-link all of them?

Similar to how you'd do this in classic Xcode, you'd pick one of them (as in the activated scheme for that extension), and that scheme would build both the app target as well as that extension target.

The experience here is similar except this will clumsily run bazel twice, with the 2nd one being a (near) no-op.

erikkerber: For reference, this is what an out-of-the-box app + share extension looks like in Xcode. Xcode automatically adds both build targets to the scheme for the extension itself.

CleanShot 2021-07-02 at 13 37 34

gyfelton:
Hey @erikkerber Should I merge your PR? Sorry we totally missed this, let me know and I will do it first thing!

erikkerber:
Hey @gyfelton! I had an admittedly sudden inability to spend too much extra-curricular time on various things and meant to provide sample cases per @jerrymarino recommendation. 

So your call on merging now or later, I may or may not be able to get to those in short order 😅

gyfelton:
> Hey @gyfelton! I had an admittedly sudden inability to spend too much extra-curricular time on various things and meant to provide sample cases per @jerrymarino recommendation.
> 
> So your call on merging now or later, I may or may not be able to get to those in short order 😅

Thanks for contributing here! I think let's hold it for now as I am going to put up big changes that requires new test caes. That is where I can add extensions etc to red->green test your change :)

PR #275, opened 2021-07-08
Remove pre/post action having references to build target (xcode crush if build target is at large scale)

gyfelton:
since projects usually have a large scale, the env vars an target can have might be a very large list that exceed xcode arg max limit (256 kb) This is resolved in MacOS 11 but for now we should not add the reference to the target so that xcode won't crash.

For use cases where we have to add it back, we can add a new flag for it or consider it when we refactor xcodeproj fule

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -693,6 +693,9 @@ def _add_pre_post_actions(target_name, scheme, key, actions):
         actions_to_add = actions[action_type]
         payload = scheme[action_type]
         list = []
+
+        # Note `settingsTarget` is explicitly omitted because
+        # for a target with lots of env vars we might exceed Xcode arg_max limit
         for action in actions_to_add:
             list.append({"script": action})
         payload[key] = list
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -694,7 +694,7 @@ def _add_pre_post_actions(target_name, scheme, key, actions):
         payload = scheme[action_type]
         list = []
         for action in actions_to_add:
-            list.append({"script": action, "settingsTarget": target_name})
+            list.append({"script": action})
         payload[key] = list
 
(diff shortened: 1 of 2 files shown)


rules/xcodeproj.bzl
```diff
@@ -693,8 +693,11 @@ def _add_pre_post_actions(target_name, scheme, key, actions):
         actions_to_add = actions[action_type]
         payload = scheme[action_type]
         list = []
+
+        # Note `settingsTarget` is explicitly omitted because
+        # for a target with lots of env vars we might exceed Xcode arg_max limit
         for action in actions_to_add:
-            list.append({"script": action, "settingsTarget": target_name})
+            list.append({"script": action})
```
thiagohmcruz, line 700: Can you clarify the impact of this change? This is a recent change so I doubt many people are using it, just wanted to understand how this limits the script.

thiagohmcruz (approved):

gyfelton:
If the removed key-value pair is set, then all env variables such as SRCROOT, PROJECT_DIR will be present for the script to consume

acecilia:
@gyfelton I have a usecase where I run a script from the preAction. The script is located next to the xcode project. Previously I did:
```
cd "$SRCROOT" && ./myScript.sh
```

But now, since `SRCROOT` is not accessible I am not sure what env variable to use in order to find the script 🤔

tinder-maxwellelliott:
Does WORKSPACE_PATH work for you?

acecilia:
@maxwellE I think it will 🙌  Could you point out where is it configured? 🙏

maxwellE:
It’s simply provided by Bazel itself, I believe. We use that in our scripts and it works great

Regards,

Maxwell Elliott

________________________________

acecilia:
Cool thanks! 🙌 I was just wondering if it is reliable, thus asking where is it defined

PR #276, opened 2021-07-14
Virtual hermetic frameworks POC

jerrymarino:
Add the ability to virtualize the framework structure. swift and clang
compile as frameworks but the files don't move and we load them via VFS.

This is a working POC - for perf benchmarks only and as a reference of
the work.

Hacked this up to see how it impacts build performance in a big app.
Cut clean build time in half in a big hybrid app from rules_ios
baseline. More interestingly, hot top level compiler invocations are
down nearly .5 orders of magnitude.

### DIFF
--- a/rules/framework/merge_vfs_overlay.swift
+++ b/rules/framework/merge_vfs_overlay.swift
@@ -1,0 +1,94 @@
+
+
+import Foundation
+
+let output = ProcessInfo.processInfo.arguments[1]
+let args = ProcessInfo.processInfo.arguments[2...]
+
+func loadVFS(path: String) -> [String: Any] {
+   do {
+          let data = try Data(contentsOf: URL(fileURLWithPath: path), options: .mappedIfSafe)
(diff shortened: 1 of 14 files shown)


jerrymarino:
Will re-open the PR once I've landed on the final approach, there are still a ton of tradeoffs and design decisions I haven't thought through.

rules/framework/merge_vfs_overlay.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    for _nextRoot in nextRoots {
+        guard let nextRoot = _nextRoot as? [String: Any] else {
+            print("Invalid", _nextRoot)
+            continue
+        }
+
+        let name = nextRoot["name"] as! String
+        if var prev = rootsByName[name] as? [String: Any] {
```
jerrymarino, line 60: this doesn't do anything - it should be commented as this but potentially deleting the entire file

Issue #211, opened 2021-02-17
Evaluate reducing compiler file system reads by reducing search paths


jerrymarino:
When looking at rules_ios, I noticed that there are a significant number of search paths passed into the compiler. It seems to be O(N) transitive dependencies.

If the compiler needs to hit these search paths for each job, there will be a O( M search paths X N imports ) performance hit on I/O: specifically it will do a lot more reads than it should. The classic fix for this, which PodToBUILD uses is to pass a single search path to a headermap which has O(1) lookups. This is just something I noticed when looking at rules_ios and haven't tested it out in depth.

segiddins:
Most everything is a framework search path, which headermaps won't work for -- we'd need a vfs overlay, and either way, this is how Xcode handles search paths for framework headers, and we don't want to deviate in behavior

jerrymarino:
Makes sense to me, I haven't tested using a VFS VS hmap recently but originally the hmap loaded a lot faster into memory and was faster for clang.  This being aside, I noticed a few things when playing around with this locally:

### A basic Xcode project with frameworks

Xcode puts built frameworks into single directory - tools would find the build frameworks at O(1) search cost if it gets the  it the build directory in the search first - it seems to in my testing.

In practice, the tools like `clang` `swift` and `ld` would up finding a matching dep framework in the build directory and won't search further.

Specifically the command lines always will have the first search path
```
-F/Path/toDerivedData/*/Build/Products/Debug-iphonesimulator
```

### Cocoapods default behavior with a framework

Cocoapods version I'm running locally seems to put frameworks in their own directory, so even if xcode were to have the path as above it won't find them. This would implicate the O(N) deps searching as above, unless of course it is using a top level VFS or other mechanism to mitigate the search in `clang` and `swift`.


### Mitigation going forward

I'm interested to see flame graphs and `--profiling` output for any solution to this before and after 📈 

Perhaps a VFS could be a path to get it to the state of O(1) framework locating: vs having to search all the transitive framework paths. I'm interested to see how this can impact the search algorithms in clang and swift, and also the cost of serializing in the json vfs. Where hmap is a binary hash table that's fast to map in.

Also, the module cache can help here. When not pulling from a populated module cache, it might still be expensive and may be worth worth profiling. The header search cost is probably paid down multiple times with per OS modules: https://github.com/bazel-ios/rules_ios/issues/225

PR #277, opened 2021-07-16
Virtual hermetic frameworks

jerrymarino:
Add the ability to virtualize the framework structure. swift and clang
compile as frameworks but the files don't move and we load them via VFS.
The VFS is llvm's in-memory file system declared by JSON. VFS is already
in use but this uses it in a very different way.

This fixes many issues:
1. we don't hit O(n) deps includes - removing thousands of framework
search paths
2. we don't have to "clean" frameworks. This feature added quite a bit
of overhead and isn't necesary since it's declarative
3. imports are fully again. traditionally a build system uses -F on a
build directory. This causes problems with reproducible builds.

It cut clean build time in half in a big hybrid app from rules_ios
baseline. More interestingly, hot top level compiler invocations are
down nearly .5 orders of magnitude.

Remaining tasks:
- address xcodeproj integration
- duplicate test suite to run with this on/off
- address xcframework imports ( right now we're using -F )

Relates to: #211

### DIFF
--- a/rules/features.bzl
+++ b/rules/features.bzl
@@ -1,0 +1,4 @@
+feature_names = struct(
+    # Virtualize means that swift,clang read from llvm's in-memory file system
+    virtualize_frameworks = "apple.virtualize_frameworks",
+)
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -26,6 +26,8 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         **kwargs: Arguments passed to the apple_library and apple_framework_packaging rules as appropriate.
     """
     framework_packaging_kwargs = {arg: kwargs.pop(arg) for arg in _APPLE_FRAMEWORK_PACKAGING_KWARGS if arg in kwargs}
+    kwargs["enable_framework_vfs"] = kwargs.pop("enable_framework_vfs", True)
+
     library = apple_library(name = name, **kwargs)
     apple_framework_packaging(
         name = name,
@@ -128,7 +130,7 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
             continue
         import_vfsoverlays.append(dep[VFSOverlayInfo].vfs_info)
 
-    # Progated interface headers - this must encompass all of them
+    # Propagated interface headers - this must encompass all of them
     propagated_interface_headers = []
 
     # We need to map all the deps here - for both swift headers and others
@@ -156,10 +158,10 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
         module_map = outputs.modulemap,
         private_hdrs = outputs.private_headers,
         has_swift = len(outputs.swiftmodule) > 0,
-        merge_vfsoverlays = import_vfsoverlays + fw_dep_vfsoverlays,
+        merge_vfsoverlays = fw_dep_vfsoverlays + import_vfsoverlays,
     )
 
-    framework_headers = depset(outputs.headers + outputs.module_map + outputs.private_headerss)
+    framework_headers = depset(outputs.headers + outputs.modulemap + outputs.private_headers)
 
     # Includes interface headers here ( handled in cc_info merge for no virtual )
     compilation_context_fields["headers"] = depset(
@@ -449,7 +451,7 @@ def _apple_framework_packaging_impl(ctx):
     # Compute cc_info and swift_info
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
     if virtualize_frameworks:
-        framework_info = _get_virtual_framework_info(ctx, compilation_context_fields)
+        framework_info = _get_virtual_framework_info(ctx, framework_files, compilation_context_fields)
     else:
         # This is empty by default
         framework_info = FrameworkInfo()
@@ -528,7 +530,7 @@ apple_framework_packaging = rule(
             mandatory = False,
             cfg = apple_common.multi_arch_split,
             doc =
-                """Objc or Swift rules to be packed by the framework rule
+                """Additional VFS for the framework to export
 """,
         ),
         "transitive_deps": attr.label_list(
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -6,6 +6,9 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_common")
 load("//rules:library.bzl", "PrivateHeadersInfo", "apple_library")
 load("//rules:transition_support.bzl", "transition_support")
+load("//rules:providers.bzl", "FrameworkInfo")
+load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", "make_vfsoverlay")
+load("//rules:features.bzl", "feature_names")
 
 _APPLE_FRAMEWORK_PACKAGING_KWARGS = [
     "visibility",
@@ -28,6 +31,7 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         name = name,
         framework_name = library.namespace,
         transitive_deps = library.transitive_deps,
+        vfs = library.import_vfsoverlays,
         deps = library.lib_names,
         platforms = library.platforms,
         platform_type = select({
@@ -54,7 +58,7 @@ def _framework_packaging_symlink_headers(ctx, inputs, outputs):
     # two different paths to the same file
     #
     # In that case fails with a msg listing the differences found
-    if len(inputs_by_basename) < len(inputs):
+    if len(inputs_by_basename) != len(inputs):
         inputs_by_basename_paths = [x.path for x in inputs_by_basename.values()]
         inputs_with_duplicated_basename = [x for x in inputs if not x.path in inputs_by_basename_paths]
         if len(inputs_with_duplicated_basename) > 0:
@@ -75,6 +79,11 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
         return []
     if inputs == [None]:
         return []
+
+    virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
+    if virtualize_frameworks:
+        return inputs
+
     if action in ctx.attr.skip_packaging:
         return []
     action_inputs = [manifest] + inputs if manifest else inputs
@@ -112,7 +121,58 @@ def _concat(*args):
             arr += x
     return arr
 
-def _apple_framework_packaging_impl(ctx):
+def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields):
+    import_vfsoverlays = []
+    for dep in ctx.attr.vfs:
+        if not VFSOverlayInfo in dep:
+            continue
+        import_vfsoverlays.append(dep[VFSOverlayInfo].vfs_info)
+
+    # Progated interface headers - this must encompass all of them
+    propagated_interface_headers = []
+
+    # We need to map all the deps here - for both swift headers and others
+    fw_dep_vfsoverlays = []
+    for dep in ctx.attr.transitive_deps + ctx.attr.deps:
+        if not FrameworkInfo in dep:
+            continue
+        framework_info = dep[FrameworkInfo]
+        fw_dep_vfsoverlays.extend(framework_info.vfsoverlay_infos)
+        propagated_interface_headers.append(framework_info.framework_headers)
+
+        # Collect generated headers. consider exposing all required generated
+        # headers in respective providers: -Swift, modulemap, -umbrella.h
+        if not CcInfo in dep:
+            continue
+        for h in dep[CcInfo].compilation_context.headers.to_list():
+            if h.is_source:
+                continue
+            propagated_interface_headers.append(depset([h]))
+
+    outputs = framework_files.outputs
+    vfs = make_vfsoverlay(
+        ctx,
+        hdrs = outputs.headers,
+        module_map = outputs.modulemap,
+        private_hdrs = outputs.private_headers,
+        has_swift = len(outputs.swiftmodule) > 0,
+        merge_vfsoverlays = import_vfsoverlays + fw_dep_vfsoverlays,
+    )
+
+    framework_headers = depset(outputs.headers + outputs.module_map + outputs.private_headerss)
+
+    # Includes interface headers here ( handled in cc_info merge for no virtual )
+    compilation_context_fields["headers"] = depset(
+        direct = outputs.headers + outputs.private_headers + outputs.modulemap,
+        transitive = propagated_interface_headers,
+    )
+
+    return FrameworkInfo(
+        vfsoverlay_infos = [vfs.vfs_info],
+        framework_headers = framework_headers,
+    )
+
+def _get_framework_files(ctx):
     framework_name = ctx.attr.framework_name
     bundle_extension = ctx.attr.bundle_extension
 
@@ -148,8 +208,6 @@ def _apple_framework_packaging_impl(ctx):
 
     # AppleBundleInfo fields
     bundle_id = ctx.attr.bundle_id
-    infoplist = None
-    current_apple_platform = transition_support.current_apple_platform(ctx.fragments.apple, ctx.attr._xcode_config)
 
     # collect files
     for dep in ctx.attr.deps:
@@ -225,7 +283,11 @@ def _apple_framework_packaging_impl(ctx):
             paths.join(framework_dir, "Modules", "module.modulemap"),
         ]
 
-    framework_manifest = ctx.actions.declare_file(framework_dir + ".manifest")
+    virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
+    if not virtualize_frameworks:
+        framework_manifest = ctx.actions.declare_file(framework_dir + ".manifest")
+    else:
+        framework_manifest = None
 
     # Package each part of the framework separately,
     # so inputs that do not depend on compilation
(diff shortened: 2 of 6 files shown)


rules/framework.bzl
```diff
@@ -340,20 +355,87 @@ def _apple_framework_packaging_impl(ctx):
             swift_common.create_module(name = swiftmodule_name, swift = swift_module),
         ]
 
-    # Eventually we need to remove any reference to objc provider
-    # and use CcInfo instead, see this issue for more details: https://github.com/bazelbuild/bazel/issues/10674
-    objc_provider = apple_common.new_objc_provider(**objc_provider_fields)
+    direct_headers = []
+
+    if virtualize_frameworks:
```
gyfelton, line 360: refactor wise: can we do if else and try to put them into sep. functions with clear inputs and outputs? Any reason not to do this? 🙏

rules/framework.bzl
```diff
@@ -340,20 +355,87 @@ def _apple_framework_packaging_impl(ctx):
...
+        import_vfsoverlays = []
+        for dep in ctx.attr.vfs:
+            if not VFSOverlayInfo in dep:
```
gyfelton, line 363: what's the situation where a dep of attr.vfs is not a VFSOverlayInfo?

rules/framework.bzl
```diff
@@ -340,20 +355,87 @@ def _apple_framework_packaging_impl(ctx):
...
+                continue
+            import_vfsoverlays.append(dep[VFSOverlayInfo].vfs_info)
+
+        # Progated interface headers - this must encompass all of them
+        propagated_interface_headers = []
```
gyfelton, line 368: this is an array of depset correct?

rules/framework.bzl
```diff
@@ -340,20 +355,87 @@ def _apple_framework_packaging_impl(ctx):
...
+        _add_to_dict_if_present(compilation_context_fields, "headers", depset(
+            direct = header_out + private_header_out + modulemap_out,
+            transitive = propagated_interface_headers,
+        ))
+        framework_info = FrameworkInfo(
+            vfsoverlay_infos = [vfs_ret.vfs_info],
+            framework_headers = framework_headers,
+        )
+
+    else:
```
gyfelton, line 409: i don't expect an else stmt here, I expect the previous if not virtualized_framework handles everything. 
so this is empty for not_virtualized framework correct?

gyfelton, line 409: i don't expect an else stmt here, I expect the previous if not virtualized_framework handles everything.
so this is empty for not_virtualized framework correct?

rules/framework.bzl
gyfelton, line 368: this is an array of depset correct?

rules/framework.bzl
gyfelton, line 363: what's the situation where a dep of attr.vfs is not a VFSOverlayInfo?

rules/framework.bzl
gyfelton, line 360: refactor wise: can we do if else and try to put them into sep. functions with clear inputs and outputs? Any reason not to do this? 🙏

rules/framework/vfs_overlay.bzl
```diff
@@ -1,40 +1,196 @@
...
+def _vfs_swift_module_contents(ctx, framework_name, root_dir):
+    # Forumlate the framework's swiftmodule - don't have the swiftmodule when
+    # creating with apple_library. Consider removing that codepath to make this
+    # and other situations easier
+    base_path = "/".join(ctx.build_file_path.split("/")[:-1])
+    bin_dir = ctx.bin_dir.path
+    rooted_path = bin_dir + "/" + base_path
+
+    # Note: Swift translates the input framework name to this because - is an
+    # invalid module name
```
gyfelton, line 110:
```suggestion
    # invalid character in module name
```

rules/framework/vfs_overlay.bzl
```diff
@@ -1,40 +1,196 @@
...
+            "contents": headers + private_headers + modules,
+        }]
+    if has_swift:
+        roots += [_vfs_swift_module_contents(ctx, framework_name, FRAMEWORK_SEARCH_PATH)]
+
+    return roots
+
+def _vfs_swift_module_contents(ctx, framework_name, root_dir):
+    # Forumlate the framework's swiftmodule - don't have the swiftmodule when
+    # creating with apple_library. Consider removing that codepath to make this
```
gyfelton, line 103: maybe a link to the codepath will make future fixes easier? also trying to understand the context here

rules/framework/vfs_overlay.bzl
```diff
@@ -1,40 +1,196 @@
...
+    )
+    return [
+        apple_common.new_objc_provider(),
+        cc_info,
+        VFSOverlayInfo(
+            files = depset([vfs_ret.vfsoverlay_file]),
+            vfs_info = vfs_ret.vfs_info,
+        ),
+
+        # Mainly a debugging mechanism
```
gyfelton, line 156: u mean this is optional and it contains the list of generated json files correct?

rules/library.bzl
```diff
@@ -450,10 +524,58 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             tags = _MANUAL,
         )
         vendored_deps.append(import_name)
+        import_headers = native.glob(
+            ["**/*.h"],
+            exclude = ["**/.DS_Store"],
```
gyfelton, line 529: blame MacOS

gyfelton: this is important... otherwise we include deleted header files right... OMG 😱

rules/library.bzl
```diff
@@ -450,10 +524,58 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+            import_module_map = None
+
+        vfs_imported_framework = _find_imported_framework_name(import_headers)
+        vfs_framework_name = vfs_imported_framework if vfs_imported_framework else namespace
+        framework_vfs_overlay(
+            name = import_name + "_vfs",
+            framework_name = vfs_framework_name,
+            modulemap = import_module_map,
+            hdrs = import_headers,
+            tags = _MANUAL,
```
gyfelton, line 571: just to confirm: this tag is for bazel right?

rules/library.bzl
```diff
@@ -450,10 +524,58 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+        )
+        import_vfsoverlays.append(import_name + "_vfs")
+
     for xcframework in kwargs.pop("vendored_xcframeworks", []):
         vendored_deps.append(_xcframework(library_name = name, **xcframework))
+        # TODO: for virtual frameworks we'll need to add this
```
gyfelton, line 577: I wonder if we can group pair on this to finish this TODO to see if we truly understand everything haha

rules/library.bzl
```diff
@@ -492,26 +621,33 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+
+    additional_objc_copts += [
+        "-ivfsoverlay$(execpath :{})".format(framework_vfs_overlay_name),
+        "-F{}".format(VFS_OVERLAY_FRAMEWORK_SEARCH_PATH),
+    ]
+    additional_swift_copts += [
+        "-Xfrontend",
+        "-vfsoverlay$(execpath :{})".format(framework_vfs_overlay_name_swift),
+        "-Xfrontend",
+        "-F{}".format(VFS_OVERLAY_FRAMEWORK_SEARCH_PATH),
```
gyfelton, line 644: 🎉  I need to do a before after experiment on building some modules to see how the compilation params differ based on these

gyfelton (commented):
This is awesome....
Here is my TODOs for this:

rules/library.bzl
```diff
@@ -653,18 +811,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-                destination = "%s.extended.modulemap" % name,
-                source = module_map,
-                swift_header = generated_swift_header_name,
-                module_name = module_name,
-                tags = _MANUAL,
-            )
-            module_map = "%s.extended.modulemap" % name
-
-    if cpp_sources and False:
+    if cpp_sources:
```
congt, line 814: Does your PR need the cc_library? If not, maybe we can enable it in a separate PR to reduce the scope of this PR?

rules/framework.bzl
```diff
@@ -241,54 +254,57 @@ def _apple_framework_packaging_impl(ctx):
...
-        )
     _add_to_dict_if_present(compilation_context_fields, "headers", depset(
         direct = header_out + private_header_out + modulemap_out,
     ))
+
     _add_to_dict_if_present(compilation_context_fields, "defines", depset(
         direct = [],
         transitive = [getattr(dep[CcInfo].compilation_context, "defines") for dep in ctx.attr.deps if CcInfo in dep],
     ))
-    _add_to_dict_if_present(objc_provider_fields, "module_map", depset(
```
congt, line 289: We need to move this to line 287, right?

rules/framework.bzl
```diff
@@ -298,7 +314,6 @@ def _apple_framework_packaging_impl(ctx):
         "multi_arch_linked_archives",
         "multi_arch_linked_binaries",
         "multi_arch_dynamic_libraries",
-        "source",
```
congt, line 301: Do we need source when not virtualize_frameworks?

rules/features.bzl
```diff
@@ -0,0 +1,4 @@
+feature_names = struct(
+    # Virtualize means that swift,clang read from llvm's in-memory file system
+    virtualize_frameworks = "apple.virtualize_frameworks",
```
congt, line 3: If there any reason why we treat it as a feature? Could we simply delete the old way to simplify the code?

rules/framework.bzl
congt: +1

rules/library.bzl
```diff
@@ -419,17 +435,48 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+    # The depend on that rule in the top level one
     for vendored_static_framework in kwargs.pop("vendored_static_frameworks", []):
         import_name = "%s-%s-import" % (name, paths.basename(vendored_static_framework))
         apple_static_framework_import(
             name = import_name,
             framework_imports = native.glob(
                 ["%s/**/*" % vendored_static_framework],
                 exclude = ["**/.DS_Store"],
+                # Note: there are some frameworks imported with private headers
+                #"%s/PrivateHeaders" % vendored_static_framework],
```
congt, line 450: Nit: Could we move the comment up to line 446?

rules/library.bzl
```diff
@@ -442,6 +489,33 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             tags = _MANUAL,
         )
         vendored_deps.append(import_name)
+        import_headers = native.glob(
+            ["%s/**/*.h" % vendored_dynamic_framework],
+            exclude = ["**/.DS_Store"],
```
congt, line 494: Nit: is it necessary to exclude .DS_Store files?

rules/library.bzl
```diff
@@ -450,10 +524,58 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             tags = _MANUAL,
         )
         vendored_deps.append(import_name)
+        import_headers = native.glob(
+            ["**/*.h"],
```
congt, line 528: Should we get the path of `vendored_static_library` in case there are multiple libraries?

rules/library.bzl
```diff
@@ -450,10 +524,58 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+            exclude = ["**/.DS_Store"],
+        )
+        import_module_maps = native.glob(
```
congt, line 531: nit: There should be no modulemap for a vendored_static_library, right?

rules/framework.bzl
jerrymarino: Probably can revert this

rules/framework.bzl
jerrymarino: Yep - this is moved to another line

rules/features.bzl
jerrymarino: Yeah it's easier to iterate on code that's behind a feautre. This and we don't want to disrupt the old code path and may roll this out as an a/b test. Probably longer term we'd want to remove the useage of frameworks altogether

jerrymarino: ( that is, no remove the "frameworks" as a feature )

rules/library.bzl
jerrymarino: Why would this be more suited there?

rules/framework/vfs_overlay.bzl
```diff
@@ -1,40 +1,196 @@
...
-        _vfs_root(root_dir = framework_path + "/PrivateHeaders", files = ctx.files.private_hdrs),
-        _vfs_root(root_dir = framework_path + "/Modules", files = ctx.files.modulemap, override_name = "module.modulemap"),
-    ]
+    roots = _make_root(ctx, framework_name, framework_path, extra_search_paths, module_map, hdrs, private_hdrs, has_swift)
+    vfs_info = _make_vfs_info(roots)
+    if len(merge_vfsoverlays) > 0:
+        vfs_info = _merge_vfs_infos(vfs_info, merge_vfsoverlays)
+        roots = vfs_info.values()
+
+    if output == None:
```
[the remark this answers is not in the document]
jerrymarino: The code returns early to avoid the extra nesting in this suggestion

rules/library.bzl
jerrymarino: The existing code does this so keeping that pattern up

rules/library.bzl
jerrymarino: In a static library, there is no subdirectory for a framework - so it doesn't use a path

rules/library.bzl
jerrymarino: Yep it's quite common to include a module map with a vendored static library

rules/library.bzl
jerrymarino: The compilation parameters are certainly different here - it now passes the VFS directly to the frontend

rules/library.bzl
jerrymarino: Yep will revert this one - I thought this was as a typo on my end originally but it's truly a dead block

rules/library.bzl
jerrymarino: This is one piece that can be tackled as a followup to keep the size smaller likely

rules/library.bzl
jerrymarino: Let me circle back on this - I think the glob is terminated with `.h` so we shouldn't need an exclude! I think the previous code has an unterminated glob so they needed to exclude it

rules/framework/vfs_overlay.bzl
jerrymarino: Yeah maybe we can just remove this now, in an early rev, it had many vfs, so now there's only 1

rules/framework.bzl
jerrymarino: The key reason to keep the same code is so we don't make divergent code paths with this feature on/off - the code paths are very similar besides that one nooping the "clean" action: perhaps can rename a bit

rules/library.bzl
jerrymarino: Yep the tags are for Bazel - by default they created a VFS with this tag. I'm not 100% sure if we can make an attribute of tags with a default, but it'd be nice if that was possible, or just make a macro

rules/framework.bzl
jerrymarino: You'll find this in the main callsite of `framework_packing` - there's a few edges around imported frameworks and we pass this in ATM

rules/framework.bzl
jerrymarino: Yep, you'll find where it's created downwind of this line https://github.com/bazel-ios/rules_ios/pull/277/files/2c7b4ba65c40c684f4f8e7efbed35c61e84ca582#diff-b5ab983bf150dfa86b7e361224e1c9b0b9856b0b57b5893dbc088c6ca6d8ba52R397

rules/framework.bzl
jerrymarino: Yeah this conditional is for the non `virtual_frameworks` path - we provide an empty one for now to thread in the `FrameworkInfo` provider

rules/library.bzl
jerrymarino: Edit: good catch - the glob is terminated with `.h` - so we can remove the exclusion clause 👍

jerrymarino:
Thanks again for all the comments on the PR - making another pass to cleanup all of your suggestions  🙏 🥳 

@gyfelton regarding the tests, we might be able to come up with some sort of test cases, my plan is to run the existing examples with the feature on - we might need to spin up more .github actions. The feature _shouldn't_ impact any other projects if they don't enable the feature - which is another benefit of making it a feature!

rules/library.bzl
congt: I thought the comment explained why you glob `["%s/**/*" % vendored_static_framework]` and generally prefer writing comments before the code.

congt (approved):

thiagohmcruz (approved):

rules/library.bzl
jerrymarino: Yep I think this suggestion makes a lot of sense - I'll move the comment up a bit!

rules/framework/vfs_overlay.bzl
jerrymarino: Here it's referring to when this is leveraged by the `apple_library` macro - https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl

rules/framework.bzl
jerrymarino: @congt  and @gyfelton I've taken the liberty to detangle some of the original code as we discussed offline, will also make it easier to remove the old way if we decide to do so!

rules/library.bzl
jerrymarino: @gyfelton Sent an invite offline - I've got a ton of stuff going on this week so it'd be real helpful if you want to jump on this 🤝

PR #278, opened 2021-07-21
Fix conflict with umbrella headers

jerrymarino:
The line to determine umbrella headers will cause issues in some
scenarios. This PR reverts to the orginal, albeit more nuanced logic.
Additionally correct the comment which would lead us to believe that
this is OK to disregard.

Will work on getting this explified in rules_ios as a followup.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -591,9 +591,11 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     framework_vfs_overlay_name = name + "_vfs"
     framework_vfs_overlay_name_swift = swift_libname + "_vfs"
 
-    # TODO: remove framework if set
+    # TODO: remove under certian circumstances when framework if set
     # Needs to happen before headermaps are made, so the generated umbrella header gets added to those headermaps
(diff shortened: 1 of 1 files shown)


rules/library.bzl
```diff
@@ -591,9 +591,11 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     framework_vfs_overlay_name = name + "_vfs"
     framework_vfs_overlay_name_swift = swift_libname + "_vfs"
 
-    # TODO: remove framework if set
+    # TODO: remove under certian circumstances when framework if set
```
gyfelton, line 594: when framework if set?

jerrymarino: It's not clear to me what the scenarios are - and the original comment doesn't really indicate when.

jerrymarino: I've reverted it to the original logic as it worked and I'm ignorant of it 
> This PR reverts to the orginal, albeit more nuanced logic.

gyfelton (approved):

PR #282, opened 2021-07-29
Exposes more fields in FrameworkInfo

congt:
Downstream targets can directly read them. A use case is to repackage a static framework into a dynamic framework.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -140,7 +140,8 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
             continue
         framework_info = dep[FrameworkInfo]
         fw_dep_vfsoverlays.extend(framework_info.vfsoverlay_infos)
-        propagated_interface_headers.append(framework_info.framework_headers)
+        framework_headers = depset(framework_info.headers + framework_info.modulemap + framework_info.private_headers)
+        propagated_interface_headers.append(framework_headers)
 
         # Collect generated headers. consider exposing all required generated
         # headers in respective providers: -Swift, modulemap, -umbrella.h
@@ -161,8 +162,6 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
         merge_vfsoverlays = fw_dep_vfsoverlays + import_vfsoverlays,
     )
 
-    framework_headers = depset(outputs.headers + outputs.modulemap + outputs.private_headers)
-
     # Includes interface headers here ( handled in cc_info merge for no virtual )
     compilation_context_fields["headers"] = depset(
         direct = outputs.headers + outputs.private_headers + outputs.modulemap,
@@ -171,7 +170,11 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
 
     return FrameworkInfo(
         vfsoverlay_infos = [vfs.vfs_info],
-        framework_headers = framework_headers,
+        headers = outputs.headers,
+        private_headers = outputs.private_headers,
+        modulemap = outputs.modulemap,
+        swiftmodule = outputs.swiftmodule,
+        swiftdoc = outputs.swiftdoc,
     )
 
 def _get_framework_files(ctx):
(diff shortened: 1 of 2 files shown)


rules/framework.bzl
```diff
@@ -140,7 +140,8 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
             continue
         framework_info = dep[FrameworkInfo]
         fw_dep_vfsoverlays.extend(framework_info.vfsoverlay_infos)
-        propagated_interface_headers.append(framework_info.framework_headers)
+        framework_headers = depset(framework_info.headers + framework_info.modulemap + framework_info.private_headers)
```
thiagohmcruz, line 143: Can you clarify why `framework_info.framework_headers` is not being appended anymore here? Is the new set already covering the same files?

rules/providers.bzl
```diff
@@ -1,6 +1,11 @@
 FrameworkInfo = provider(
     fields = {
         "vfsoverlay_infos": "Merged VFS overlay infos, present when virtual frameworks enabled",
-        "framework_headers": "Headers part of the framework's public interface",
```
thiagohmcruz, line 4: So this is a breaking API change right? Just checking to ensure this is intentional

thiagohmcruz (approved):

rules/providers.bzl
congt: Yes, FrameworkInfo was just introduced to support [vfs framework](https://github.com/bazel-ios/rules_ios/pull/276) by @jerrymarino. I believe it's only used internally.

congt: @jerrymarino PTAL

rules/framework.bzl
congt: Yes, FrameworkInfo now includes headers, modulemap, private_headers in separate fields.

rules/providers.bzl
```diff
@@ -1,6 +1,11 @@
...
+        "binary": "The binary of the framework",
+        "headers": "Headers of the framework's public interface",
+        "private_headers": "Private headers of the framework",
+        "modulemap": "The module map of the framework",
+        "swiftmodule": "The swiftmodule",
+        "swiftdoc": " The Swift doc",
```
gyfelton, line 9: what;s the relationship between swiftmoduel and swiftdoc? right now based on what i have seen it feels like swiftdoc is a child of swiftmodule... along with swiftinterface

rules/providers.bzl
```diff
@@ -1,6 +1,11 @@
 FrameworkInfo = provider(
     fields = {
         "vfsoverlay_infos": "Merged VFS overlay infos, present when virtual frameworks enabled",
-        "framework_headers": "Headers part of the framework's public interface",
+        "binary": "The binary of the framework",
```
jerrymarino, line 4: This one is returned inside of the apple binary already - any reason to replicate here?

rules/providers.bzl
```diff
@@ -1,6 +1,11 @@
...
+        "headers": "Headers of the framework's public interface",
+        "private_headers": "Private headers of the framework",
+        "modulemap": "The module map of the framework",
+        "swiftmodule": "The swiftmodule",
```
jerrymarino, line 8: These ones are in swiftinfo - any reason to put here too?

jerrymarino (commented):
Minor suggestions on duping over material into this provider we propagate  in other providers. ESP on binary and swift, ideally we avoid having the entirety of these providers and behavior implemented and handled 2x 

I was also curious on what the was for levering this - currently, each framework governs it's own VFSOverlay - then that's propagated upwards.

rules/providers.bzl
congt: I want to reuse FrameworkInfo to package a dynamic framework. By adding binary and swiftmodule here, I won't need to pass around AppleBundleInfo or swiftinfo.

jerrymarino: What I'm saying is that today, upstream dependees of this rule are required to retrieve the swiftmodule, and binaries. 
https://github.com/bazel-ios/rules_ios/blob/master/rules/framework.bzl#L497 - so it's already passed around. To access the attributes, just use `framework[SwiftInfo].swiftmodule` or the other existing ones

jerrymarino (approved):
This seems fine, I think it makes sense

rules/providers.bzl
congt: Totally understood. I want to make FrameworkInfo self-contain from the API's perspective. Then downstream targets can only return FrameworkInfo without returning SwiftInfo or AppleBundleInfo.

rules/providers.bzl
congt: I guess that's just how they're packaged in a framework.

thiagohmcruz: They're all binary formats, you can read more about swift modules here: https://github.com/apple/swift/blob/main/docs/Serialization.md

(TL;DR functionally they're like headers that expose public APIs)

Also IIRC .swiftdoc files hold documentation related things like when you use "///" to document a public class and want its description to show up in the UI for who is consuming it (it could be holding more info than that but AFAIK this format is not documented)

Issue #281, opened 2021-07-27
[xcodeproj] IDE issues with test projects / xcodeproj


jerrymarino:
I'm working to add a new way to speed up rules_ios builds that use frameworks, AKA virtual frameworks ( #277 ). On the Xcode side  hitting a few hiccups with test projects today. I've attempted to get a baseline of what is already working in rules_ios but, most of the example projects are broken from the IDE's perspective. This includes debugging, jump to definition, and other features. 

To reproduce:
```
git clone git@github.com:bazel-ios/rules_ios.git
cd rules_ios
bazel run tests/ios/xcodeproj:Test-Imports-App-Project
open tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj
```

I've noticed that rules_ios requires a build to copy some files into derived data. I tried doing a build of the app in the test project `TestImports-App` inside of Xcode as well.

## Debugger
1. Set a breakpoint inside of the main file
2. Run the program
3. _observe that breakpoints don't resolve - see attached screenshot_

## Source editor
1. Open the file `empty.swift`
2. Navigate to the interface of a symbol from "SomeFramework"
2. _observe it navigating to a generated location_

<img width="1860" alt="RulesIOSBrokenJumptoDef" src="https://user-images.githubusercontent.com/1245820/127222104-097ec02d-974e-4718-b7ac-2e8bb6f9c950.png">

gyfelton:
I think this is a known issue: we did not do some treatment that we do in downstream project that enabled debugging in sample projects.  (specifying search paths in build settings etc.)
One thing to try is to manually set things under build setting under xcode such as framework search paths and you might get debugger running again.

jerrymarino:
Yep it looks like there's a few issues in the samples causing both indexing and debugging to not work. 

That said, I think there's a few issues to or next steps to break out:
1. getting the examples working so devs can iterate on features in a small use caes
2. levering the examples to use as integration test cases
3. increasing the coverage on IDE indexing in general
4. ensuring that the incantation of the downstream project isn't busted

For me, I filed this issue less around 2, 3, 4, and more on 1. As a maintainer, It was hard to come up with a baseline for what is and what isn't working then make changes to it. XCHammer has a nice methodology and extensive test suite for this by evaluating compiler invocations in a build. Pt 4 is more difficult but definitely achievable depending on how the test is structured

PR #284, opened 2021-07-30
Use absolute vfs paths via LLVM `VirtualFileSystem`

jerrymarino:
There are issues when using non absolute-path VFS outside of the
compilation directory, even when setting the `working-directory`. We
can't use absolute paths in this file in Bazel for many reasons. To work
around this, use VFS paths relative to the VFS root via compiler
internals.

Internal to swift and clang - LLVM `VirtualFileSystem` object can
serialize paths relative to the absolute path of the overlay. This
requires the paths are relative to the overlay. While deriving the
in-memory tree roots, it pre-pends the prefix of the `vfsoverlay` path
to each of the entries.

In practice, this is required to use rules_ios vfsoverlay with Xcode's
indexer and SourceKit: e.g. in Bazel Xcode project generators like the
one in rules_ios or XCHammer.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -11,11 +11,33 @@ VFSOverlayInfo = provider(
     },
 )
 
+# Computes the "back" segment for a path length
+def _make_relative_prefix(length):
+    dots = "../"
+    prefix = ""
+    for i in range(0, length + 1):
+        prefix += dots
+    return prefix
+
+# The LLVM `VirtualFileSystem` object can serialize paths relative to the
+# absolute path of the overlay. This requires the paths are relative to the
+# overlay. While deriving the in-memory tree roots, it pre-pends the prefix of
+# the `vfsoverlay` path to each of the entries.
+def _get_external_contents(prefix, path_str):
+    return prefix + path_str
+
 # Make roots for a given framework. For now this is done in starlark for speed
 # and incrementality. For imported frameworks, there is additional search paths
 # enabled
 def _make_root(ctx, framework_name, root_dir, extra_search_paths, module_map, hdrs, private_hdrs, has_swift):
     extra_roots = []
+
+    # Compute with the build file path, assuming the build file path is adjacent
+    # and not nested. This method is used in contexts like frameworks which
+    # don't actually output the VFS
+    prefix_len = len((ctx.bin_dir.path + ctx.build_file_path).split("/")) - 1
+    vfs_prefix = _make_relative_prefix(prefix_len)
+
     if extra_search_paths:
         sub_dir = "Headers"
 
@@ -26,13 +48,13 @@ def _make_root(ctx, framework_name, root_dir, extra_search_paths, module_map, hd
         extra_roots = [{
             "type": "file",
             "name": file.path.replace(rooted_path, ""),
-            "external-contents": file.path,
+            "external-contents": _get_external_contents(vfs_prefix, file.path),
         } for file in hdrs]
 
         extra_roots += [{
             "type": "file",
             "name": framework_name + "/" + file.path.replace(rooted_path, ""),
-            "external-contents": file.path,
+            "external-contents": _get_external_contents(vfs_prefix, file.path),
         } for file in hdrs]
 
     modules_contents = []
@@ -40,7 +62,7 @@ def _make_root(ctx, framework_name, root_dir, extra_search_paths, module_map, hd
         modules_contents.append({
             "type": "file",
             "name": "module.modulemap",
-            "external-contents": module_map[0].path,
+            "external-contents": _get_external_contents(vfs_prefix, module_map[0].path),
         })
 
     modules = []
@@ -56,7 +78,7 @@ def _make_root(ctx, framework_name, root_dir, extra_search_paths, module_map, hd
         {
             "type": "file",
             "name": file.basename,
-            "external-contents": file.path,
+            "external-contents": _get_external_contents(vfs_prefix, file.path),
         }
         for file in hdrs
     ])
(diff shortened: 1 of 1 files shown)


rules/framework/vfs_overlay.bzl
```diff
@@ -11,11 +11,33 @@ VFSOverlayInfo = provider(
...
+def _get_external_contents(prefix, path_str):
+    return prefix + path_str
+
 # Make roots for a given framework. For now this is done in starlark for speed
 # and incrementality. For imported frameworks, there is additional search paths
 # enabled
 def _make_root(ctx, framework_name, root_dir, extra_search_paths, module_map, hdrs, private_hdrs, has_swift):
     extra_roots = []
+
+    # Compute with the build file path, assuming the build file path is adjacent
```
jerrymarino, line 35: This is a fair assumption. There is one flaw in this plan: the memoization of this vfs will no longer be valid if the entire VFS must be relative

rules/framework/vfs_overlay.bzl
```diff
@@ -161,35 +180,76 @@ def _merge_vfs_infos(base, vfs_infos):
 
 # Internally the "vfs obj" is represented as a dictionary, which is keyed on
 # the name of the root. This is an opaque value to consumers
-def _make_vfs_info(roots):
+def _make_vfs_info(name, data):
     keys = {}
-    for root in roots:
-        name = root["name"]
-        keys[name] = root
+    keys[name] = data
```
gyfelton, line 185: still makes sense to make it a sep. method now that it just do a keyvalue assignment?

rules/framework/vfs_overlay.bzl
```diff
@@ -113,7 +132,7 @@ def _vfs_swift_module_contents(ctx, framework_name, root_dir):
     return {
         "type": "file",
         "name": root_dir + "/" + name,
-        "external-contents": external_contents,
+        "external-contents": _get_external_contents(vfs_prefix, external_contents),
```
gyfelton, line 135: ❤️ having code into simple helper methods for readability

rules/framework/vfs_overlay.bzl
```diff
@@ -161,35 +180,76 @@ def _merge_vfs_infos(base, vfs_infos):
...
+        hdrs = hdrs,
+        private_hdrs = private_hdrs,
+        has_swift = has_swift,
+    )
+
+    vfs_info = _make_vfs_info(framework_name, data)
     if len(merge_vfsoverlays) > 0:
         vfs_info = _merge_vfs_infos(vfs_info, merge_vfsoverlays)
-        roots = vfs_info.values()
+        roots = _roots_from_datas(vfs_parent, vfs_info.values() + [data])
```
gyfelton, line 244: `vfs_info.values() + [data]` is to merge the two arrays correct?

gyfelton: vfs_info is simply a dict with key being framework_name and value being `data`, now we call `vfs_info.values()` which just returns an array whose only element is `data`, and then we merge it with `[data]?

rules/framework/vfs_overlay.bzl
jerrymarino: The reason we want a factory for the `vfs_info` is that it sets a clear boundary: callers if this method generally aren't concerned about _what_ the details of `vfs_info`. It also improves readability to localizing this logic by related other code. The simple function abstractions here also make it easier to refactor the code

rules/framework/vfs_overlay.bzl
jerrymarino: Yea exactly, you'll also find this in the factory's comment https://github.com/bazel-ios/rules_ios/pull/284/files#diff-450b1330485549712d2fd4e65b2b3fd0a7692fe8f4ec43a0b7b60bbb739313afL163

gyfelton: Guess my confusion is that it is essentially merging two arrays whose elements are both just `data`? Am I correct? If so is this needed?

gyfelton (commented):
Have we tested this against a downstream project? This is potentially breaking change because the overlay_relative is turned on and the path can potentially be very different based on say what the root is. I wonder if we can have a way to compare the generated vfs overlay to 1) always have a clear idea on what the root is 2) to safeguard any regression that changes base of the root and thus breaking things subtly such as debugging ability

rules/framework/vfs_overlay.bzl
jerrymarino: The starlark dictionary function `.values` returns an array, so data is in an array add them together

jerrymarino:
@gyfelton This is a non functional change from the users perspective as far as I have seen. The change happens internally to trigger LLVM creating an absolute path at runtime.  If we compared this Bazel VFS to master, it's quite different but produces the same search. The new one is now compat with Xcode. I can put a followup to dupe the test suite over to VFS.

This PR is for Xcode issues. I'm curious to know how we've been developing and catching regressions on IDE functionality for the rule in general. On master I didn't find tests for indexing and or assertions for IDE functionality. https://github.com/bazel-ios/rules_ios/issues/281 tracks that examples are broken but the tests are green.

gyfelton (approved):

gyfelton:
@jerrymarino regarding how we test debugger side, it's pretty manual that we have manual test scripts to ensure downstream projects have debugger working. We def. did not pay attention to ensure rules_ios examples have debugger working as no CI job can stop the regression there at this point.

jerrymarino:
@gyfelton makes sense, I suspect getting the tests sorted is larger undertaking and potentially another thread. We could maybe automate LLDB attachment with the python or C++ API. I was thinking of this in other contexts to measure symbolication / attachment times too, which might be a useful overall. At least now we've got the other  issue to noodle on this separately

PR #286, opened 2021-07-31
[xcodeproj] Adds `rules_ios` virtual framework args

jerrymarino:
This PR adds the virtulized framework aruguments for rules_ios into the
xcode proj generator.

The generator needs to behave differently with `virtualize_frameworks`.
It disables the current behavior like copying swift modules and setting
header search paths for LLDB and the indexer.

Note that the PR extends the existing approach rules_ios' took to adding
hmap args. It adds vfs args ( like the other args ) with assumptions
about how rules_ios rules work.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -5,7 +5,9 @@ load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//rules:hmap.bzl", "HeaderMapInfo")
+load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", VFS_OVERLAY_FRAMEWORK_SEARCH_PATH = "FRAMEWORK_SEARCH_PATH")
 load("//rules:additional_scheme_info.bzl", "AdditionalSchemeInfo")
+load("//rules:features.bzl", "feature_names")
 
 def _get_attr_values_for_name(deps, provider, field):
     return [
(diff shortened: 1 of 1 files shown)


rules/xcodeproj.bzl
```diff
@@ -5,7 +5,9 @@ load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//rules:hmap.bzl", "HeaderMapInfo")
+load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", VFS_OVERLAY_FRAMEWORK_SEARCH_PATH = "FRAMEWORK_SEARCH_PATH")
```
gyfelton, line 8: never seen this syntax before ` VFS_OVERLAY_FRAMEWORK_SEARCH_PATH = "FRAMEWORK_SEARCH_PATH"` is this just to declare a constant? 🙏

rules/xcodeproj.bzl
```diff
@@ -67,6 +69,108 @@ def _xcodeproj_aspect_collect_hmap_paths(deps, target, ctx):
                 hmap_paths.append(relative_path)
     return hmap_paths
 
+def _make_swift_vfs_args(vfs_path):
+    return [
+        "-Xfrontend",
+        "-vfsoverlay$BAZEL_WORKSPACE_ROOT/{}".format(vfs_path),
+        "-Xfrontend",
+        "-F{}".format(VFS_OVERLAY_FRAMEWORK_SEARCH_PATH),
```
gyfelton, line 77: to confirm: this is essentailly ` "-F{}".format("FRAMEWORK_SEARCH_PATH"),`?

gyfelton (commented):
How feasible it is to have the existing tests to have this feature on so we can what the xcodeproj looks like?

jerrymarino:
@gyfelton yep, today we've got tests which prevent regressions. When the feature is off these tests are certainly green. I'm thinking to dupe over test suite for now to VFS to atleast cover builds, I can submit a followup PR for that one

rules/xcodeproj.bzl
jerrymarino: Yeah this is a starlarkism  to import a symbol under a different name, it just makes for better semantic usage from file to file. I can dig up the docs on the `load` for this if you'd want to read more about it

Kind of like pythons `import x from y as f` notion

gyfelton (approved):
Yay to populating VFS  related flags in xcodeproj side!

thiagohmcruz (approved):

PR #288, opened 2021-08-06
[virtual-frameworks] Fix non-modular framework umbrella error

jerrymarino:
In order to correctly identify some imports as a framework import, clang
needs the entire subtree under an imported framework. The previous path
triggered an edge case in some vendor frameworks.

This builds out the VFS subtrees for a given set of paths in this situation,
and then adds them to the VFS.

This manifested in one very odd permutations of a vendor framework
when using umbrella headers the headers. I suspect it's more
generalizable than the previous logic overall.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -158,13 +158,13 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
     # can build this out if-empty as a followup
     if not extra_search_paths:
         headers_contents.extend([
-        {
-            "type": "file",
-            "name": file.basename,
-            "external-contents": _get_external_contents(vfs_prefix, file.path),
-        }
-        for file in hdrs
-    ])
+            {
+                "type": "file",
+                "name": file.basename,
+                "external-contents": _get_external_contents(vfs_prefix, file.path),
+            }
+            for file in hdrs
+        ])
 
     headers = []
     if len(headers_contents):
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -174,22 +174,23 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
             "contents": headers_contents,
         }]
 
-    private_headers_contents = {
-        "name": "PrivateHeaders",
-        "type": "directory",
-        "contents": [
+    if not extra_search_paths:
+        private_headers_contents.extend([
             {
                 "type": "file",
                 "name": file.basename,
                 "external-contents": _get_external_contents(vfs_prefix, file.path),
             }
             for file in private_hdrs
-        ],
-    }
+        ])
 
     private_headers = []
-    if len(private_hdrs):
-        private_headers = [private_headers_contents]
+    if len(private_headers_contents):
+        private_headers = [{
+            "name": "PrivateHeaders",
+            "type": "directory",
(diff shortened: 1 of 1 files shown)


rules/framework/vfs_overlay.bzl
```diff
@@ -30,30 +30,83 @@ def _get_external_contents(prefix, path_str):
...
+def _vfs_json_lookup_contents(contents, name):
+    for c in contents:
+        if name == c["name"]:
+            return c
+    return None
+
+# Builds out the VFS subtrees for a given set of paths. This is useful to
+# construct an in-memory rep of a tree on the file system
+def _build_subtrees(paths, vfs_prefix):
+
```
gyfelton, line 43:
```suggestion
```

gyfelton (approved):
I am so glad that we can fix this! The new helper method totally makes sense and is indeed more generalized. Since this is about building a tree structure represented by json, having unit tests as a way to give examples will be the best but we can tackle that later.

thiagohmcruz (approved):
Thx for adding detailed comments here ;)

rules/framework/vfs_overlay.bzl
```diff
@@ -30,30 +30,83 @@ def _get_external_contents(prefix, path_str):
...
+        # current pointer to the current subdirs while walking the path
+        curr_subdirs = subdirs
+
+        # Loop the _framework_ path and add each dir to the current tree.
+        # Assume the last bit is a file then add it as a file  This loop is
+        # quite short ( e.g. O(1) since we loop the VFS path ( e.g. X.h ) and
+        # not the file path.
+        idx = 0
+        for part in parts:
```
jerrymarino, line 60: We may want to take a pass on the "Big O" of this code. See comment for when it len(parts) > 1. Figuring out the issue and getting right VFS produced in the code was the majority of the work here and we may do a pass on the entire VFS construction algorithm runtime now that it's correct.

jerrymarino: This is now fixed on latest.

rules/framework/vfs_overlay.bzl
```diff
@@ -36,31 +36,82 @@ def _find_top_swiftmodule_file(swiftmodules):
...
+        # Assume the last bit is a file then add it as a file
+        for part in parts:
+            if idx == len(parts) - 1:
+                ext_c = _get_external_contents(vfs_prefix, path_info.path)
+                curr_subdirs.dict[part] = -1
+                curr_subdirs.json["contents"].append({"name": part, "type": "file", "external-contents": ext_c})
+                break
+
+            # Lookup a value for the current subdirs, otherwise append
+            next_subdirs = curr_subdirs.dict.get(part, None)
```
congt, line 64: I might misunderstand the logic, but looks like curr_subdirs may be None here if len(parts) > 1.

curr_subdirs is initialized as subdirs, which is struct(dict = {}, json = subdirs_json).
next_subdirs will be None if len(parts) > 1.
Then curr_subdirs will be set None on line 72.

jerrymarino: `curr_subdirs` isn't set to non-None value for what I could tell. If you trace the execution, you'll likely find this invariant to be held true. Mainly, `next_subdirs` is always set to a non-None value. If you have an example which invalidates the invariant I'd be curious to see.

congt: You're right, somehow I missed the `not`

congt (approved):

PR #280, opened 2021-07-26
Support dynamic framework/xcframework in Virtualized file system

gyfelton:
Goal is to have VFS info correctly surface for static/dynamic frameworks, static library and xcframework in this PR.

### Changes made
1. Introduce a new attr for vfs_overlay rule to pass in content of swiftmodule. Right now the asusmption is that .swiftmodule is really a dir that contains .swiftinterface and .swiftdoc. This fixes normal static/dynamic framework not expressing swiftmodules in VFS 
2. In current _xcframework macro, added one more `select` which encapsulates the vfs overlay info that is based on the platform settings passed in. 

### How to test
Will add tests in future as we want to ensure VFS works for downstream project first as a standalone feature

TODO 
- [x] Rename swiftmodule -> swiftmodule_contents 
- [X] Handle the case when .swiftmodule is just a binary?  -> We now assume .swiftmodule binary is under .swiftmodule dir

PR #294, opened 2021-08-18
Reduce VFS assumptions about swiftmodules

jerrymarino:
The previous code relied on assumptions about swiftmodules which may be
invalid, depending on the input.

Now:
- It puts the name of the parent in the framework if it's a
  swiftmodule-directory. This handles user provided directory
  swiftmodules. The code now checks if its a `.swiftmodule` because
  previously, it could take _any_ swiftmodule and give it the name of the
  parent: e.g. the build directory.

- It takes the name of the swiftmodule and puts it into the root as
  opposed to computing it.

It can fall back to the code path with assumptions to improve useability
of the rule. There would need to be further refactoring remove this:
Ideally remove this "apple libary only" codepath in the future.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -30,6 +30,12 @@ def _get_external_contents(prefix, path_str):
 def _get_vfs_parent(ctx):
     return (ctx.bin_dir.path + "/" + ctx.build_file_path)
 
+def _find_top_swiftmodule_file(swiftmodules):
+    for file in swiftmodules:
+        if file.basename.endswith(".swiftmodule"):
+            return file
+    return None
+
 # Make roots for a given framework. For now this is done in starlark for speed
 # and incrementality. For imported frameworks, there is additional search paths
 # enabled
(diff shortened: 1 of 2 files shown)


rules/framework/vfs_overlay.bzl
```diff
@@ -130,12 +147,31 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
...
 
-def _vfs_swift_module_contents(bin_dir_path, build_file_path, vfs_prefix, framework_name, root_dir):
+def _provided_vfs_swift_module_contents(swiftmodules, vfs_prefix, root_dir):
+    swiftmodule_file = _find_top_swiftmodule_file(swiftmodules)
+
+    # Note: here we need to import the "top" swiftmodule where the compiler can
+    # find it. Followup if it's possible and gainful to present this inside of
+    # a framework. User provided directory swiftmodules are presented under
+    # frameworks above.
+    if swiftmodule_file == None or swiftmodule_file.is_source:
```
gyfelton, line 166: :TIL is_source

rules/framework/vfs_overlay.bzl
```diff
@@ -64,20 +70,31 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
...
+        # user provided source swiftmodules nested in a directory.
+        #
+        # Note: This code has some edges: perhaps for imported/source
+        # frameworks it should mirror the users's structure directly. See the
+        # related PR to import that source.
+        if any_swiftmodule_file.is_source:
+            # Handle a glob of files inside of a .swiftmodule e.g. for xcframework
+            parent_dir_base_name = any_swiftmodule_file.dirname.split("/").pop()
+            swiftmodule_file = _find_top_swiftmodule_file(swiftmodules)
+            if not swiftmodule_file and parent_dir_base_name.endswith(".swiftmodule"):
```
gyfelton, line 85: so if this is not true, aka we have a .swiftmodule file as a source file not dir, and parent dir is not ending with .swiftmodule, we don't append the content to module_content.. 
on the other hand (code below) if we have a list of swiftmodules, it does the other side of logic... (return None if...)
so these two blocks of code actually have some links between each other?

rules/framework/vfs_overlay.bzl
```diff
@@ -64,20 +70,31 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
...
-            ],
-        })
+        any_swiftmodule_file = swiftmodules[0]
+
+        # Virtualize a swiftmodules directory inside of the framework only for
+        # user provided source swiftmodules nested in a directory.
+        #
+        # Note: This code has some edges: perhaps for imported/source
+        # frameworks it should mirror the users's structure directly. See the
+        # related PR to import that source.
```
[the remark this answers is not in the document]
jerrymarino: Ah yeah I must have wrote a note on this PR, will remove the comment 👍

rules/framework/vfs_overlay.bzl
jerrymarino: In this PR ( https://github.com/bazel-ios/rules_ios/pull/280 ) it started added the parent build file's directory, so that would put unexpected entires in VFS: e.g. if you had a parent of `Public` the could would add a `Public`  add that in the modules directory in the VFS. 

The code makes it more explicit to handle or other issues

gyfelton: discussed and we think it's possible to merge this with code down below. Will deal with this in another PR but for now all these makes sense

gyfelton (approved):

PR #295, opened 2021-08-24
[virtual-frameworks] fix transitive swiftmodule bug

jerrymarino:
If the swiftmodule isn't propagated upwards, it will cause issues when
re-compiling and re-linking in the case of a remote cache or when the
local cache is primed with existing state.

This an example serial chain of swiftdeps to showcase the situation
```
a->b->c
```

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -458,6 +458,8 @@ def _apple_framework_packaging_impl(ctx):
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
     if virtualize_frameworks:
         framework_info = _get_virtual_framework_info(ctx, framework_files, compilation_context_fields)
+        if outputs.swiftmodule:
+            swift_info_fields["modules"] = _copy_swiftmodule(ctx, framework_files)
     else:
         framework_info = FrameworkInfo(
             headers = outputs.headers,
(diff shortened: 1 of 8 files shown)


gyfelton (approved):
nice find! assuming it does apply to non swift side because we already handle them well via module map?
curious what error u got if this two lines are not in and we build the strict-deps target?

justinseanmartin (approved):

thiagohmcruz (approved):

rules/framework.bzl
```diff
@@ -458,6 +458,8 @@ def _apple_framework_packaging_impl(ctx):
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
     if virtualize_frameworks:
         framework_info = _get_virtual_framework_info(ctx, framework_files, compilation_context_fields)
+        if outputs.swiftmodule:
```
congt, line 461: nit:  Move it out of the if-else clause since we have the exactly same code snippet in the else clause (i.e., line 475-477).

congt (approved):

jerrymarino:
@gyfelton if you change a public interface function in`c` it will fail at the linker - basically proving it didn't recompile the dependees. I don't think we have a way to test this kind of file changing ATM, but its the example I used to repro it locally.

rules/framework.bzl
jerrymarino: Sure, I think we can take a pass at how `swiftinfo` is built out too. I can send this over as a followup

PR #299, opened 2021-09-01
Supports Info.plist in framework

congt:
When distributing a framework to third-party developers, we will want to include a Info.plist file into the framework.
This PR uses rules_apple's resource_actions.merge_root_infoplists to merge info.plists and converts them into binary.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -1,7 +1,10 @@
 """Framework rules"""
 
 load("@build_bazel_rules_apple//apple/internal:apple_product_type.bzl", "apple_product_type")
-load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo")
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "AppleSupportToolchainInfo")
+load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_support")
+load("@build_bazel_rules_apple//apple/internal:resource_actions.bzl", "resource_actions")
+load("@build_bazel_rules_apple//apple/internal:rule_support.bzl", "rule_support")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_common")
 load("//rules:library.bzl", "PrivateHeadersInfo", "apple_library")
@@ -9,6 +12,7 @@ load("//rules:transition_support.bzl", "transition_support")
 load("//rules:providers.bzl", "FrameworkInfo")
 load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", "make_vfsoverlay")
 load("//rules:features.bzl", "feature_names")
+load("//rules:plists.bzl", "info_plists_by_setting")
 
 _APPLE_FRAMEWORK_PACKAGING_KWARGS = [
     "visibility",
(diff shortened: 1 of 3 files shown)


rules/framework.bzl
```diff
@@ -442,6 +464,53 @@ def _get_merged_swift_info(ctx, framework_files):
...
+        executable_name = bundle_name,
+        environment_plist = ctx.file.environment_plist,
+        input_plists = ctx.files.infoplists,
+        launch_storyboard = None,
+        output_plist = output_plist,
+        output_pkginfo = None,
+        platform_prerequisites = platform_support.platform_prerequisites(
+            apple_fragment = ctx.fragments.apple,
+            config_vars = ctx.var,
+            device_families = ["iphone", "ipad"],
```
jerrymarino, line 495: How does this work with macOS?

rules/framework.bzl
```diff
@@ -442,6 +464,53 @@ def _get_merged_swift_info(ctx, framework_files):
...
+            explicit_minimum_os = None,
+            explicit_minimum_deployment_os = None,
```
jerrymarino, line 497: Needed for frameworks?

rules/framework.bzl
```diff
@@ -589,6 +666,9 @@ Valid values are:
             default = "framework",
             doc = "The extension of the bundle, defaults to \"framework\".",
         ),
+        "environment_plist": attr.label(
+            allow_single_file = True,
```
jerrymarino, line 670: Consider linking to the rules_apple doc on this one - I didn't know what it was used for.

rules/framework.bzl
```diff
@@ -28,21 +32,39 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
...
         platforms = library.platforms,
         platform_type = select({
-            "@build_bazel_rules_ios//rules/apple_platform:macos": "macos",
             "@build_bazel_rules_ios//rules/apple_platform:ios": "ios",
+            "@build_bazel_rules_ios//rules/apple_platform:macos": "macos",
             "@build_bazel_rules_ios//rules/apple_platform:tvos": "tvos",
             "@build_bazel_rules_ios//rules/apple_platform:watchos": "watchos",
             "//conditions:default": "",
         }),
+        environment_plist = select({
```
jerrymarino, line 61: This would probably better go as a default on the rule attribute if people need to over-ride it?

jerrymarino (approved):
Minor tweaks inline regarding docs / default value s

tests/ios/frameworks/strict-deps/a/BUILD.bazel
```diff
@@ -3,6 +3,7 @@ load("//rules:framework.bzl", "apple_framework")
 apple_framework(
     name = "a",
     srcs = glob(["*.swift"]),
+    infoplists = ["//rules/test_host_app:Info.plist"],
```
gyfelton, line 6: does this test the new code? Aka if we add this line without all other work, will we still get a green build (just no op?)

gyfelton (approved):
i am def. less exp on this part :)

tests/ios/frameworks/strict-deps/a/BUILD.bazel
congt: It will fail with the error
```
ERROR: /Users/cshi/Development/rules_ios/tests/ios/frameworks/strict-deps/a/BUILD.bazel:3:16: //tests/ios/frameworks/strict-deps/a:a_swift: no such attribute 'infoplists' in 'swift_library' rule
ERROR: /Users/cshi/Development/rules_ios/tests/ios/frameworks/strict-deps/a/BUILD.bazel:3:16: //tests/ios/frameworks/strict-deps/a:a_objc: no such attribute 'infoplists' in 'objc_library' rule
ERROR: error loading package 'tests/ios/frameworks/strict-deps/a': Package 'tests/ios/frameworks/strict-deps/a' contains errors
```

rules/framework.bzl
congt: default doesn't allow `select` method, but we ca use the provided environment_plist from kargs if it exists.

rules/framework.bzl
congt: Good catch. Now use rule_descriptor.allowed_device_families to properly handle it.

rules/framework.bzl
congt: minimum_deployment_os is derived from xcode config. explicit_minimum_deployment_os is to override the default setting. No need to do that for frameworks. We can add an attribute if necessary later.

PR #300, opened 2021-09-02
[virtual-frameworks] Fix issue with Sourcekit clang

jerrymarino:
This fixes 2 issue which manifest under certain clang files. We
ended up with the argument used for the swift frontend here.
Additionally, it was unintentionally excluding some hmaps from the
invocation.

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -87,10 +87,11 @@ def _make_swift_vfs_args(vfs_path):
 def _make_swift_copts(target, deps, ctx):
     hmap_files = []
     vfs_files = []
+    collected_vfs = False
     for dep in deps:
-        if VFSOverlayInfo in dep:
+        if VFSOverlayInfo in dep and not collected_vfs:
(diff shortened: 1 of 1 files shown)


congt (approved):

rules/xcodeproj.bzl
```diff
@@ -87,10 +87,11 @@ def _make_swift_vfs_args(vfs_path):
 def _make_swift_copts(target, deps, ctx):
     hmap_files = []
     vfs_files = []
+    collected_vfs = False
     for dep in deps:
-        if VFSOverlayInfo in dep:
+        if VFSOverlayInfo in dep and not collected_vfs:
```
thiagohmcruz, line 92: So we're assuming that only one `dep` will hold the vfs provider right? Is that always the case or is it possible that more than one `dep` could be holding this provider?

thiagohmcruz (approved):
I have one small question but LGTM!

rules/xcodeproj.bzl
jerrymarino: Yeah exactly, the existing code did that so was going off that 👍

PR #298, opened 2021-08-31
Add support for assemble (.S) files to apple_framework

nkoroste:
`objc_library` rules supports assemble files just fine there is no reason not to pass them directly to it. Added a test to verify it works as intended.

PR #302, opened 2021-09-15
Suppress legacy build system error in Xcode 13

jschear:
In Xcode 13, the warning about the deprecation of the legacy build system was upgraded to an error. Enabling this setting makes it a warning again.

Found the setting name here: https://github.com/Kitware/CMake/commit/71a2664ebb0fd6a8d56eadee058aecc431a40594

### DIFF
--- a/tests/ios/xcodeproj/Test-BuildForDevice-Project.xcodeproj/project.pbxproj
+++ b/tests/ios/xcodeproj/Test-BuildForDevice-Project.xcodeproj/project.pbxproj
@@ -409,7 +409,7 @@
 				CLANG_ENABLE_OBJC_ARC = YES;
 				FRAMEWORK_SEARCH_PATHS = "$(PLATFORM_DIR)/Developer/Library/Frameworks";
 				GCC_PREPROCESSOR_DEFINITIONS = "$(inherited)";
-				HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/app/OnlySources_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/app/OnlySources_private_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
+				HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-dbg-ST-c32106249197/bin/tests/ios/app/OnlySources_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-dbg-ST-c32106249197/bin/tests/ios/app/OnlySources_private_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
 				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
 				MACH_O_TYPE = "$(inherited)";
 				ONLY_ACTIVE_ARCH = YES;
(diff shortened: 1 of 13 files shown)


tests/ios/xcodeproj/Test-BuildForDevice-Project.xcodeproj/project.pbxproj
```diff
@@ -409,7 +409,7 @@
 				CLANG_ENABLE_OBJC_ARC = YES;
 				FRAMEWORK_SEARCH_PATHS = "$(PLATFORM_DIR)/Developer/Library/Frameworks";
 				GCC_PREPROCESSOR_DEFINITIONS = "$(inherited)";
-				HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-dbg-ST-c32106249197/bin/tests/ios/app/OnlySources_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-dbg-ST-c32106249197/bin/tests/ios/app/OnlySources_private_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
+				HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/app/OnlySources_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-7bf874b56ea0/bin/tests/ios/app/OnlySources_private_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
```
jschear, line 412: I regenerated the projects using this command: ` bazel query 'kind(xcodeproj, ...)' | xargs -n 1 bazel run`.

I figured I could get these paths to reference `arm64` instead if I passed ` --ios_multi_cpus=arm64`, but that gives me errors about unspecified provisioning profiles:
```
❯ bazel query 'kind(xcodeproj, ...)' | xargs -n 1 bazel run --ios_multi_cpus=arm64
Loading: 0 packages loaded
Loading: 1 packages loaded
Loading: 1 packages loaded
INFO: Build option --ios_multi_cpus has changed, discarding analysis cache.
ERROR: /Users/jschear/dev/rules_ios/rules/test_host_app/BUILD.bazel:10:24: in ios_application rule //rules/test_host_app:iOS-10.0-AppHost:
        File "/private/var/tmp/_bazel_jschear/7436e6242b3207e4692683d02ddfce6d/external/build_bazel_rules_apple/apple/internal/ios_rules.bzl", line 321, column 41, in _ios_application_impl
                processor_result = processor.process(
        File "/private/var/tmp/_bazel_jschear/7436e6242b3207e4692683d02ddfce6d/external/build_bazel_rules_apple/apple/internal/processor.bzl", line 711, column 36, in _process
                partial_outputs = [partial.call(p) for p in partials]
        File "/private/var/tmp/_bazel_jschear/7436e6242b3207e4692683d02ddfce6d/external/bazel_skylib/lib/partial.bzl", line 38, column 28, in _call
                return partial.function(*function_args, **function_kwargs)
        File "/private/var/tmp/_bazel_jschear/7436e6242b3207e4692683d02ddfce6d/external/build_bazel_rules_apple/apple/internal/partials/provisioning_profile.bzl", line 40, column 13, in _provisioning_profile_partial_impl
                fail(
Error in fail: ERROR: In //rules/test_host_app:iOS-10.0-AppHost:
Building for device, but no provisioning_profile attribute was set.
```

(Also, does the fact that these paths are CPU specific mean that the same generated project can't build for simulator and device?)

jschear: Also saw this happen here: https://github.com/bazel-ios/rules_ios/pull/298#discussion_r700448574

jschear: Ah, I think I see what I need to do here (which also answers my question): regenrate `:Test-BuildForDevice-Project` _only) with `--ios_multi_cpus=arm64`: https://github.com/bazel-ios/rules_ios/blob/f5515c35135d7016f12cf3c960af6d01ba9e8ef1/tests/xcodeproj-tests.sh#L58

thiagohmcruz: You're right, currently the generated proj can't build for both archs. This has been discussed before as is something we definitely should fix but we just couldn't allocate time yet to scope and draft an implementation plan.

What I usually do is to run the shell you just linked here since all invocations have the expected CLI flags there. There's definitely room for improvement here, I added this change just as a way to ensure that at least there is a way to build for device and have the debugger working.

Thx for the contribution btw!

thiagohmcruz (approved):

tests/ios/xcodeproj/Test-BuildForDevice-Project.xcodeproj/project.pbxproj
jschear: For sure! Sounds good 👍 

One more question if you don't mind -- are `HEADER_SEARCH_PATHS` and `FRAMEWORK_SEARCH_PATHS` necessary to get the indexer working (or at least not failing), get lldb working, or for some other reason?

thiagohmcruz: (I don't think that change affects the indexer though, that was LLDB-related only)

gyfelton: for indexing side FRAMEWORK_SEARCH_PATHS is especially important.
The way to test if indexing is working or not to run in terminal: `defaults write com.apple.dt.Xcode IDEIndexShowLog -bool YES` so you got indexing logs on the xcode (at the place where it reports build failure). Then from there check files you want to index by editing it a bit and see whether it got compilation error inside the index log (it will show full clang/swiftc invocation). What we did is to compare the parameters passed to clang/swiftc (especially `-F<path_to_framework>`) between how Xcode compile a source and how Bazel compile. The flags xcode compile with is influenced by build settings that's why we need to configure xcodeproj properly.

tinder-maxwellelliott: I am seeing this working fine for the built target, but how can transitive modules get pulled into the indexer? I am doing some clunky transitive fetching of `*.frameworks` and placing them in derived data, this works but is questionable

Indexing seems to work fine for the target being built, unsure why we cannot do it transitively

PR #303, opened 2021-09-16
[xcodeproj] Add output_path parameter to xcodeproj rule, so xcode project can be generated in a subdirectory

acecilia:
👋 

This PR adds the `output_path` parameter to the `xcodeproj` rule, so the xcode project can be generated in a subdirectory. This is useful when the preference is to keep the top level clean, and instead generate the project under a gitignored folder.

Because the project is generated in a subdirectory, the current testing infrastructure (which relies on all xcode projects being on the same directory) cant be used. As an alternative I added a test comparing `Test-LLDB-Logs-Project.xcodeproj` with `custom_output_path/Test-LLDB-Logs-Project.xcodeproj`


rules/xcodeproj.bzl
```diff
@@ -845,18 +845,28 @@ def _add_pre_post_actions(target_name, scheme, key, actions):
             list.append({"script": action})
         payload[key] = list
 
+def _validate_output_path(output_path):
+    output_path_components = output_path.split("/")
+    prohibited_components = [".", ".."]
+    contains_dot_paths = any([elem in output_path_components for elem in prohibited_components])
+    if contains_dot_paths:
+        fail("The output_path provided cant contain the following path components: <%s>" % prohibited_components)
```
gyfelton, line 853: does it mean the output has to be absolute path? if so i think it's worth calling out in the doc below?

gyfelton (approved):
Thx so much for this important feature! One thing we found on our side is that because xcode has a reference to working-directory being the repo root, for some reason if the project is generated in a sub dir, indexing stops working for us. We want to figure out why but it keeps getting de-prioritized for now.  Please let me know if this does not happen on you side assuming u already got indexing working

jschear:
The project is currently copied sibling to the `BUILD` file the `xcodeproj` target is declared in -- instead of specifying a different output path, would it be simpler to move the target to a different directory? In our case, our main app target is under `/App/`, but we generate a project for it under `/project-gen/`. (We are having some indexing trouble that we're currently investigating, though...).

rules/xcodeproj.bzl
acecilia: Yep. The reason for this is that we cant calculate the relative path to the sources. Same issue as in skylib `paths. relativize`, see: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/paths_doc.md#pathsrelativize

> Because we do not have access to the underlying file system, this implementation differs slightly from Python's os.path.relpath in that it will fail if path is not beneath start

How would you be more explicit?

acecilia:
> Please let me know if this does not happen on you side assuming u already got indexing working

Got it. I will test it out and come back with an answer

> would it be simpler to move the target to a different directory?

Mmm how would you define a target inside a folder, if the full folder is added to `.gitignore`? Isnt `project-gen` in your `.gitignore`

jschear:
> Mmm how would you define a target inside a folder, if the full folder is added to .gitignore? Isnt project-gen in your .gitignore?

Ah -- we `.gitgnore` only the generated project, not the entire directory. Which I guess you could do regardless of where the project ends up. 🤷

rules/xcodeproj.bzl
gyfelton: sry i meant in the doc section `The output path to use when generating the xcode project` maybe update to `The output path to use when generating the xcode project, must be a relative path to workspace root`

acecilia: Done 🚀 

```
        The output path to use when generating the xcode project.
        Must be a relative path beneath the package where the xcodeproj rule is defined
```

acecilia:
> Please let me know if this does not happen on you side assuming u already got indexing working

On this very simple example, indexing appears to work:

Screenshot 2021-09-16 at 23 10 39

That is generating the project without including the source from transitive dependencies, and placing it in a subdirectory. Xcode 12.5.1.

I cant tell about a more complex scenario, we are not using this generated projects on scale yet

acecilia:
Also seems to work:
* Breakpoints in tests
* Showing tests in the issue navigator
* Going to a test by clicking on it in the issue navigator
* Running only one test by clicking on the play button next to the class declaration
* Properly updating the test status on the issue navigator (success/failure)

jerrymarino (commented):
Nice PR! One other thing to watch out for is debug settings with `target.source-map` - does that work for both swift/ objc? I assume you've tried to debug that e.g
 https://github.com/bazel-ios/rules_ios/pull/303/files#diff-23b9447656f353f462ed4e664c476453ecce99989c98337501a54abe8e416511R19

acecilia:
@jerrymarino I dont know what `target.source-map` is 🙏 How could I test it?

Seems like all projects generated by `xcodeproj` rule have [the line](https://github.com/bazel-ios/rules_ios/pull/303/files#diff-23b9447656f353f462ed4e664c476453ecce99989c98337501a54abe8e416511R19) you mentioned

thiagohmcruz:
> @jerrymarino I dont know what `target.source-map` is 🙏 How could I test it?
> 
> Seems like all projects generated by `xcodeproj` rule have [the line](https://github.com/bazel-ios/rules_ios/pull/303/files#diff-23b9447656f353f462ed4e664c476453ecce99989c98337501a54abe8e416511R19) you mentioned

You can launch `lldb` and type `apropos source-map`:
```sh
(lldb) apropos source-map
No commands found pertaining to 'source-map'. Try 'help' to see a complete list of debugger commands.

The following settings variables may relate to 'source-map':


  target.source-map -- Source path remappings are used to track the change of location between a source file when built, and where it exists on the current system.  It consists of an array of duples, the first
                       element of each duple is some part (starting at the root) of the path to the file when it was built, and the second is where the remainder of the original build hierarchy is rooted on the
                       local system.  Each element of the array is checked in order and the first one that results in a match wins.
```

That said I believe the dependency of `BAZEL_WORKSPACE_ROOT` on `SRCROOT` (see [this](https://github.com/bazel-ios/rules_ios/blob/a94a772df5cc076a77cdbcbe0be2b4f3912749d3/rules/xcodeproj.bzl#L874)) means this should be fine @jerrymarino? Also curious to know what you'd suggest to validate that otherwise.

tests/ios/xcodeproj/BUILD.bazel
```diff
@@ -186,3 +186,15 @@ xcodeproj(
         "//tests/ios/app:FW",
     ],
 )
+
+xcodeproj(
+    name = "Test-Custom-Output-Path",
+    additional_lldb_settings = [
+        "log enable -f /tmp/lldb.log lldb types",
+    ],
```
thiagohmcruz, line 194: So you're using an existing proj so you can diff the paths at the end right?

thiagohmcruz (approved):

tests/ios/xcodeproj/BUILD.bazel
acecilia: Yep. Since I cant test the project properly because current test infrastructure depends on having the xcode projects at `tests/ios/xcodeproj` (see issue description), I am using an alternative way of testing this. I am generating the same `Test-Custom-Output-Path` project under the `custom_output_path`, and then comparing the diff, which you can find in `tests/ios/xcodeproj/custom_output_path/test_expected_diff.txt `

tests/ios/xcodeproj/custom_output_path/test_expected_diff.txt
```diff
@@ -0,0 +1,13 @@
+diff -r ./tests/ios/xcodeproj/Test-LLDB-Logs-Project.xcodeproj/project.pbxproj ./tests/ios/xcodeproj/custom_output_path/Test-LLDB-Logs-Project.xcodeproj/project.pbxproj
```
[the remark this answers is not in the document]
acecilia: Done

thiagohmcruz:
I think this is good to land? @jerrymarino @gyfelton

thiagohmcruz:
@acecilia let me know if there's anything else you want to add or need feedback here. If not I'll land it.

jerrymarino (approved):

acecilia:
Thanks everybody! 🙌

jerrymarino:
You've got it! @acecilia would you confirm if the following cases are still working:

1. overriding bazel in-tree, say `tools/bazel`
2. generating the project at the workspace root: e.g. `bazel-project.xcodeproj`

It has broken both of these cases which I assume that is probably a common way that people are using it. 

cc @thiagohmcruz ^

thiagohmcruz:
> You've got it! @acecilia would you confirm if the following cases are still working:
> 
> 
> 
> 1. overriding bazel in-tree, say `tools/bazel`
> 
> 2. generating the project at the workspace root: e.g. `bazel-project.xcodeproj`
> 
> 
> 
> It has broken both of these cases which I assume that is probably a common way that people are using it. 
> 
> 
> 
> cc @thiagohmcruz ^

Yep, I'd appreciate too if you could check that. I assume this was an opt-in feature and wouldn't affect the default use case

PR #304, opened 2021-09-21
Add support for mixed objective-c and pure c++ code in same apple_library

ivan-golub:
**Background:**
- It is possible to create library that will contain mix of objective-c, objective-c++, pure c and pure c++ code in vanilla Xcode.
- [objc_library](https://docs.bazel.build/versions/main/be/objective-c.html#objc_library) supports all languages above as well
- Current `apple_library` adds .m, .mm, .c sources as well as .hm .hh, .hpp headers headers but excludes .cc and .cpp which causes compilation error if library has pure c++ code

**Changes:**
- Add c++ source code (.cc and .cpp) files in srcs of apple_library
- Add test project that mixes plain objective-c and pure c++

**Test Plan:**
- Run `bazel test //tests/ios/frameworks/objcpp:ObjCppMixedLibTests `

**Risks:**
- Logic of https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl#L866 is not entirely clear from comment. If enable it, then lib containing both c++ and objective-c will create two libraries, one with objective-c sources and objective-c + c++ headers (likely not compilable due to missing c++ implementations) and cc library that contains all objective-c private headers and c++ implementations. `cc_library` idea needs clarification and either should be removed and this PR merged or revived in form that will allow tests from this PR to pass

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -898,7 +898,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
 
     objc_library(
         name = objc_libname,
-        srcs = objc_sources + objc_private_hdrs + objc_non_exported_hdrs,
+        srcs = objc_sources + objc_private_hdrs + objc_non_exported_hdrs + cpp_sources,
         non_arc_srcs = objc_non_arc_sources,
         hdrs = objc_hdrs,
         copts = copts_by_build_setting.objc_copts + objc_copts + additional_objc_copts + index_while_building_objc_copts,
--- a/tests/ios/frameworks/objcpp/BUILD.bazel
+++ b/tests/ios/frameworks/objcpp/BUILD.bazel
@@ -1,0 +1,56 @@
+load("//rules:framework.bzl", "apple_framework")
+load("//rules:test.bzl", "ios_unit_test")
+
+apple_framework(
+    name = "ObjCppMixedLib",
+    srcs = glob(
+        [
+            "ObjCppMixedLib/*.mm",
+            "ObjCppMixedLib/*.cpp",
+        ],
+    ),
+    objc_copts = [
+        "-fmodules",
+        "-fcxx-modules",
+    ],
+    platforms = {"ios": "10.0"},
+    public_headers = glob(
+        [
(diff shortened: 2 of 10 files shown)


nkoroste:
@jerrymarino fyi

rules/library.bzl
```diff
@@ -898,7 +898,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
 
     objc_library(
         name = objc_libname,
-        srcs = objc_sources + objc_private_hdrs + objc_non_exported_hdrs,
+        srcs = objc_sources + objc_private_hdrs + objc_non_exported_hdrs + cpp_sources,
```
jerrymarino, line 901: I wonder what the the tradeoffs of this line and https://github.com/bazel-ios/rules_ios/pull/304/files#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478L867 ?

I don't know why https://github.com/bazel-ios/rules_ios/pull/304/files#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478L867 was disabled, one of the problems with putting the sources in the same library is there is no way to specify the C++ specific copts at the library level.

jerrymarino (commented):
Thanks for the PR and adding the example! 

I'd be in favor to re-enable https://github.com/bazel-ios/rules_ios/pull/304/files#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478L867 to set the language level copts at the lib level

jerrymarino:
Also Cocoapods, it compiles C++ sources as a separate [`objc_library`](https://github.com/pinterest/PodToBUILD/blob/master/Sources/PodToBUILD/ObjcLibrary.swift#L154) maybe we can just apply similar logic in that line to make the logic more close to that.

mattrobmattrob:
Closing since this should be handled in https://github.com/bazel-ios/rules_ios/pull/550. Reopen https://github.com/bazel-ios/rules_ios/issues/537 if it doesn't!

PR #310, opened 2021-09-24
[VFS] Nest frameworks in FRAMEWORK_PATH

jerrymarino:
Currently root entries in the vfsoverlay are formed with the path
`/build_bazel_rules_ios/frameworks/$FRAMEWORK.framework`. This causes
issues when these vfsoverlays are used within Xcode's indexing system,
and Bazel's dependency pruning code. For instance, a path of a header
would be
`/build_bazel_rules_ios/frameworks/$FRAMEWORK.framework/Headers/Header.h`
as opposed to `$SRCROOT/path/to/Header.h`. In the IDE, this bug can
break jump to definition under some situations and dependency pruning
in Bazel.

The is a longstanding issue in `rules_ios`'s
`rules/framework/vfs_overlay.bzl` rule. It impacts both virtual or non
virtual framework usage of the rule.

By putting the framework in it's own path entry, the `VirtualFileSystem`
correctly serializes the paths.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -201,15 +201,15 @@ def _make_root(vfs_parent, bin_dir_path, build_file_path, framework_name, swiftm
         })
 
     if len(swiftmodules) > 0:
-        contents = _provided_vfs_swift_module_contents(swiftmodules, vfs_prefix, FRAMEWORK_SEARCH_PATH)
+        contents = _provided_vfs_swift_module_contents(swiftmodules, vfs_prefix)
         if contents:
             roots.append(contents)
     elif has_swift:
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):

congt (approved):

rules/framework/vfs_overlay.bzl
```diff
@@ -360,8 +359,12 @@ def make_vfsoverlay(ctx, hdrs, module_map, private_hdrs, has_swift, swiftmodules
         "version": 0,
         "case-sensitive": True,
         "overlay-relative": True,
-        "use-external-names": False,
-        "roots": roots,
+        "use-external-names": True,
```
gyfelton, line 362: can explain a bit / point to doc explaining what does this boolean do? Tried to find its doc but can only fine VirtualFileSystem.h doc for it :pray: @jerrymarino

Issue #311, opened 2021-09-24
[xcodeproj] Project generation issue


jerrymarino:
For me the following cases are not working:

1. overriding bazel in-tree, say tools/bazel
2. generating the project at the workspace root: e.g. bazel-project.xcodeproj

It has broken both of these cases which I assume that is probably a common way that people are using it. We may want to harden the testing infra or consider replacing it wit XCHammer altogether which tests generation at the root.

jerrymarino:
Seems caused by this: https://github.com/bazel-ios/rules_ios/pull/303

PR #312, opened 2021-09-27
Fix project generation when xcodeproj rule is declared at the top level of the workspace

acecilia:
Fixes https://github.com/bazel-ios/rules_ios/issues/311

I tried to add a test for this at top level, but current test scripts rely on xcodeproj tests being under `tests/ios/xcodeproj`


gyfelton (approved):

thiagohmcruz (approved):
Cc @jerrymarino want to take a quick 👀 here since you filed the issue?

thiagohmcruz:
Thx so much @acecilia btw!

PR #314, opened 2021-09-30
Fix LLDB internal module compilation error

jerrymarino:
When LLDB compiles swiftmodules during attachment and `po` it is missing
some clang importer arguments. This adds missing copts directly to
LLDB in the lldbinit.

Note that I have also added a _example_ test case for LLDB and clarified some of
the requirments here. For now, it atleast documents this requirment:
```
Xcode project: bazel run tests/ios/xcodeproj:Test-Imports-App-Project
&& open tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj/
Scheme: `TestImports-App`
Breakpoints: tests/ios/unit-test/test-imports-app/empty.swift:6
Breakpoints: tests/ios/unit-test/test-imports-app/empty.swift:15
Xcode Version: 12.5.1
```

Longer term, can "Reverse sherlock" some LLDB automation testing from
swift-LLDB to verify debugging of many examples. Testing is further
blocked by getting the debugger working on the examples. As of today,
the examples just don't work on master
https://github.com/bazel-ios/rules_ios/issues/281

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -701,9 +701,12 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = " ".join(
             ["\"%s\"" % d for d in defines_without_equal_sign],
         )
-        target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(
-            ["-D%s" % d for d in target_info.cc_defines.to_list()],
-        )
+        extra_clang_flags = ["-D%s" % d for d in target_info.cc_defines.to_list()]
+
+        if virtualize_frameworks:
+            extra_clang_flags += ["-Xcc %s" % d for d in _objc_copts_for_target(target_name, all_transitive_targets)]
+
+        target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(extra_clang_flags)
 
         target_settings["BAZEL_LLDB_INIT_FILE"] = lldbinit_file
 
--- a/tests/ios/unit-test/test-imports-app/Header.h
+++ b/tests/ios/unit-test/test-imports-app/Header.h
@@ -2,4 +2,4 @@
 // - Xcode target name
 // - "Module" name
 #import <TestImports-App/Header2.h> // Path from headermap
-#import <TestImports_App/Header2.h> // Actual path
+// #import <TestImports_App/Header2.h> // Actual path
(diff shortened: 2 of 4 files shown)


tests/ios/unit-test/test-imports-app/Header.h
```diff
@@ -1 +1,5 @@
-#import <TestImports-App/Header2.h>
+// Note: rules_ios currently supports both of the names.
+// - Xcode target name
+// - "Module" name
+#import <TestImports-App/Header2.h> // Path from headermap
+#import <TestImports_App/Header2.h> // Actual path
```
thiagohmcruz, line 5: CI is saying this was not found? Is `<>` the expected usage here for the actual path?

rules/xcodeproj.bzl
```diff
@@ -701,9 +701,12 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = " ".join(
             ["\"%s\"" % d for d in defines_without_equal_sign],
         )
-        target_settings["BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS"] = " ".join(
-            ["-D%s" % d for d in target_info.cc_defines.to_list()],
-        )
+        extra_clang_flags = ["-D%s" % d for d in target_info.cc_defines.to_list()]
+
+        if virtualize_frameworks:
+            extra_clang_flags += ["-Xcc %s" % d for d in _objc_copts_for_target(target_name, all_transitive_targets)]
```
thiagohmcruz, line 707: I assume this list can potentially be very long? Any chance of hitting ARGMAX somewhere? Just mentioning this in case it's worth running this against one complex app first

Ideally that would be automated but we haven't had the time yet to improve CI and run it against one of the internal apps

thiagohmcruz (approved):
Thx for fixing this! Added a couple of questions inline.

rules/xcodeproj.bzl
gyfelton: looking at the implementation of `_objc_copts_for_target` I don't think the list will be long since it is for `target_name` only not all of its deps?

thiagohmcruz: Right yeah, also the same list goes in `OTHER_CFLAGS` so what I'm bringing up here would've been noticed before if it was a problem. K thx!

jerrymarino: Yeah I don't think either of these would be long in practice, the `ARGMAX` issue can probably happen without virtual frameworks enabled given a large enough app

tests/ios/unit-test/test-imports-app/Header.h
jerrymarino: Yeah I think that one is probably a fringe behavior that isn't 100% supported.

PR #315, opened 2021-09-30
Multiple architecture support in apple_framework_packaging

### LINKED ISSUE #186: Multi architecture support / ios_multi_cpus

congt:
This fixes #186.

When building for multiple archs, we need to uses ctx.split_attr.[attr] to access the attributes which attach the apple_common.multi_arch_split transition. This allows us to use the artifacts built for the correct arch in the current target. To identify the correct configuration of a target at build time, we use ctx.fragments.apple.single_arch_cpu and tx.fragments.apple.single_arch_platform.platform_type to find the current architecture and Apple platform type.


rules/framework.bzl
```diff
@@ -512,8 +512,28 @@ def _merge_root_infoplists(ctx):
...
 def _apple_framework_packaging_impl(ctx):
-    framework_files = _get_framework_files(ctx)
+    # The current build architecture
+    arch = ctx.fragments.apple.single_arch_cpu
+
+    # The current Apple platform type, such as iOS, macOS, tvOS, or watchOS
+    platform = str(ctx.fragments.apple.single_arch_platform.platform_type)
+
+    # When building with multiple architectures (e.g., --ios_multi_cpus=x86_64,arm64), we should only pick the slice for the current configuration.
+    split_slice_key = "{}_{}".format(platform, arch)
```
gyfelton, line 531: how abt `arm64`?

rules/framework.bzl
```diff
@@ -621,6 +641,7 @@ apple_framework_packaging = rule(
         ),
         "transitive_deps": attr.label_list(
             mandatory = True,
+            cfg = apple_common.multi_arch_split,
```
gyfelton, line 644: this `cfg` gives us those fragments right?

rules/framework.bzl
```diff
@@ -146,9 +146,9 @@ def _concat(*args):
             arr += x
     return arr
 
-def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields):
+def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields, deps, transitive_deps, vfs):
     import_vfsoverlays = []
-    for dep in ctx.attr.vfs:
```
gyfelton, line 151: hope we have a way to forbid us from using `ctx.attr` instead inject them in so that we don't encounter bug because of references to ctx.attr.<xxx>

rules/framework.bzl
congt: If we build with --ios_multi_cpus=arm64, split_slice_key is ios_arm64. Is it what you're asking?

rules/framework.bzl
congt: apple_common.multi_arch_split is a 1:k transition. When we attach a transition to an attr, we need to set the transition to the cfg. This applies a 1:k outgoing transition to transitive_deps.

rules/framework.bzl
congt: Yeah, but I don't find any way to do that.

rules/framework.bzl
```diff
@@ -512,8 +512,28 @@ def _merge_root_infoplists(ctx):
 
     return output_plist
 
+def _attrs_for_split_slice(attrs_by_split_slices, split_slice_key):
+    if len(attrs_by_split_slices.keys()) == 0:
+        return []
+    elif len(attrs_by_split_slices.keys()) == 1:
+        return attrs_by_split_slices.values()[0]
+    else:
+        return attrs_by_split_slices[split_slice_key]
```
thiagohmcruz, line 521: Should this fail `if split_slice_key not in attrs_by_split_slices`?

rules/framework.bzl
gyfelton: ahh i see yup makes sense!

rules/transition_support.bzl
```diff
@@ -15,13 +15,17 @@ def _current_apple_platform(apple_fragment, xcode_config):
 
 def _cpu_string(platform_type, settings):
     """Generates a <platform>_<arch> string for the current target based on the given parameters."""
+
+    # If the cpu value has already been transformed to the correct value, we must not change it anymore.
+    # Otherwise, we may build for the wrong arch.
+    cpu_value = settings["//command_line_option:cpu"]
+    if (platform_type == "macos" and cpu_value.startswith("{}_".format(platform_type))) or cpu_value.startswith("{}_".format(platform_type)):
+        return cpu_value
```
thiagohmcruz, line 23: So IIUC in the old logic `ios_multi_cpus` would return early if present and this check wouldn't happen. By moving this up you're giving `cpu` a higher priority and not doing the check for `ios_multi_cpus` that would've happened in the old logic.

Can this change in behaviour be considered a breaking change?

docs/transitions_and_multi_arch_support.md
```diff
@@ -0,0 +1,60 @@
+# Transitions and Multiple Architecture Support
```
thiagohmcruz, line 1: Thx for adding this!

thiagohmcruz (approved):
Thx for adding the docs here! Added a couple of questions inline. Curious to know if what would be the best way to test this? Is there any check we can do in the outputs to validate this behaviour?

rules/framework.bzl
congt: The [] operator will fail if the key is not in the dict.

rules/transition_support.bzl
congt: The original logic is broken for multi arches because 1) ios_multi_cpus  is never changed by any transition, and 2) `apple_common.multi_arch_split` may have already changed cpu based on ios_multi_cpus.

Consider this example: two frameworks ( fw1 -> fw2) are built with --ios_multi_cpus=x86_64,arm64. The outgoing edge transition of fw1 will change the cpu of the configured target fw2_arm64 to ios_arm64. In the original logic, the incoming edge transition of fw2 will always change cpu to ios_x86_64 because x86_64 is the first one of ios_multi_cpus. Then fw2_arm64 is actually built for x86_64.

This transition behavior doesn't change for a single ios_multi_cpus. Therefore, I will not say it's a breaking change.

gyfelton (approved):

rules/transition_support.bzl
thiagohmcruz: Awesome thx for the detailed explanation!

PR #323, opened 2021-10-04
Genrule mixed module build test

jerrymarino:
Per discussions on github last week we found a hole in test coverage.
While this may not be failing specifically, it rules out the possibility
of future regressions

### DIFF
--- a/tests/ios/frameworks/genrule-inputs/b/lib.swift
+++ b/tests/ios/frameworks/genrule-inputs/b/lib.swift
@@ -1,0 +1,10 @@
+import c
+
+public struct B {
+    public static func run() {
+       C.run()
+       guard LibC != "B" else {
+           fatalError()
+       }
+    }
+}
--- a/tests/ios/frameworks/genrule-inputs/c/BUILD.bazel
+++ b/tests/ios/frameworks/genrule-inputs/c/BUILD.bazel
@@ -1,0 +1,39 @@
+load("//rules:framework.bzl", "apple_framework")
+
+genrule(
+   name="lib-src.swift",
+   cmd = """
+cat > $(OUTS) <<-EOF
+public struct C {
+
+    public static func run() {
+       print("runs here") 
+       guard LibC == "LibC" else {
+           fatalError()
+       }
+    }
+}
+EOF
+   """,
+   outs = ["lib.swift"]
+)
+
+genrule(
+   name="lib-src.h",
+   cmd = """
+cat > $(OUTS) <<-EOF
+   @import Foundation;
+   static NSString *const LibC = @"LibC";
+EOF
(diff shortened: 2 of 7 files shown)


tests/ios/frameworks/genrule-inputs/b/lib.swift
```diff
@@ -0,0 +1,10 @@
+import c
+
+public struct B {
+    public static func run() {
+       C.run()
+       guard LibC != "B" else {
```
thiagohmcruz, line 6: Sorry If I'm missing something but the only place `LibC` is defined is in `tests/ios/frameworks/genrule-inputs/c/BUILD.bazel` as
```
static NSString *const LibC = @"LibC";
```
so why would you check for `"B"` here?

tests/ios/frameworks/genrule-inputs/c/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
+load("//rules:framework.bzl", "apple_framework")
+
+genrule(
+   name="lib-src.swift",
+   cmd = """
+cat > $(OUTS) <<-EOF
+public struct C {
+
+    public static func run() {
+       print("runs here")
```
thiagohmcruz, line 10: Debugging log that can be removed?

tests/ios/frameworks/genrule-inputs/b/lib.swift
jerrymarino: This is just some dummy code to both use the symbol and exercise it for a debugger test in the future.

jerrymarino: Also it needs to look at `LibC` to make sure that it can load a symbol from a header file defined in `LibC`

tests/ios/frameworks/genrule-inputs/c/BUILD.bazel
jerrymarino: For me it was valuable to have sentential to the user know this app was doing anything on startup other than just having a white screen

congt (approved):

tests/ios/frameworks/genrule-inputs/b/lib.swift
thiagohmcruz: K thx for clarifying!

tests/ios/frameworks/genrule-inputs/c/BUILD.bazel
thiagohmcruz: 👍

thiagohmcruz (approved):

gyfelton (approved):
Thx for increasing test coverage!

PR #324, opened 2021-10-04
Conditionally compile with transitive VFS

jerrymarino:
Conditionally adds VFS all targets, which right now has conditional side
effect to compile them all as virtual frameworks. In the interim we want
to include the VFS in order to pickup deps that are virtual frameworks.

It also includes a stopgap hack for issues with merging imported
frameworks. The inline comment may be addressed

Additionally, it builds the entire test suite with virtual frameworks
to pickup regressions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,6 +24,24 @@ jobs:
         with:
           name: bazel-testlogs
           path: bazel-testlogs
+  build_virtual_frameworks:
+    # Build the entire tree with this feature enabled. Longer term, we'll likely
+    # consider merging this feature into the default behavior and can re-align
+    # the CI job
+    name: Build ( Virutal Frameworks )
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Select Xcode 12.2
+        run: sudo xcode-select -s /Applications/Xcode_12.2.app
+      - name: Build and Test
+        run: |
+          bazelisk build //... --features apple.virtualize_frameworks
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: bazel-testlogs
+          path: bazel-testlogs
   buildifier:
     name: Check Starlark and Docs
     runs-on: macos-latest
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -669,34 +669,32 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
                 **kwargs
             )
 
-    if enable_framework_vfs:
-        framework_vfs_overlay(
-            name = framework_vfs_overlay_name_swift,
-            framework_name = module_name,
-            modulemap = module_map,
-            has_swift = len(swift_sources) > 0,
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -24,6 +24,24 @@ jobs:
         with:
           name: bazel-testlogs
           path: bazel-testlogs
+  build_virtual_frameworks:
+    # Build the entire tree with this feature enabled. Longer term, we'll likely
+    # consider merging this feature into the default behavior and can re-align
+    # the CI job
+    name: Build ( Virutal Frameworks )
```
gyfelton, line 31: same test as the other PR so need to remove one of them

rules/library.bzl
```diff
@@ -827,12 +824,18 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-            copts = copts_by_build_setting.swift_copts + swift_copts + additional_swift_copts,
-            deps = deps + private_deps + lib_names,
+            # Note: by default it used a vfs but not the entire virtual framwork
+            # feature.
+            copts = select({
+                "@build_bazel_rules_ios//:virtualize_frameworks": framework_vfs_swift_copts,
+                "//conditions:default": framework_vfs_swift_copts if enable_framework_vfs else [],
+            }) + copts_by_build_setting.swift_copts + swift_copts + additional_swift_copts,
+            deps = deps + private_deps + lib_names + select({
+                "@build_bazel_rules_ios//:virtualize_frameworks": [framework_vfs_overlay_name_swift],
```
gyfelton, line 834: to make sure I understand correctly: in either situation we need to include vfs info: this feature is enabled via a condition (aka transitive) OR enable_framework_vfs is set to TRUE. Am I correct here?

tests/ios/unit-test/test-imports-app/BUILD.bazel
```diff
@@ -76,5 +76,10 @@ ios_unit_test(
     visibility = ["//visibility:public"],
     deps = [
         ":TestImports-App_framework_unlinked",
-    ],
+    ] + select({
+        # Note: without using apple_framework, it won't set enable_framework_vfs
+        # imported it won't actually be seen
```
gyfelton, line 81: hard to understand what this line means... `who` imported `what` won't actually be seen?

.github/workflows/tests.yml
jerrymarino: The other PR has this one as a dependency, so once it is rebased it will be there 👍

tests/ios/unit-test/test-imports-app/BUILD.bazel
jerrymarino: Ah good point, I think this one is out of place.

rules/library.bzl
jerrymarino: Yep it uses a `select` statement to switch on `"@build_bazel_rules_ios//:virtualize_frameworks"`, and add other copts and deps. Longer term, I suspect it would be the default

PR #326, opened 2021-10-05
Supply FrameworkInfo from ad-hoc framework imports

jerrymarino:
Today, there are effectivly 2 different ways that frameworks are being
imported. The first one is `apple_library` where it creates a framework
import for adjacent frameworks. The other case is calling
`apple_*_framework_import` from the rules. For now, create framework
import for the latter, consider alignment of these codepaths longer
term.

This commit just instantiates a `FrameworkInfo` here so it's passed
upwards.

Includes @tinder-maxwellelliott 's reproducer


rules/apple_patched.bzl
```diff
@@ -80,12 +91,40 @@ def _apple_framework_import_modulemap_impl(ctx):
...
+        # imported from rules/library.bzl. There are effectivly 2 different ways
+        # that frameworks are being imported.  Otherwise, a test case for
+        # swiftmodules ensuring it extracts from the provider. The swiftmodules
+        # for this case should be loaded into the VFS
+        swiftmodules = [],
+        private_hdrs = [],
+        has_swift = False,
+        framework_name = imported_framework_name,
+    )
+    framework_info = FrameworkInfo(
```
jerrymarino, line 113: Perhaps this might need to be conditional until `virtualize_frameworks` is default.

tinder-maxwellelliott:
Trying to top hat this myself, seeing this error 

```
ERROR: /private/var/tmp/_bazel_maxwellelliott/db37ab0e01ac1d9a8e5ea282dd04a720/external/FirebaseAnalytics/BUILD.bazel:43:32: in _apple_framework_import_modulemap rule @FirebaseAnalytics//:GoogleAppMeasurement: 
        File "/private/var/tmp/_bazel_maxwellelliott/db37ab0e01ac1d9a8e5ea282dd04a720/external/build_bazel_rules_ios/rules/apple_patched.bzl", line 101, column 26, in _apple_framework_import_modulemap_impl
                vfs = make_vfsoverlay(
        File "/private/var/tmp/_bazel_maxwellelliott/db37ab0e01ac1d9a8e5ea282dd04a720/external/build_bazel_rules_ios/rules/framework/vfs_overlay.bzl", line 323, column 34, in make_vfsoverlay
                framework_name = ctx.attr.framework_name
Error: No attribute 'framework_name' in attr. Make sure you declared a rule attribute with this name.
```

rules/apple_patched.bzl
```diff
@@ -80,12 +92,43 @@ def _apple_framework_import_modulemap_impl(ctx):
...
+        vfsoverlay_infos = [vfs.vfs_info],
+        headers = hdrs_list,
+        private_headers = [],
+        modulemap = modulemap_list,
+        swiftmodule = [],
+        swiftdoc = [],
+    )
+
+    virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
+    additional_providers = [framework_info] if virtualize_frameworks else []
```
gyfelton, line 124: since `framework_info` is only needed if virtualize_frameworks feature is on, why need to construct it beforehand? or it has some side effects?

jerrymarino:
@tinder-maxwellelliott much appreciated! I wonder what commit sha that's failing on and if you're using `apple.virtualize_frameworks`.  Reason being, the line was changed : https://github.com/bazel-ios/rules_ios/pull/326/files#diff-450b1330485549712d2fd4e65b2b3fd0a7692fe8f4ec43a0b7b60bbb739313afR303 .

If this is passing your reproducer, I wonder if something else is happening differently for FirebaseAnaytics. GoogleAppMeasurement Is there any chance these are XCFrameworks? I checked out at their SDK and it seems so ( https://github.com/firebase/firebase-ios-sdk ) . Would you kindly post up what the build file looks like you're using for that?

FWIW I think the XCFramework virtual framework import will probably need to be updated to handle your build file, ATM it works with `apple_framework.vendored_xcframeworks`.

rules/apple_patched.bzl
jerrymarino: Good call 👍

tinder-maxwellelliott:
> @tinder-maxwellelliott much appreciated! I wonder what commit sha that's failing on and if you're using `apple.virtualize_frameworks`. Reason being, the line was changed : https://github.com/bazel-ios/rules_ios/pull/326/files#diff-450b1330485549712d2fd4e65b2b3fd0a7692fe8f4ec43a0b7b60bbb739313afR303 .
> 
> If this is passing your reproducer, I wonder if something else is happening differently for FirebaseAnaytics. GoogleAppMeasurement Is there any chance these are XCFrameworks? I checked out at their SDK and it seems so ( https://github.com/firebase/firebase-ios-sdk ) . Would you kindly post up what the build file looks like you're using for that?
> 
> FWIW I think the XCFramework virtual framework import will probably need to be updated to handle your build file, ATM it works with `apple_framework.vendored_xcframeworks`.

At the end of the day we are simply calling `apple_dynamic_framework_import` with the nested framework path inside of the xcframework. The issue seems to be because there are no headers in this framework

Screen Shot 2021-10-05 at 12 53 58 PM

gyfelton (approved):

PR #332, opened 2021-10-07
Use --copy-links in rsync call to copy symlink files

erikkerber:
This is mostly relevant if:

- Not using index-while-building
- Build a scheme with a non-app primary output

Specifically, headers were not being copied over because they were symlinks and the default behavior of rsync is to skip these. Xcode indexing would fail because the modulemap would point to these files that were not being copied to derived data.

image
image


tinder-maxwellelliott:
Thanks for fixing this @erikkerber , I fixed this myself locally but did not think to upstream it 🤦🏻

gyfelton:
Thx for the contribution! CI failure are all because of `Error: 'NoneType' value has no field or method 'compare_to'`, is this sth you can fix?

jerrymarino (approved):

PR #333, opened 2021-10-07
Fix and assert swift imported header propagation

jerrymarino:
This PR fixes a bug in swift compilation with imported framework
transitive headers. Additionally it refactors the testing rules into a
package that can be used with `rules_ios` or other repositories. These
tests iclude existing tests for Bazel transitions being incorrect in `rules_ios`
and the new generalized header test. The headers test allows a user to
assert that tests are propagated to actions

For cases:

1. add a frameworks directory to the app. This represents the case of
   `apple_framework.vendored_static_frameworks` which is commonly used
   as a way to import frameworks

2. test framework import under test-imports-app. Asserts transitive headers
   are loaded from TesnorFlow.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -161,17 +161,14 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
     for dep in transitive_deps + deps:
         # Collect transitive headers. For now, this needs to include all of the
         # transitive headers
-        if not CcInfo in dep:
-            continue
-        compilation_context = dep[CcInfo].compilation_context
-        propagated_interface_headers.append(compilation_context.headers)
-
-        if not FrameworkInfo in dep:
-            continue
-        framework_info = dep[FrameworkInfo]
-        fw_dep_vfsoverlays.extend(framework_info.vfsoverlay_infos)
-        framework_headers = depset(framework_info.headers + framework_info.modulemap + framework_info.private_headers)
-        propagated_interface_headers.append(framework_headers)
+        if CcInfo in dep:
+            compilation_context = dep[CcInfo].compilation_context
+            propagated_interface_headers.append(compilation_context.headers)
+        if FrameworkInfo in dep:
+            framework_info = dep[FrameworkInfo]
+            fw_dep_vfsoverlays.extend(framework_info.vfsoverlay_infos)
+            framework_headers = depset(framework_info.headers + framework_info.modulemap + framework_info.private_headers)
+            propagated_interface_headers.append(framework_headers)
 
     outputs = framework_files.outputs
     vfs = make_vfsoverlay(


rules/framework.bzl
```diff
@@ -159,22 +160,20 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
     # We need to map all the deps here - for both swift headers and others
     fw_dep_vfsoverlays = []
     for dep in transitive_deps + deps:
+        # Collect transitive headers. For now, this needs to include all of the
+        # transitive headers
+        if not CcInfo in dep:
+            continue
+        for h in dep[CcInfo].compilation_context.headers.to_list():
+            propagated_interface_headers.append(depset([h]))
```
gyfelton, line 168: will h.is_source be true for headers? Or true only to a set of headers? The original logic of continue if is_source is baffling me...

rules/framework.bzl
```diff
@@ -159,22 +160,20 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
     # We need to map all the deps here - for both swift headers and others
     fw_dep_vfsoverlays = []
     for dep in transitive_deps + deps:
+        # Collect transitive headers. For now, this needs to include all of the
+        # transitive headers
+        if not CcInfo in dep:
```
congt, line 165: Could you explain why the order matters? Is it because an dep (except apple_framework_packaging targets) may return CcInfo, but won't return FrameworkInfo? Could we change it to something like
```
if CcInfo in dep:
  ...
if FrameworkInfo in dep:
 ...
```

jerrymarino: > Could you explain why the order matters? 
The code has a `continue` statement inside of the if block so it will skip ahead. 

> Could we change it to something like

I haven't tried that out yet but wouldn't be opposed to it. What are the benefits to changing the code to that?

rules/apple_patched.bzl
```diff
@@ -95,6 +95,7 @@ def _get_framework_info_providers(ctx, old_cc_info, old_objc_provider):
         private_hdrs = [],
         has_swift = False,
         framework_name = imported_framework_name,
+        extra_search_paths = imported_framework_name,
```
thiagohmcruz, line 98: I don't see this being used?

tests/ios/unit-test/test-imports-app/frameworks/Basic/BUILD.bazel
```diff
@@ -0,0 +1,13 @@
+load("//rules:framework.bzl", "apple_framework")
+
+apple_framework(
+    name = "Basic",
+    # Note: it is totally possible that a user would write a glob like this..
+    # srcs = glob([
+    #    "sources/Basic/**/*.h",
+    # ]),
```
thiagohmcruz, line 8: So the ideal is to keep this as a note for someone trying to understand what is under test right? Why is it relevant to know that such a `glob` could be written?

thiagohmcruz (approved):
Thx for adding these! A few questions inline 🙏

rules/framework.bzl
congt: > The code has a continue statement inside of the if block so it will skip ahead.

I got that, but was  wondering why we need the headers of imported frameworks.

> What are the benefits to changing the code to that?

It might just be me, but the current logic is a little hard to understand. I had to know that if a dep doesn't return CcInfo, it won't return FrameworkInfo, either.

congt (approved):

rules/analysis_tests/transitive_header_test.bzl
```diff
@@ -0,0 +1,32 @@
...
+    env = analysistest.begin(ctx)
+    asserts.true(env, len(ctx.attr.deps) > 0)
+    target_under_test = analysistest.target_under_test(env)
+    target_headers = target_under_test[CcInfo].compilation_context.headers.to_list()
+    for dep in ctx.attr.deps:
+        asserts.true(env, CcInfo in dep)
+        dep_headers = dep[CcInfo].compilation_context.headers.to_list()
+        for dep_header in dep_headers:
+            # Assert that all of the dep headers are in the target
+            if not dep_header.extension == "h":
```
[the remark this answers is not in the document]
jerrymarino: We could by default have that and better expose this to the user to match other extensions if they'd like

tests/ios/unit-test/test-imports-app/frameworks/Basic/BUILD.bazel
jerrymarino: Yeah it is possible a user to craft an redundant build file which duplicates `srcs` and `framework_imports` - probably better to have the comment but I can remove it doesn't make sense. The rule will handle it anyways.

thiagohmcruz: I'm ok with keeping it. Was just curious to know the details of the edge case you're mentioning here 👍

rules/analysis_tests/identical_outputs_test.bzl
```diff
@@ -0,0 +1,82 @@
...
+_TestFiles = provider(
+    fields = {
+        "files": "A glob of files collected for later assertions",
+    },
+)
+
+def _identical_outputs_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    all_files = []
+    asserts.true(env, len(ctx.attr.deps) > 1)
```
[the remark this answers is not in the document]
jerrymarino: The test doesn't assert the problem to have 1 dev because it asserts identical outputs amongst multiple `deps` https://github.com/bazel-ios/rules_ios/pull/333/files/5138482e5b1fcd470f76b793ef40e6ebc930cc73#diff-fa2c728306ed72d291b632ffa2b0458ea872ba83445c911667ac90ff0c4263a0R56

This is the only way to ensure that https://github.com/bazelbuild/bazel/issues/12171 isn't happening internal to a rule!

.github/workflows/tests.yml
```diff
@@ -28,15 +30,24 @@ jobs:
...
+          # Host config
+          bazelisk test --features apple.virtualize_frameworks --local_test_jobs=1 -- //... -//tests/ios/...
+
+          # `deleted_packages` is needed below in order to override the value of the .bazelrc file
+          bazelisk test  --features apple.virtualize_frameworks \
+              --local_test_jobs=1 \
+              --apple_platform_type=ios \
+              --deleted_packages='' \
+              -- //tests/ios/... \
+              -//tests/ios/frameworks/sources-with-prebuilt-binaries/... # Needs more work for pre-built binaries
```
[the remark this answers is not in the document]
jerrymarino: Want to file an issue to see what is wrong with it? I don't have an idea about what to do ATM

rules/framework.bzl
jerrymarino: Also we be better off just checking the providers like here https://github.com/bazel-ios/rules_ios/pull/333/files/5138482e5b1fcd470f76b793ef40e6ebc930cc73#r724580342 I don't believe there's exclusivity here but need to test it a bit more.

PR #346, opened 2021-11-04
A rule to update ARM dependencies to run on M1 arm64 sim

jerrymarino:
This PR is an RFC on an _additional_ rule to rules_ios that adds the ability to
update the Mach-o header on imported libraries and frameworks to get arm64
binaires running on Apple silicon simulator. For the demo, it's added in
`app.bzl` but will need a feature flag and more.

Why bother doing this? Well some apps have many dependencies which could take
along time on vendors or other parties to update. Because the M1 chip has the
same ISA as iOS ARM64, most binaries will run transparently. Much iOS developers
code is high level enough and isn't specifc to a device or simulator. There are
many caveats and exceptions but getting it running is better than nothing. ( e.g.
`TARGET_OS_SIMULATOR` )

This solves the problem at the build system level with the power of bazel. The
idea is pretty straight forward:
1. collect all imported paths
2. update the macho headers with Apples vtool and arm-to-sim
3. update the linker invocation to use the new libs

Now it updates all of the inputs automatically - the action can be taught to do
all of this conditionally if necessary.

Additionally the apsect was generating the action but I moved it into the rule.
Mainly because you can't make an executable input to an aspect to scheudle
actions there.  This has an interesting propery: you get a single path for
framework lookups at link time.  Today with bazel and rules_ios ld has an
O(MFrameworks*NFrameworkPaths) search requirmenet. This is an additional
consequence and worth comparing link time impovements but is a future exercise.

There are some minor followups, but seeing what people think about it


tools/m1_utils/update_dynamic.sh
```diff
@@ -0,0 +1,21 @@
...
+# Note: this program is the simplest version of patching a dynamic library
+# Other code should handle making frameworks, etc
+patch() {
+     EXTDIR=$(mktemp -d)
+     OUTD="$PWD"
+     pushd $EXTDIR
+     lipo "$FWF" -thin arm64 -output "$OF.ar" || cp "$FWF" "$OF.ar"
+
+     # FIXME: Versions should be input from the build system
+     xcrun vtool -arch arm64 -set-build-version 7 11.0 11.0 -replace -output "$OUTD/$OF" "$OF.ar"
```
bogo, line 12: I remember having to force codesign the output binary here due to restrictions of Apple silicon:
```
xcrun codesign --force --sign - "$OUTD/$OF"
```

...but this could cause the builds to not be reproducible per @keith's [article](https://www.smileykeith.com/2021/10/05/codesign-m1/). Did you run into this as well?

bogo:
👍 Love the idea!

Every time we bump one of the few holdout frameworks, it's an adventure full of unarchiving, transmogrifying, archiving,. vtooling, codesigning, etc. This would make the whole process transparent and will be even more important down the line, when these holdout frameworks become increasingly rare and obscure.

One (likely completely out of scope) improvement here could be automating `.framework` transmogrification into a multi-platform XCFramework. Then it'd be a truly one stop shop.

jerrymarino:
@bogo first off, thanks for the excellent write-up on all of this! Second, thank you for publishing the source to handle the object file fix-up 🙏 

Having this capability in the build system dramatically simplifies dealing with this and makes M1 a reality. There's a lot of iOS dependencies out there without an xcframework right now and a lot of people looking to get up and running on the M1 simulator. I suspect a decent amount of these dependencies could continue to have updates that don't have an xcframework slice, making updates an ongoing problem.

> One (likely completely out of scope) improvement here could be automating transmogrification of older .frameworks into multi-platform .xcframeworks. Then this would be a truly one stop shop.

The current version of this code takes any `.framework` and provides a compatible one to the linker and bundling so there is no need to create an xcframework for the Bazel with this for apps just running on a simulator. To your point, it would be a nice ability to _produce_  an xcframework for consumers.  With the primitives in this PR, you could take the output frameworks from the `import_middleman` and produce a new shiny xcframework - coming full circle! We should be able to produce xcframeworks from Bazel in some capacity, I'm not 100% sure that exists yet.

tools/m1_utils/update_dynamic.sh
jerrymarino: > I remember having to force codesign the output binary here due to restrictions of Apple silicon:

This is handled here by the Bazel bundling rules, [rules_apple](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/framework_import.bzl#L181) - mainly the that code interfaces with the output    artifacts of the `import_middleman` applies an ad-hoc code signature. Pretty nice!

A reproducibility situation can problematic one for Bazel depending on how the outputs from actions are consumed. In some cases, the bundling rules we use here Bazel already assume this because developers [might have a different identity on each machine](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/partials/framework_import.bzl#L181). Because of many factors, I don't have issues with the reproducibility problems  in that article yet. It'd be ideal to get rules_apple codesign reproducible - if there's any action items we should file a bazelbuild issue.

jerrymarino:
Fixed many issues and cleaned this to remove RFC status. There's a couple planned TODOs which I'll fix tomorrow AM:
- [ ] build the arm-to-sim dep from source or create a gh action to release a fat binary
- [ ] conditionally enable the middleman feature

Additionally, as other PRs:
- [ ] Form M1 sim in general, Bazel 5 is a dep but needs more work for release. I can publish stand-in Bazel 4 "frankenstein" c++ toolchain mentioned on [mobile native forum](https://github.com/MobileNativeFoundation/discussions/discussions/141#discussioncomment-1621530)
- [ ] follow up on hack in rules_swift toolchain or publish a tag with above hack

rules/app.bzl
```diff
@@ -55,9 +56,15 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
...
+    import_middleman(name = name + ".import_middleman", deps = library.deps)
     rules_apple_ios_application(
         name = name,
-        deps = library.lib_names,
+        # TODO: add a conditional define for this feature
+        deps = [name + ".import_middleman"],
+
+        # TODO: add a conditional define for apple silicon
+        # deps = library.deps,
+        output_discriminator = None,
```
thiagohmcruz, line 67: Are these TODOs for this PR or for follow up work?

rules/import_middleman.bzl
```diff
@@ -0,0 +1,227 @@
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleFrameworkImportInfo")
+load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "AppleSupportToolchainInfo")
+
+_Provider = provider(fields = {
```
thiagohmcruz, line 4: Nit: Should this name be more descriptive instead of `_Provider`?

rules/import_middleman.bzl
```diff
@@ -0,0 +1,227 @@
...
+    "imported_library_file": "",
+    "static_framework_file": "",
+    "dynamic_framework_file": "",
+    "import_infos": "",
+})
+
+def _do_framework(ctx, framework):
+    # Enusre that this will correctly propagate all framework files
```
thiagohmcruz, line 12:
```suggestion
    # Ensure that this will correctly propagate all framework files
```

rules/import_middleman.bzl
```diff
@@ -0,0 +1,227 @@
...
+
+     ditto "$FW_DIR" "$OUT_DIR"
+     "$TOOL" "$OUT_DIR/$(basename "$FRAMEWORK_BINARY")"
+   """.format(ctx.executable.update_in_place.basename, framework.path, out_dir.path)
+
+    ctx.actions.run_shell(outputs = [out_dir, out_file], inputs = depset([framework]), command = cmd)
+    return out_file
+
+def _do_lib(ctx, imported_library):
+    # Enusre that this will correctly propagate all framework files
```
thiagohmcruz, line 34: Super nit: explicitly say "all library files" here instead of "framework"?

rules/import_middleman.bzl
```diff
@@ -0,0 +1,227 @@
...
+        "sdk_framework",
+        "weak_sdk_framework",
+        "imported_library",
+        "force_load_library",
+        "source",
+        "link_inputs",
+        "linkopt",
+        "library",
+        "dynamic_framework_file",
+        "static_framework_file",
```
thiagohmcruz, line 146: Mind sharing how did you build this list?

rules/import_middleman.bzl
```diff
@@ -0,0 +1,227 @@
...
+        **objc_provider_fields
+    )
+
+    return [
+        DefaultInfo(files = depset(dynamic_framework_dirs)),
+        objc,
+        _make_imports([depset(dynamic_framework_dirs)]),
+    ]
+
+import_middleman = rule(
```
thiagohmcruz, line 221: I'd appreciate if you could add a doc to this rule to help others in the future understand the goals, usage, expected outputs etc

rules/test.bzl
```diff
@@ -58,9 +59,12 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
 
     library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": ios_test_kwargs.get("minimum_os_version")}, **kwargs)
 
+    # TODO: make this conditional
+    import_middleman(name = name + ".import_middleman", deps = library.lib_names)
     rule(
         name = name,
-        deps = library.lib_names,
+        # deps = library.lib_names,
+        deps = [name + ".import_middleman"],
```
thiagohmcruz, line 67: I assume we want this to be configurable by a config_setting instead of replacing the current implementation right?

thiagohmcruz (commented):
Just a few questions on the starlark side, are you planning to track the `TODO`s in issues or is the goal to resolve all in this PR?

rules/import_middleman.bzl
jerrymarino: SGTM

thiagohmcruz (approved):
🚀

PR #348, opened 2021-11-09
Allow opting out of the default umbrella header, to support strict imports

acecilia:
👋 

This PR better supports strict imports, meaning [files only import what they use, and they import everything they use](https://forums.swift.org/t/supporting-strict-imports/42472). 

See documentation in code for the full context 🙏

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -125,7 +125,7 @@ module {module_name} {{
 def _write_umbrella_header(
         name,
         library_tools,
-        generate_legacy_umbrella_header,
+        generate_default_umbrella_header,
         public_headers = [],
         private_headers = [],
         module_name = None,
@@ -137,7 +137,7 @@ def _write_umbrella_header(
 
     content = ""
 
-    if generate_legacy_umbrella_header:
+    if generate_default_umbrella_header:
         content += """\
 #ifdef __OBJC__
 #    import <Foundation/Foundation.h>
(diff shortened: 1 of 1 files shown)


acecilia:
I think the failure on this build is not related with the PR. cc @gyfelton @thiagohmcruz

rules/library.bzl
```diff
@@ -439,6 +451,24 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+    #
+    # As a result, when writing swift code, there is no need for importing neither Foundation nor
+    # UIKit. See:
+    # * https://github.com/facebookarchive/xcbuild/issues/92#issuecomment-234400427
+    #
+    # But these automatic imports are a problem when strict imports are wanted. See:
+    # * https://forums.swift.org/t/supporting-strict-imports/42472
+    #
+    # So provide here two behaviours:
+    # * By default, follow xcode and cocoapods and populate the umbrella header with the usual content
```
jerrymarino, line 468: @acecilia thank you for adding this! This is definitely less than ideal for many reasons as a default IMO. Longer term we can probably set it to false. The problem is that much iOS code was written to omit those imports - so that code still needs it. 

My main suggestion is to rename it from `legacy` to `generate_sdk_umbrella_header` or `generate_default_umbrella_header` , unless the "legacy" nomenclature is aligned with a cocoapods name

acecilia: Done ✌️  Renamed to `generate_default_umbrella_header`

acecilia: BTW, some extra context. Yes I had to go around the full codebase adding `import Foundation` and `import UIKit` where needed 💩 . But I did not do it manually. I wrote a script that:

1. Builds the codebase with bazel
2. Captures the output, filtering the lines saying ` error: cannot find`
3. Collect the files with above error
4. Adds `import Foundation` on the top of the file, or `import UIKit` if `import Foundation` is found already
5. Back to step 1

jerrymarino (approved):

gyfelton:
@acecilia Hi curious if you folks ever get any performance gain by turning it off?

acecilia:
👋 Mmm it has been a long time, but I do not recall observing any major benefit on app size nor build time

PR #353, opened 2021-11-17
arm64 on Bazel 4 - temporary stand-in for Bazel 5

jerrymarino:
Run the entire CI on arm64 + Bazel4. This includes a toolchain to build the
arm64 sim under Bazel 4. The problem is that Bazel should have a new CPU but
that needs much work in Bazel. The idea of this toolchain is to conditionally
compile x86_64 simulator as arm64.

Why add a build setting? There needs to be some additional context passed into
Bazel so it "knows" to override `x86_64`. Ideally it would use a build setting and
transition but there are bugs with that right now. 

These features could theoretically be auto-configured on an arm64 host but is a
flag for the moment.

Example usage:
```
common:bazel4_arm64_simulator_arm64_host --features apple.arm64_simulator_use_device_deps
common:bazel4_arm64_simulator_arm64_host --@build_bazel_rules_ios//:apple_host_cpu_override=arm64
common:bazel4_arm64_simulator_arm64_host --@build_bazel_rules_ios//:apple_simulator_cpu_override=arm64
```

```
common:bazel4_arm64_simulator_x86_64_host --features apple.arm64_simulator_use_device_deps
common:bazel4_arm64_simulator_x86_64_host --@build_bazel_rules_ios//:apple_simulator_cpu_override=arm64
```

I also renamed `apple.arm64_simulator_use_device_deps` to make it aligned with
the `arm64` name

Please see related patches in rules_swift for `bazel4`


.github/workflows/tests.yml
```diff
@@ -53,6 +53,35 @@ jobs:
...
+  build_bazel4_arm64_simulator:
+    name: Build arm64 Simulator
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Select Xcode 12.2
+        run: sudo xcode-select -s /Applications/Xcode_12.2.app
+      - name: Build and Test
+        run: |
+          bazelisk clean # Note: there is a problem with some reuse in rules_swift
```
thiagohmcruz, line 66: What problem? Is there an issue we can link to here?

.github/workflows/tests.yml
```diff
@@ -53,6 +53,35 @@ jobs:
         with:
           name: bazel-testlogs
           path: bazel-testlogs
+
+  build_bazel4_arm64_simulator:
+    name: Build arm64 Simulator
```
thiagohmcruz, line 58: Nit: So we don't have a `bazel4_arm64_simulator_arm64_host` test too because atm we can't run on such hosts on CI right? Should this name explicitly mention `x86_64` host?

tools/toolchains/bazel4_local_config_cc/BUILD
```diff
@@ -0,0 +1,126 @@
...
+        objcopy_files = ":empty",
+        strip_files = ":osx_tools_" + arch,
+        supports_param_files = 1,
+        toolchain_config = arch,
+        toolchain_identifier = arch,
+    )
+    for arch in OSX_TOOLS_ARCHS
+]
+
+print("ARCH", OSX_TOOLS_ARCHS)
```
thiagohmcruz, line 86: Is this a debug log?

tools/toolchains/bazel4_local_config_cc/BUILD
```diff
@@ -0,0 +1,126 @@
...
+[
+    cc_toolchain_config(
+        name = arch,
+        compiler = "compiler",
+        cpu = arch,
+        cxx_builtin_include_directories = [
+            "/Applications/",
+            "/Library/",
+            # Note: we may sub this out
+            "/Users/jmarino/Library/",
```
thiagohmcruz, line 97: - [ ] Remove before landing?

tools/toolchains/bazel4_local_config_cc/BUILD
```diff
@@ -0,0 +1,126 @@
...
+    as_files = ":empty",
+    compiler_files = ":empty",
+    dwp_files = ":empty",
+    linker_files = ":empty",
+    objcopy_files = ":empty",
+    strip_files = ":empty",
+    supports_param_files = 1,
+    toolchain_config = ":stub_armeabi-v7a",
+    # HACK - conflicts with other
+    #name = "cc-compiler-armeabi-v7ax",
```
thiagohmcruz, line 121: Remove?

tools/toolchains/bazel4_local_config_cc/armeabi_cc_toolchain_config.bzl
```diff
@@ -0,0 +1,82 @@
+# Copyright 2019 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
thiagohmcruz, line 3: I'll treat these files as "copy + pasted" and not review in depth, let me know if that makes sense or not. In case you made any changes I'd appreciate if you could point me to the relevant bits

thiagohmcruz (approved):
LGTM! A few questions inline 🚀

.github/workflows/tests.yml
jerrymarino: It seems to flake out when cross compiling, fixed in Bazel 5

tools/toolchains/bazel4_local_config_cc/armeabi_cc_toolchain_config.bzl
jerrymarino: There's 1-2 code line changed - you can grep `bazel4.` in the PR.

tools/toolchains/bazel4_local_config_cc/BUILD
jerrymarino: Was removed

.github/workflows/tests.yml
jerrymarino: Yeah exactly, you'd need an Apple Silicon CI!

tools/toolchains/bazel4_local_config_cc/BUILD
jerrymarino: I was fine with leave this one here, mainly because it is less changes

tools/toolchains/bazel4_local_config_cc/BUILD
jerrymarino: Good catch will do

PR #358, opened 2021-11-23
A rule to link with `-force_load` for direct `deps`

jerrymarino:
ld has different behavior when loading members of a static library VS objects
as far as visibility. Under `-dynamic`
- linked _swift object files_ can have public visibility
- symbols from _swift static libraries_ are omitted unless used, and not visible otherwise

By using `-force_load`, we can load static libraries in the attributes of an
application's direct depenencies. These args need go at the _front_ of the
linker invocation otherwise these arguments don't work with lds logic.

Why not put it into `rules_apple`? Ideally it could be, and perhaps consider a
PR to there .The underlying java rule, `AppleBinary.linkMultiArchBinary`
places `extraLinkopts` at the end of the linker invocation. At the time of
writing these args need to go into the current rule context where
`AppleBinary.linkMultiArchBinary` is called.

One use case of this is that iOS developers want to load above mentioned
symbols from applications. Another alternate could be to create an aspect, that
actually generates a different application and linker invocation instead of
force loading symbols. This could be more complicated from an integration
perspective so it isn't used. Another case is present in dynamic frameworks in
rules_apple ( https://github.com/bazelbuild/rules_apple/issues/332 )

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -2,6 +2,7 @@ load("@build_bazel_rules_apple//apple:ios.bzl", rules_apple_ios_application = "i
 load("//rules:library.bzl", "apple_library")
 load("//rules:plists.bzl", "info_plists_by_setting")
 load("//rules:import_middleman.bzl", "import_middleman")
+load("//rules:force_load_direct_deps.bzl", "force_load_direct_deps")
 
 # We need to try and partition out arguments for obj_library / swift_library
 # from ios_application since this creates source file libs internally.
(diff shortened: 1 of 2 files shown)


thiagohmcruz (approved):

rules/app.bzl
```diff
@@ -56,12 +57,14 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
     application_kwargs["launch_storyboard"] = application_kwargs.pop("launch_storyboard", library.launch_screen_storyboard_name)
     application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
 
-    import_middleman(name = name + ".import_middleman", deps = library.deps, tags = ["manual"])
+    force_load_direct_deps(name = name + ".__internal__.force_load_direct_deps", deps = library.deps, tags = ["manual"])
```
congt, line 60: Should we only force load swift_library?

jerrymarino: Nope we need to force load c++/objc too

Issue #361, opened 2021-11-23
Bazel auto-configured xcode_version is flaky: ERROR: /private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/BUILD:87:24: in cc_toolchain_config rule @local_config_cc//:darwin_x86_64: 


jerrymarino:
This issue is plaguing the CI and also local developers.  It calls `xcode-select -s` each time ( although doesn't have a developer directory ).

```
ERROR: /private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/BUILD:87:24: in cc_toolchain_config rule @local_config_cc//:darwin_x86_64: 
75
Traceback (most recent call last):
76
	File "/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/cc_toolchain_config.bzl", line 495, column 61, in _impl
77
		flags = _deterministic_libtool_flags(ctx) + [
78
	File "/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/cc_toolchain_config.bzl", line 53, column 38, in _deterministic_libtool_flags
79
		if _can_use_deterministic_libtool(ctx):
80
	File "/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/cc_toolchain_config.bzl", line 45, column 25, in _can_use_deterministic_libtool
81
		if _compare_versions(xcode_version, _SUPPORTS_DETERMINISTIC_MODE) >= 0:
82
	File "/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/local_config_cc/cc_toolchain_config.bzl", line 38, column 15, in _compare_versions
83
		return dv1.compare_to(apple_common.dotted_version(v2))
```

jerrymarino:
Here's a related Bazel issue: https://github.com/bazelbuild/bazel/issues/13509

jerrymarino:
Details in here on current status https://github.com/bazel-ios/rules_ios/pull/362 and upstream fix

PR #362, opened 2021-11-23
Measures to debug Bazel's Xcode ingestion flake

jerrymarino:
Fork Bazel's Xcode autoconfigure for usability improvements and debugging

Fork and improve Bazel's xcode configuration code. This is not a measure to fix
the problem, but to debug it and unblock CI. Github actions is incrediblly
flaky but Bazel should deal with a flaky host better and this is a weak point
(https://github.com/bazel-ios/rules_ios/issues/361)

Bazel puts this error in the build file which is a problem on the CI because it
swallows the error and it requires a user to clean --expunge. Bazel requires
the user to expunge because it thinks the repository was successfully written.
Probably a better pattern is to fail, but that won't shed light on the problem.

This is not just broken for CI but local developers. The plan is to use
rules_ios CI to reproduce this and gather more information then propose a fix
when we know more info.

Ammedum: I have confirmed with this PR, it's related to host over-subscribing
on GitHub Actions. This at least improves it because the user doesn't have to
expunge. PR to fix that to follow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Select Xcode 12.2
-        run: sudo xcode-select -s /Applications/Xcode_12.2.app
+      - name: Select Xcode
+        run: .github/workflows/xcode_select.sh
       - name: Build and Test
         run: |
           # Host config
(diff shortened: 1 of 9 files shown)


jerrymarino:
Taking5MinutesToFetchXcodeConfig

congt (approved):

thiagohmcruz (approved):

jerrymarino:
Will circle back on this next week, proposing to Bazel:
https://github.com/bazelbuild/bazel/pull/14328

PR #367, opened 2021-11-30
Add target_settings_by_config and pass config type in configs attribute

thiagohmcruz:
Change `configs` attribute to also requires the config type XcodeGen uses `debug/release`:

1. New `configs` API (from list to dict), **this is an API breaking change**:
```
Dictionary keyed at config name present in the .bazelrc file and values at one of these two config types: 'debug', 'release'

A Xcode build configuration will be created for each entry and a '--config=$CONFIGURATION' will
be appended to the underlying bazel invocation. Effectively allowing the configs in the .bazelrc file
to control how Xcode builds each build configuration.

If not present the 'Debug' and 'Release' Xcode build configurations will be created by default without
appending any additional bazel invocation flags.
```

2. `target_settings_by_config`:
```
Additional optinal dictionary with Xcode build settings to be added to all targets grouped by config (see 'configs' attribute).

Example:

target_settings_by_config = {
        "my_config": [
            "PRODUCT_BUNDLE_IDENTIFIER=com.company.app.my_config",
            "FOO=bar_1",
        ]
        "Release": [
            "PRODUCT_BUNDLE_IDENTIFIER=com.company.app",
            "FOO=bar_2",
        ],
}

Each config has to exist in 'configs' and be set to one of 'debug'/'release' so xcodegen knows about its existence when creating the target.

Also, note that bazel doesn't support attributes with strings as keys and values at dictionaries so the proposal here is
to pass the settings as an array of strings where each element has the format 'SETTING_NAME=VALUE'.

Read more:
- https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#settings
- https://docs.bazel.build/versions/main/skylark/lib/attr.html
```

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -697,6 +697,27 @@ env -u RUBYOPT -u RUBY_HOME -u GEM_HOME $BAZEL_BUILD_EXEC $BAZEL_BUILD_TARGET_LA
 $BAZEL_INSTALLER
 """
 
+def _set_target_settings_by_config(ctx, target_settings):
+    if len(ctx.attr.target_settings_by_config.keys()) == 0:
+        return target_settings
+
+    updated_target_settings = {}
+    updated_target_settings["base"] = target_settings
+    settings_by_config = target_settings.get("configs", {})
+
+    for (config, settings) in ctx.attr.target_settings_by_config.items():
+        if config not in settings_by_config.keys():
+            settings_by_config[config] = {}
+
+        for setting in settings:
+            k = setting.rsplit("=", 1)[0]
+            v = setting.rsplit("=", 1)[1]
+            settings_by_config[config][k] = v
+
+    updated_target_settings["configs"] = settings_by_config
+
+    return updated_target_settings
+
 def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_transitive_targets):
     """Helper method to generate dicts for targets and schemes inside Xcode context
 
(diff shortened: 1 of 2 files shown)


rules/xcodeproj.bzl
```diff
@@ -1142,6 +1167,49 @@ Tags for configuration:
 
         If not present the 'Debug' and 'Release' Xcode build configurations will be created by default without
         appending any additional bazel invocation flags.
+        """),
+        "build_type_by_config": attr.string_dict(default = {}, doc = """
```
jerrymarino, line 1171: One of the reasons behind the original config design API we wanted to pass an array of configs to Bazel was to prevent the xcodeproj from diverging from CLI. Shouldn't these be already available to the build system - specifically that Bazel will be setting these attributes?

The aspect would read in these values from `ios_application.bundle_id` for instance

jerrymarino (commented):
Minor question

rules/xcodeproj.bzl
thiagohmcruz: The value is available when the aspect is collecting info but that's only for the config active while the proj is being generated, meaning that one has to generate multiple projects one in each config flavour. 

With the setup I'm proposing here one can map configs to `debug/release` and set any flavour specific setting in `target_settings_by_config` (e.g. `PRODUCT_BUNDLE_IDENTIFIER` is relevant when deploying to device) so that building things on that config is just a matter of selecting a different Xcode build configuration on the GUI instead all using a single generated project

congt (approved):

PR #369, opened 2021-12-03
Update Xcode version to 12.5.1.

jgraves:
`macos-latest` runners have been updated with Xcode version 12.5.1.


thiagohmcruz:
Should we also update the generated files [here](https://github.com/bazel-ios/rules_ios/tree/master/data)?

.github/workflows/tests.yml
```diff
@@ -94,13 +94,13 @@ jobs:
...
-    name: .xcodeproj Tests on Xcode 12.2
+  xcodeproj_tests_xcode_12_5_1:
+    name: .xcodeproj Tests on Xcode 12.5.1
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Select Xcode 12.2
-        run: sudo xcode-select -s /Applications/Xcode_12.2.app
+      - name: Select Xcode 12.5.1
+        run: sudo xcode-select -s /Applications/Xcode_12.5.1.app
```
jerrymarino, line 103: It looks like this one got missed - consider to just update it to use `.xcode_select.sh`

jerrymarino (approved):
LGTM - after fixing @thiagohmcruz inline. I believe the repo has a command to update that:
```
bazel run data_generators/extract_xcspecs
```
Could add a doc for the next person 👍

jerrymarino:
@jgraves this new PR is hardcoding the version to remove flake - this adds further complexity to bump:
https://github.com/bazel-ios/rules_ios/commit/f742371c186da00a6e3a6718ac272919851907a7#diff-7fc57714ef13c3325ce2a1130202edced92fcccc0c6db34a72f7b57f60d552a3R40

In order to auto-generate this info, we can run the command:
```
cat $(bazel info execution_root)/../../external/local_config_xcode/BUILD
```

If you'd like to address all of this - that's fine, otherwise, back-port the changes that extend the timeout, and we can remove the code which hardcodes versions.

thiagohmcruz (approved):

BUILD.bazel
```diff
@@ -59,3 +59,23 @@ xcode_config(
...
+  version = '12.5.1.12E507',
+  aliases = ['12.5.1' ,'12' ,'12.5' ,'12.5.1.12E507'],
+  default_ios_sdk_version = '14.5',
+  default_tvos_sdk_version = '14.5',
+  default_macos_sdk_version = '11.3',
+  default_watchos_sdk_version = '7.4',
+)
+
+xcode_config(name = 'host_xcodes',
+  versions = [':version12_5_1_12E507'],
```
jerrymarino, line 74: I think this change is causing a problem with the existing `host_xcodes` -  `version12_5_1_12E507` should be added to `versions` in the existing `host_xcodes`

BUILD.bazel
```diff
@@ -59,3 +59,23 @@ xcode_config(
...
+  default = ':version12_5_1_12E507',
+)
+
+available_xcodes(name = 'host_available_xcodes',
```
jerrymarino, line 78: Same as above - this is conflicting with the above definition I believe

jerrymarino:
Thanks for the PR @jgraves !

jerrymarino:
Well according to the action this passed but status isn't updated:
https://github.com/bazel-ios/rules_ios/actions/runs/1575518515

PR #371, opened 2021-12-08
Import swiftsourceinfos

thii:


justinseanmartin:
The change itself seems reasonable. What functionality does this fix/change in Xcode? Thanks for the contribution!

thii:
From https://forums.swift.org/t/proposal-emitting-source-information-file-during-compilation/28794, IIUC they can provide diagnostics info to Xcode.

jerrymarino:
Thanks for the fix @thii 🙇

jerrymarino (approved):

PR #373, opened 2021-12-09
Avoid running xcode-locator which can be super slow (and can fail) if the machine has many Xcode versions

thii:
This fixes the flaky CI.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -33,6 +33,10 @@ build --features swift.index_while_building
 # Prevents leaking unexpected binaries via PATH to tests
 build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
+# Avoid running xcode-locator which can be super slow if the machine has many
+# Xcode versions
+build --xcode_version_config=//:host_xcodes
+
 # A toolchain to build the arm64 sim under Bazel 4. The problem is
 # that Bazel should have a new CPU but that needs much work in Bazel. The idea of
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -33,6 +33,10 @@ build --features swift.index_while_building
 # Prevents leaking unexpected binaries via PATH to tests
 build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
+# Avoid running xcode-locator which can be super slow if the machine has many
+# Xcode versions
+build --xcode_version_config=//:host_xcodes
```
thiagohmcruz, line 38: So IIUC this prevents the 
```
Error in fail: Generating xcode-locator-bin failed.  return code 256, stderr: Timed out, stdout: 
```
coming from the toolchain that includes `tools/toolchains/xcode_configure/xcode_configure.bzl`. @jerrymarino I see you added that correct? Any implications here?

It would be nice to prevent this annoying CI flakes indeed.

jerrymarino:
@thii thanks for the PR - this seems good for now! 

We have a fix for this upstream as well: [  bazelbuild/bazel#14328 ](https://github.com/bazelbuild/bazel/pull/14328)

jerrymarino:
I should have time at best early next week to fix this upstream in Bazel ( the timeout )

jerrymarino (approved):

PR #374, opened 2021-12-10
Only import iOS swiftmodules

thii:
If you have the same module also compiled for the host tools, the macOS
one could overwrite the iOS one and cause indexing error, for example:

```
module 'SwiftProtobuf' was created for incompatible target
x86_64-apple-macosx12.0
```


jerrymarino (approved):

jerrymarino:
Thanks for the PR @thii !

PR #375, opened 2021-12-10
Remap remote developer dir to local developer dir

thii:
Not sure if this is required or not, but since it's embedded in the
index store data, it should be better match with the local one.


tools/xcodeproj_shims/installers/_indexstore.sh
```diff
@@ -30,9 +30,13 @@ readonly xcode_module="$BUILT_PRODUCTS_DIR/\$1.swiftmodule/$ARCHS.swiftmodule"
 readonly bazel_external="$bazel_root/external"
 readonly xcode_external="$BAZEL_WORKSPACE_ROOT/bazel-$(basename "$SRCROOT")/external"
 
+readonly remote_developer_dir="^/.*/.+?\.app/Contents/Developer"
```
jerrymarino, line 33: @thii  this effectively maps _any_ of the Xcodes to the local one? It seems like a good thing to have. Maybe add a comment on this. I think it's fine to assume it's always `Contents/Developer` is the developer dir for now

jerrymarino (approved):

PR #385, opened 2021-12-17
Update to bazel5

amberdixon:


rules/library.bzl
```diff
@@ -971,7 +971,8 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             "@build_bazel_rules_ios//:virtualize_frameworks": framework_vfs_objc_copts,
             "//conditions:default": framework_vfs_objc_copts if enable_framework_vfs else [],
         })
-
+    if module_map:
```
amberdixon, line 974: We solve a problem with modulemaps by appending the module_map to the objc headers list. An alternative approach is here: https://github.com/bazel-ios/rules_ios/commit/7f4fe87a13edd72c9276bfbdc2d41155c464e571

However this approach may be preferable because bazel sees that modulemap is a dependency of the objc-library target.

tools/toolchains/bazel4_local_config_cc/BUILD
```diff
@@ -89,7 +89,7 @@ cc_toolchain_suite(
         compiler = "compiler",
         cpu = arch,
         cxx_builtin_include_directories = [
-            "/Applications/",
+            "/",
```
amberdixon, line 92: This is a temporary change to workaround a problem here (https://github.com/bazelbuild/bazel/issues/14346) with inclusion validation logic. We won't need this bazel5 toolchain stuff once we update to the next bazel5.0 release candidate. We need the following change (https://github.com/bazelbuild/bazel/commit/005361c895da334beb873901e93aff06d180256e)  which the bazel release team accidentally omitted from the rc3 cut. see https://github.com/bazelbuild/bazel/pull/14440#issuecomment-996394207 as well.

The current bazel5 release candidate is rc3, which can be determined by checking the bazel-discuss email list: https://groups.google.com/g/bazel-discuss/

.bazelversion
```diff
@@ -1 +1 @@
-4.1.0
+5.0.0rc3
```
jerrymarino, line 1: Great that this is all working now!

rules/framework.bzl
```diff
@@ -279,6 +279,8 @@ def _get_framework_files(ctx, deps):
                     header_in.append(hdr)
                     destination = paths.join(framework_dir, "Headers", hdr.basename)
                     header_out.append(destination)
+                elif hdr.path.endswith(".modulemap") and hdr.owner != dep.label:
```
jerrymarino, line 282: So we're guaranteed that the owner isn't going to be the `dep`? Would you mind dropping a comment about why this is so? If we make a dep of some modulemap rule is that going to work?

jerrymarino (approved):
@amberdixon LGTM and great this is working. We should have a followup to ensure the M1 CPU is working natively: removing the hack that we needed to make it work under Bazel 4.

rules/library.bzl
amberdixon: For this change, we return module_map target from the apple_library macro, since the objc-provider in bazel5 no longer supports it. See bazelbuild/bazel@31bec27#diff-eb9c179576423931e9db781a5acff548a82a4361a23ce1c1dc0c6b73b30d114bL35 and slack conversation here: https://ios-dev-at-scale.slack.com/archives/CBVEC7KRQ/p1637262740176000

jerrymarino:
@amberdixon are we able to merge this and the other other Bazel 5 fix? It looks like a problem with github actions - e.g. the check ran but it didn't ping back with status? https://github.com/bazel-ios/rules_ios/actions/runs/1664400771

thiagohmcruz (approved):

PR #390, opened 2022-01-11
Add bundle_name to list of iOS application kwargs

thiagohmcruz:
So consumers can customize this property

See: https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_application


justinseanmartin (approved):

gyfelton:
:TIL: why we need this list here

jerrymarino (approved):
LGTM!

PR #396, opened 2022-01-22
[Apple Silicon] Small improvments to stderr of update_static.sh

### LINKED ISSUE #394: Disable printing LC_BUILD_VERSION error for arm-to-sim codepath
When presented with a simulator arm binary this rule emits an error. Instead, just ditto the archive into the right place. 

```
arm64_to_sim/main.swift:141: Fatal error: This arm64 binary already contains an LC_BUILD_VERSION load command!
/private/var/tmp/_bazel_eamorde/979ccde8f2b7c523a76a6fe66ef56c48/execroot/register/external/build_bazel_rules_ios/tools/m1_utils/update_static.sh: line 69: 83755 Trace/BPT trap: 5       "$ARM64_TO_SIM_PATH" "$file" 11 11 --obj
```

jerrymarino:
This PR fixes 2 output messages that are bothersome in Xcode when
building many targets. It doesn't intend to change any functionality.

1. Fix "error" when running through Sim ARM64 objs through patch script
2. Fix ar output when used in "fat object file" archives

Fixes #394

### DIFF
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -74,7 +74,7 @@ patch() {
          chmod 777 "$file"
          # Test if the LC_BUILD_VERSION is 24, e.g. we already gave it an m1 arm64
          # object and exit
-         if otool -l "$file" | grep -q LC_BUILD_VERSION | tail -1; then
+         if otool -l "$file" | grep -q LC_BUILD_VERSION; then
             cp "$FWF" "$OF"
 
             exit 0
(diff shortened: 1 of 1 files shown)


tools/m1_utils/update_static.sh
```diff
@@ -12,11 +12,12 @@ unarchive() {
 
     # Assume that if unarchiving fails, it's already an object.
```
jerrymarino, line 13: Note: this comment is obsolete now as we've got fat object archive specifically handled below

thiagohmcruz (approved):

PR #397, opened 2022-01-24
Add objc_defines and swift_defines

thiagohmcruz:
Proposal to solve #400 by introducing `swift_defines` and `objc_defines`

**Why not stop supporting `defines`?**

Backwards compatibility in case consumers with more simple use cases are already setting this value

**Why not manipulate `defines` and extract swift/objc from it?**

The `defines` attribute is already in use for a while and it's being propagated "as is" so consumer could be setting `select()` statements in this attribute to handle different settings, meaning we can't assume this attribute contains a list that can be manipulated.

**Why add this as an explicit parameter in `apple_library`?**

It's a good way to ensure there's documentation for these attributes it's clear to consumers what is expected in each. Open to suggestions though.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -407,7 +407,7 @@ def _find_imported_framework_name(outputs):
         return fw_name
     return None
 
-def apple_library(name, library_tools = {}, export_private_headers = True, namespace_is_module_name = True, default_xcconfig_name = None, xcconfig = {}, xcconfig_by_build_setting = {}, **kwargs):
+def apple_library(name, library_tools = {}, export_private_headers = True, namespace_is_module_name = True, default_xcconfig_name = None, xcconfig = {}, xcconfig_by_build_setting = {}, objc_defines = [], swift_defines = [], **kwargs):
     """Create libraries for native source code on Apple platforms.
 
     Automatically handles mixed-source libraries and comes with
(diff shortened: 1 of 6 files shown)


thiagohmcruz:
- [ ] Add some tests for this

rules/library.bzl
```diff
@@ -991,6 +996,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         pch = pch,
         data = [] if swift_sources else [module_data],
         tags = tags_manual,
+        defines = defines + objc_defines,
```
gyfelton, line 999: If `foo=bar` is declared in `defines` and `foor=bar1` in `objc_defines`, do you know whether the compiler will take the later or will it spit error? 🙏

gyfelton (approved):
just one question

rules/library.bzl
thiagohmcruz: You can quickly test that in the [dummy project](https://github.com/thiagohmcruz/swift-objc-libraries-defines-issue) I linked in the issue. It would just trigger a warning:
```sh
<command line>:4:9: warning: 'foo' macro redefined [-Wmacro-redefined]
#define foo bar1
        ^
<command line>:3:9: note: previous definition is here
#define foo bar
```
and the context would define what value is used

jerrymarino (approved):
The backwards compat here is interesting, specifically because it's hard to know what to expect. For now, consider adding a warning, and possibly longer term we can add an `incompatible` layer like Bazel has to get rid of deprecated behavior

PR #399, opened 2022-01-25
Support both relative and absolute paths in Xcode output processor

### LINKED ISSUE #360: IDE error integrations do not work with Xcode 13+
Xcode 13/Swift 5.5 introduced a regression where, when VFS overlays are used, error outputs have absolute paths instead of relative paths. This breaks the output processing done via regex in `output-processor.rb`

The end result is that build errors do not show up in the Xcode IDE text editor inline with the location of the error.

erikkerber:
I noticed in our project that in some cases, modules would output relative while others output absolute. This change allows the output processor to support both.

It also allows rules_ios to be backwards compatible to Xcode 12, for what that's worth.

Re-resolves #360

### DIFF
--- a/tools/xcodeproj_shims/output-processor.rb
+++ b/tools/xcodeproj_shims/output-processor.rb
@@ -13,6 +13,8 @@ class BazelOutputLine
   def self.from(line)
     if (message_line = CompilerMessageLine.create_from(line))
       message_line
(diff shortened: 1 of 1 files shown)


jerrymarino (approved):
Thanks or the PR @erikkerber  ✅

PR #404, opened 2022-01-29
Quote ARCH variable in update_static.sh

sethfri:
Shellcheck is a static analysis tool that highlights syntactic issues
that lead to common shell bugs. It highlighted that `ARCH` is not quoted
in this script, and when I quote it, the script does not work correctly.

Opening this as a PR at @jerrymarino's request so that he can see the
change.

### DIFF
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -47,7 +47,7 @@ patch() {
      ARCHS=()
      mkdir -p "$(dirname $OF)"
      while read ARCH; do
-          ARCHS+=($ARCH)
+          ARCHS+=("$ARCH")
      done < <(lipo -archs "$FWF")
      if test "${#ARCHS[@]}" -gt 1; then 
          lipo "$FWF" -thin arm64 -output "OF.ar" || cp "$FWF" "OF.ar"


tools/m1_utils/update_static.sh
```diff
@@ -47,7 +47,7 @@ patch() {
      ARCHS=()
      mkdir -p "$(dirname $OF)"
      while read ARCH; do
-          ARCHS+=($ARCH)
+          ARCHS+=("$ARCH")
```
jerrymarino, line 50: @sethfri this change would break the program because it currently it relies on wordsplitting in bash to split the variable into an array.

I think it's safe to read to an array too - does that pass the shellcheck config?
```
read -a ARCHS < <(lipo -archs "$FWF")
```

jerrymarino (commented):
Thanks for the PR @sethfri! We don't use shellcheck on this repo but would happily accept this tweak to pass it 👍

tools/m1_utils/update_static.sh
jerrymarino: Edit: _to fix that, you can replace the entire while loop with the above statement_ ^

PR #405, opened 2022-01-31
Integrate XCTest stdout workaround for Xcode 12 - pass 1

jerrymarino:
Based on https://github.com/ios-bazel-users/ios-bazel-users/blob/d99366f5bb0b8e945a1a758f832e1b65fb42044f/jump_to_test_failure.md

Enables jumping to test failures when using Bazel Xcode projects in
Xcode 12+, by expanding relative file URLs to be relative to the
SOURCE_ROOT environment variable, which is already set for all tests /
schemes.

This works OK for Xcode 12 and we'll need to patch this for Xcode 13.
This should be handled as part of rules_ios, instead of by the users of
it.

### DIFF
--- a/tools/xctest_absolute_source_locations/xctest_absolute_source_locations.m
+++ b/tools/xctest_absolute_source_locations/xctest_absolute_source_locations.m
@@ -1,0 +1,68 @@
+// Based on https://github.com/ios-bazel-users/ios-bazel-users/blob/d99366f5bb0b8e945a1a758f832e1b65fb42044f/jump_to_test_failure.md
+// Enables jumping to test failures when using Bazel Xcode projects in Xcode 12+,
+// by expanding relative file URLs to be relative to the
+// SOURCE_ROOT environment variable, which is already set for all tests / schemes.
+
+#import <objc/message.h>
+#import <objc/runtime.h>
+
+#import <Foundation/Foundation.h>
+#import <XCTest/XCTest.h>
+
+static NSURL *remapURL(NSURL *fileURL, NSString *srcroot)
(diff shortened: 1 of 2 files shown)


thiagohmcruz (approved):

tools/xctest_absolute_source_locations/xctest_absolute_source_locations.m
```diff
@@ -0,0 +1,68 @@
...
+
+@interface XCTestAbsoluteSourceLocationsSwizzlingLoader : NSObject
+@end
+
+
+@implementation XCTestAbsoluteSourceLocationsSwizzlingLoader
+
++ (void)load
+{
+    NSString *srcroot = [[[NSProcessInfo processInfo] environment] objectForKey:@"SOURCE_ROOT"];
```
jerrymarino, line 48: Most of this will go away

tools/xctest_absolute_source_locations/xctest_absolute_source_locations.m
```diff
@@ -0,0 +1,68 @@
...
+static NSURL *remapURL(NSURL *fileURL, NSString *srcroot)
+{
+    if ([fileURL.path hasPrefix:srcroot]) {
+        return fileURL;
+    }
+
+    return [NSURL fileURLWithPath:[NSString stringWithFormat:@"%@/%@", srcroot, fileURL.relativePath]];
+}
+
+static NSURL *parentGitRepoOfURL(NSURL *searchURL)
```
jerrymarino, line 21: same as below

PR #414, opened 2022-02-24
Integrate dynamic linking / embedding of apple_framework

jerrymarino:
This PR adds the ability to easily link an `apple_framework` dynamically
and embed it into consumers with a single flag `link_dynamic`.

It also introduces a new provider to AvoidDeps. In dynamic linking, we
need to avoid linking deps that are in dep frameworks. This PR
propagates `AvoidDeps` upwards in the build graph to leverage in
linking. It currently uses `rules_apple` partials to help bundle: e.g.
handle resources, debug etc and interface with other bundling code.

Additionally, I have added a build test example that splits an app into
to 2 framework binaries, and resources are correctly paced.

```
app -> a.framework[dynamic] -> b.framework[dynamic] -> c
```

I considered using `ios_framework` but there were too many problems:
First it did't work with our plist rules. Then, it required me to add
more rules and macros, then replicate the framework structure to
consumers: a framework for `apple_framework` and a framework for
`ios_framework`, in `deps` and `frameworks`. That also doesn't integrate
well with custom plist rules. Additionally, I'd like to determine this
information at analysis time longer term - say auto-configure some
targets to `link_dynamic`. This becomes easier when it's determined in
the analysis phase. Finally, some `apple_frameworks` should always
`link_dynamic` - and this makes it easy to ensure that happens.

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -33,6 +33,7 @@ _IOS_APPLICATION_KWARGS = [
     "minimum_deployment_os_version",
     "ipa_post_processor",
     "include_symbols_in_bundle",
+    "frameworks",
 ]
 
 def ios_application(name, apple_library = apple_library, infoplists_by_build_setting = {}, **kwargs):
@@ -59,9 +60,9 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
     application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
 
     force_load_name = name + ".force_load_direct_deps"
-    force_load_direct_deps(name = force_load_name, deps = library.lib_names, tags = ["manual"])
-    default_deps = [force_load_name] + library.lib_names
+    force_load_direct_deps(name = force_load_name, deps = kwargs.get("deps"), tags = ["manual"])
 
+    default_deps = [force_load_name] + library.lib_names
     import_middleman(name = name + ".import_middleman", deps = default_deps, tags = ["manual"])
     rules_apple_ios_application(
         name = name,
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -1,24 +1,34 @@
 """Framework rules"""
 
-load("@build_bazel_rules_apple//apple/internal:apple_product_type.bzl", "apple_product_type")
-load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "AppleSupportToolchainInfo")
-load("@build_bazel_rules_apple//apple/internal:platform_support.bzl", "platform_support")
-load("@build_bazel_rules_apple//apple/internal:resource_actions.bzl", "resource_actions")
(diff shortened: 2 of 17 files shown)


rules/app.bzl
```diff
@@ -59,9 +60,9 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
     application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
 
     force_load_name = name + ".force_load_direct_deps"
-    force_load_direct_deps(name = force_load_name, deps = library.lib_names, tags = ["manual"])
-    default_deps = [force_load_name] + library.lib_names
+    force_load_direct_deps(name = force_load_name, deps = kwargs.get("deps"), tags = ["manual"])
```
congt, line 63: Do we still need to add library.lib_names in the deps?

rules/framework.bzl
```diff
@@ -515,6 +525,252 @@ def _attrs_for_split_slice(attrs_by_split_slices, split_slice_key):
     else:
         return attrs_by_split_slices[split_slice_key]
 
+def _bundle_dynamic_framework(ctx, avoid_deps):
+    """Packages this as dynamic framework
```
congt, line 529: Ah, so the apple_framework_packaging rule essentially reimplements ios_framework when link_dynamic is true.

congt (approved):

thiagohmcruz (approved):

rules/framework.bzl
jerrymarino: Yea, for now it calls into `rules_apple` bundling logic, which is somewhat similar to `ios_framework`. This is somewhat an implementation detail ATM. 

I'm a bit torn between doing this and using `ios_framework` - I like that this doesn't expand build file complexity or require us to make more macros. Possibly longer term we could propose a public API in rules apple:

```
bundling_support.bundle_dynamic_framework(binary=)
```

rules/app.bzl
jerrymarino: No it needs the actual deps vs the `lib_names` in order to compute force loading given dynamic frameworks

rules/framework.bzl
```diff
@@ -515,6 +525,252 @@ def _attrs_for_split_slice(attrs_by_split_slices, split_slice_key):
...
+    )
+
+def _bundle_static_framework(ctx, outputs):
+    """Returns bundle info for a static framework commonly used intra-build"""
+    infoplist = _merge_root_infoplists(ctx)
+
+    current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
+
+    # Static packaging - archives are passed from library deps
+    # Merges Info.plists and converts them into binary
```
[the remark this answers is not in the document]
jerrymarino: Ah good point, looks like that got refactored a bit

PR #415, opened 2022-02-24
Fix unable to append settings that has new line character in it

gyfelton:
also address shellcheck linter warnings


thiagohmcruz (approved):
Thx for fixing that!

congt (approved):

PR #419, opened 2022-03-03
[apple.arm64_simulator_use_device_deps] 13.1-13.2.1 LLDB fix

jerrymarino:
Xcode 13.1-13.2.1 workaround ( no issue for 12.5.1 ) Strip iOS
device outputs file to mitigate LLDB default SDK selection from
the DWARF entry - DW_AT_APPLE_sdk. No harm since many third party
dist binaries ship with release + DSYM. Without this, Swift
debugging is broken. To fix without stripping, implement a macho
rewriter that removes or updates DW_AT_APPLE_sdk. Upstream LLVM
has redesigned the logic and it's fixed there once devs have that
patch

### DIFF
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -77,7 +77,7 @@ patch() {
             cp "$FWF" "$OF"
 
             exit 0
-         elif ! "$ARM64_TO_SIM_PATH" "$file" 11 11 --obj; then 
+         elif ! "$ARM64_TO_SIM_PATH" "$file" 11 11 --obj; then
              # TODO: Versions should be input from the build system - hardcoded to 11
              cp "$FWF" "$OF"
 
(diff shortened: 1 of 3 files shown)


tools/m1_utils/update_static.sh
```diff
@@ -87,8 +87,22 @@ patch() {
...
+
+             # Xcode 13.1-13.2.1 workaround ( no issue for 12.5.1 ) Strip iOS
+             # device outputs file to mitigate LLDB default SDK selection from
+             # the DWARF entry - DW_AT_APPLE_sdk. No harm since many third party
+             # dist binaries ship with release + DSYM. Without this, Swift
+             # debugging is broken. To fix without stripping, implement a macho
+             # rewriter that removes or updates DW_AT_APPLE_sdk. Upstream LLVM
+             # has redesigned the logic and it's fixed there once devs have that
+             # patch
+             strip -S $OF
```
thiagohmcruz, line 99: 🤯

thiagohmcruz (approved):

tools/m1_utils/update_static.sh
gyfelton: am I right that one day we can revert this patch once upstream LLVM is fixed?

Issue #421, opened 2022-03-08
`index-import` locking up M1 after a big build / big rebase


jerrymarino:
`index-import` is locking up machine after a big rebase. Longer term we'll remove `index-import` because it has to process a large amount of data locally in this situation. Please mitigate `index-import` from locking it up.

Currently we run with `parallel-strides=32` - Consider setting this and/or reducing QOS:
https://github.com/MobileNativeFoundation/index-import/blob/main/index-import.cpp#L41

Issue #227, opened 2021-03-05
Index while building V2 - sans index-import for local exeuction


jerrymarino:
One major shortcut exists locally to remove index-import: Xcode is fixed to read directly from Bazel's global index store. Then, instead of writing to global index and copying to bazel-out for remote caching, Xcode should can read right from Bazels' global index. A "out of band" sentenial value can noop the code path for bazel-out. (out of band means abstracted from Bazel)

See more info in the design doc

jerrymarino:
cc @thiagohmcruz JFYI we'll also end up resolving this one we nix index-import from the post build script: https://github.com/bazel-ios/rules_ios/issues/497

PR #422, opened 2022-03-08
Set index-import parallel-stride to n CPUs

thiagohmcruz:
Resolves https://github.com/bazel-ios/rules_ios/issues/421 

* Invoke `index-import` with `parallel-stride` using different values for Intel vs M1
* Kill `index-import` process early to make sure there's only one instance running at a time
* Ensure only valid stores with both `units` and `records` directories are propagated ([to prevent this failure mode](https://github.dev/MobileNativeFoundation/index-import/blob/7c67cc40fe691f5d78ffa1f6e71cf32d59cf56fc/index-import.cpp#L386-L391))
* Also adds more information to the logs to help us troubleshoot in the future (fixing one bug that was prevent logs to be echo-ed if no stores were found)

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
@@ -36,14 +36,25 @@ readonly local_developer_dir="$DEVELOPER_DIR"
 # Uses all available cores for intel and a fraction of the performance cores on M1
 PARALLEL_STRIDE=$(sysctl -n hw.physicalcpu)
 
-# M1s have more performance cores than efficiency so dividing by 2 here should take
+# M1 machines have more performance cores than efficiency cores so dividing by 2 here should take
 # a good percentage of the performance cores to do this work
 if [[ $(arch) == 'arm64' ]]; then
   PARALLEL_STRIDE=$(expr $PARALLEL_STRIDE / 2)
 fi
 
+# Kill existing index-import process if it's running
+# in some cases this process is taking too much resources on the machine so
+# givin priority only to the latest run
+DERIVED_DATA="$BUILD_DIR"/../../..
+if [[ -f $DERIVED_DATA/index-import.pid ]]; then
+  PID=$(cat $DERIVED_DATA/index-import.pid)
+  if ps -p $PID > /dev/null; then
+    kill -10 $PID
+  fi
+fi
+
 $BAZEL_INSTALLERS_DIR/index-import \
-    -parallel-stride $PARALLEL_STRIDE
+    -parallel-stride $PARALLEL_STRIDE \
     -incremental \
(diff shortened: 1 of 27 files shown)


tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
```diff
@@ -35,6 +35,7 @@ readonly local_developer_dir="$DEVELOPER_DIR"
 
 
 $BAZEL_INSTALLERS_DIR/index-import \
+    -parallel-stride $(sysctl -n hw.physicalcpu)
```
jerrymarino, line 38: We might need a different code path for M1 VS intel here - becuase M1 has the efficeny cores. We can probably just take this number and divide it by 2

jerrymarino:
The one other thing I was thinking about, which is riddled with problems is to terminate existing `index-import` processes because today, these processes back up. It might be better than letting them run, perhaps add something like this:
```
! [[ -f $DERIVED_DATA/index-import.pid ]] && kill -10 $DERIVED_DATA/index-import.pid
index-import ...
echo $! > $DERIVED_DATA/index-import.pid
```

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
thiagohmcruz: Makes sense, will push the changes

jgraves (approved):
This looks good to me. As we discussed this morning, I think we should increase the amount of information we're logging (especially when we start killing processes, etc.), but that's more of a general discussion and not really a blocker for landing this PR.

tools/xcodeproj_shims/installers/indexstores.sh
```diff
@@ -6,12 +6,18 @@ set -euo pipefail
 
 echo "Start remapping index files at `date`"
 
-FOUND_INDEXSTORES=`grep -o 'command_line: "\(.*\.indexstore\)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | sed 's/.*command_line: "//' | uniq`
+FOUND_INDEXSTORES=`grep -o 'command_line: "\(.*\.indexstore\)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | sed 's/.*command_line: "//' | uniq` || true
```
jerrymarino, line 9: QQ: If the command to `FIND_INDEX_STORES` fails what will it loop over?

jerrymarino:
Thanks @thiagohmcruz for adding that line, yeah there is really no great solution to fix this other than what we've got planned in #227

tools/xcodeproj_shims/installers/indexstores.sh
jerrymarino: Well the effect is that it will terminate existing `index-import` processes - this is a big improvement over locking up the machine 👍 !

jerrymarino (approved):

PR #424, opened 2022-03-11
Add a fix for more modern rules_swift

ob:
This was motivated by a failure in producing the `-Swift.h` header while compiling the `//raven:Base` target of https://github.com/andyyhope/raven-rulesios. @thii noted that said repo was using a more modern `rules_swift` than what is in the dependencies of this repo. We weren't checking `CcInfo` for headers. I think this is backwards compatible so it shouldn't break anything and allow people to use more modern `rules_swift`. Kudos to @thii for fixing it... I'm just committing the patch.


rules/framework.bzl
```diff
@@ -270,9 +270,17 @@ def _get_framework_files(ctx, deps):
                 destination = paths.join(framework_dir, "PrivateHeaders", hdr.basename)
                 private_header_out.append(destination)
 
+        has_header = False
+        if CcInfo in dep:
+            for hdr in dep[CcInfo].compilation_context.direct_headers:
```
jerrymarino, line 275: This block is pretty similar below - consider we have a way to extract all hdrs from providers, then looping that 
```
for hdr in _get_dep_hdrs(dep):
```

jerrymarino (approved):
Thanks for the update @ob - added a minor inline comment. We can also bump to HEAD of rules_swift here in a followup to ensure it's working, there's some other changes / features we might want to start using here anyhow!

ob:
Yeah, I was doing the safe thing and not bumping `rules_swift` but I can try it.

rules/framework.bzl
ob: I was unsure about the `.modulemap`... I _think_ that got cleaned up in `rules_swift` as well no?

thiagohmcruz (approved):

rules/framework.bzl
jerrymarino: Under `rules_ios` we now propagate that via hdrs, given that, I wonder if we also need this from `CcInfo` if it's not picked up anywhere else. Maybe related to why the CI is failing?

jerrymarino:
```
bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-a7ed3b5187e2/bin/tests/ios/in-tree-vendor-prebuilt-deps/dynamic/app/FW/FW.framework/Headers/FW-umbrella.h:20:1: error: umbrella header for module 'FW' does not include header 'FW_Private.h' [-Werror,-Wincomplete-umbrella]
tests/ios/in-tree-vendor-prebuilt-deps/dynamic/app/App/main.m:1:9: fatal error: could not build module 'FW'
@import FW;
 ~~~~~~~^~
```

ob: good catch, that was exactly it.

thii:
Thanks @ob!

rules/framework.bzl
```diff
@@ -270,35 +270,37 @@ def _get_framework_files(ctx, deps):
...
-                continue
-
-            if SwiftInfo in dep and dep[SwiftInfo].direct_modules:
-                # apple_common.Objc.direct_module_maps is broken coming from swift_library
-                # (it contains one level of transitive module maps), so ignore SwiftInfo from swift_library,
-                # since it doesn't have a module_map field anyway
-                continue
+        providers = [CcInfo, apple_common.Objc]
+        for provider in providers:
+            if provider in dep:
```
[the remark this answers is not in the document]
ob: 🤦 I shouldn't be allowed near a computer...

PR #425, opened 2022-03-11
Generate lldbinits from installer for LLDB test

jerrymarino:
Feeds the LLDB init test from the rules into the xcodeproj

It sxport Xcode build settings JSON to an env file, sources it and then
runs `lldb-settings.sh` Gist is to dump showBuildSettings to replicate
xcode runscript actions

This is a quick hack for testing, and the env file is not reproducible
because the absolute paths.  Theoretically the LLDBInit should be
relative, the LLDBInit is less necessary it's current form with "virtual
frameworks".

Additionally silence output that is large on a big app

### DIFF
--- a/rules/test/lldb/lldb_sim_runner.py
+++ b/rules/test/lldb/lldb_sim_runner.py
@@ -94,11 +94,10 @@ def pack_tree_artifact(ios_application_output_path, app_name):
                     "--delete",
                     "--checksum",
                     "--chmod=u+w",
-                    "--verbose",
                     os.path.realpath(ios_application_output_path),
                     archive_app_path
                     ], check=True)
-    subprocess.run(["zip", "-r",  ipa_path, "Payload", ],
+    subprocess.run(["zip", "-rq", ipa_path, "Payload", ],
                    check=True, cwd=archive_root)
 
     return ipa_path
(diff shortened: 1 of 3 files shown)


tests/ios/lldb/app/BUILD.bazel
```diff
@@ -1,6 +1,7 @@
 load("//rules/test/lldb:lldb_test.bzl", "ios_lldb_breakpoint_command_test", "ios_lldb_breakpoint_po_test")
 load("//rules:xcodeproj.bzl", "xcodeproj")
 load("//rules:app.bzl", "ios_application")
+load("@build_bazel_rules_ios//rules:xcodeproj.bzl", "xcodeproj_lldbinit")
```
jerrymarino, line 4: Fix import convention

rules/test/lldb/lldb_sim_runner.py
```diff
@@ -94,11 +94,10 @@ def pack_tree_artifact(ios_application_output_path, app_name):
                     "--delete",
                     "--checksum",
                     "--chmod=u+w",
-                    "--verbose",
                     os.path.realpath(ios_application_output_path),
                     archive_app_path
                     ], check=True)
-    subprocess.run(["zip", "-r",  ipa_path, "Payload", ],
+    subprocess.run(["zip", "-rq", ipa_path, "Payload", ],
```
thiagohmcruz, line 100: This is just to intentionally make logs less verbose right? Just wanted to confirm if the intend is to check `-q` in

thiagohmcruz (approved):
LGTM!

rules/test/lldb/lldb_sim_runner.py
jerrymarino: Yep, this one prints out a lot of output ATM

PR #427, opened 2022-03-14
Bump rules_swift and rules_apple

ob:
* Bump `rules_swift` and `rules_apple` to latest.
* Upgrade `buildifier`
* Remove the dependency on `rules_cc`
* Fix #424 which turned out to be incomplete.


tinder-maxwellelliott (approved):

jerrymarino:
Thanks for the bump OB! Left a comment on the other PR regarding headers - possibly related 👁️

ob:
I'm thinking that keeping backward compatibility isn't a good idea... maybe we should just suck it up and move to `CcInfo`. Thoughts?

rules/framework.bzl
```diff
@@ -275,7 +267,7 @@ def _get_framework_files(ctx, deps):
             if provider in dep:
                 # collect headers
                 has_header = False
-                for hdr in dep[apple_common.Objc].direct_headers:
+                for hdr in _get_direct_headers(provider, dep):
```
jerrymarino, line 270: Is it possible the semantic of these providers has changed, but the API hasn't? Maybe try something like this:
```
if _detect_old_provider(provider):
     hdrs = _get_hdrs_old()
else
     hdrs = get_hdrs_new()

.. then process hdrs below in the same routine.
```

I can take look at this as well - it should build on a recent cut of rules swift, maybe worth breaking backwards compat if it's too hard

rules/library.bzl
```diff
@@ -156,7 +156,7 @@ def _write_umbrella_header(
 
 """
 
-    for header in public_headers:
+    for header in public_headers + private_headers:
```
ob, line 159: I'm not sure about this... should the umbrella header include the private headers?

ob: Without this the error is 

```
While building module 'FW' imported from /private/var/tmp/_bazel_ob/c86c2b68df7abbb07c0dd5103a56923b/sandbox/darwin-sandbox/43/execroot/build_bazel_rules_ios/tests/ios/app/App/main.m:1:
In file included from <module-includes>:1:
/private/var/tmp/_bazel_ob/c86c2b68df7abbb07c0dd5103a56923b/sandbox/darwin-sandbox/43/execroot/build_bazel_rules_ios/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-a7ed3b5187e2/bin/tests/ios/app/FW-modulemap/FW-umbrella.h:20:1: error: umbrella header for module 'FW' does not include header 'FW_Private.h' [-Werror,-Wincomplete-umbrella]
/private/var/tmp/_bazel_ob/c86c2b68df7abbb07c0dd5103a56923b/sandbox/darwin-sandbox/43/execroot/build_bazel_rules_ios/tests/ios/app/App/main.m:1:9: fatal error: could not build module 'FW'
@import FW;
 ~~~~~~~^~
2 errors generated.
```

ob: Looking a bit deeper I think this diff is wrong, the problem is that the private headers are being included in the `Headers` section of the VFS overlay...

jerrymarino: @ob yeah that problem is a manifestation of plugging in private headers to the interface where they shouldn't be! This PR seems to pass CI now BTW - maybe just fixing docs?

ob: I didn't have time to look at this today but I want to figure out why the VFS overlay is wrong.

ob: Added the real fix... now only the docs are failing.

rules/framework.bzl
ob: Found the problem, it was our use of the `PrivateHeadersInfo` provider.

ob:
@jerrymarino @tinder-maxwellelliott I think this is ready now... would appreciate another quick look.

jerrymarino:
@ob thanks for wrapping this up, LGTM overall 👍

PR #428, opened 2022-03-17
Add xcodeproj 'provide_build_settings_from_target_for_pre_post_actions'

larrytaowang:
# Summary
This PR add an attribute "provide_build_settings_from_target_for_pre_post_actions" for xcodeproj, so pre/post action scrips can access the build settings env.

run `bazel run tests/ios/xcodeproj:Test-Provide-Build-Settings-From-Target-For-Pre-Post-Actions` to generate the test files.

# Test Done
The pre/post build action in generated Xcode project has the target

Screen Shot 2022-03-16 at 5 00 35 PM


jerrymarino (approved):
Thanks for the PR. Only suggestion is to condense this to another test case instead of having its own proj - mainly to reduce the number of tests projs we have to generate. Otherwise, LGTM

PR #431, opened 2022-03-18
Dynamic framework test-usage fixes and tests

jerrymarino:
Filter redundant dylib deps in test hosts and tests. In some transitive
deps situations it would double link - provide a `dep_middleman` to
gauretee they are deduped on `AvoidDepsInfo`. It uses this middleman to
control linking without forking the rule. Only force load deps when
dynamic linking - that's handed upstream elsewhere.

Additonally fix a bug with packing transitive resources and add test
coverage for misc transitive framework and resource packing. This issue
has an `output_test` rule to check that outputs are there.


rules/test/output/output_test.bzl
```diff
@@ -0,0 +1,35 @@
...
+        template = ctx.file.output_checker,
+        is_executable = True,
+        substitutions = {
+            "%TARGET%": target_file.short_path,
+            "%EXPECT%": "( " + " ".join(ctx.attr.expect) + " )",
+        },
+    )
+    return [DefaultInfo(runfiles = ctx.runfiles(files = ctx.attr.target[DefaultInfo].files.to_list()))]
+
+output_test = rule(
```
thiagohmcruz, line 23: 🎉  thx for adding this!

rules/test.bzl
```diff
@@ -57,20 +57,25 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
...
         host_args = []
+    library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": ios_test_kwargs.get("minimum_os_version")}, **kwargs)
+
+    # Setup framework middlemen - need to process deps and libs
+    fw_name = name + ".framework_middleman"
+    framework_middleman(name = fw_name, framework_deps = kwargs.get("deps", []) + library.lib_names, tags = ["manual"])
+    frameworks = [fw_name] + ios_test_kwargs.pop("frameworks", [])
+
+    dep_name = name + ".dep_middleman"
+    dep_middleman(name = dep_name, deps = kwargs.get("deps", []) + library.lib_names, tags = ["manual"], test_deps = host_args)
```
thiagohmcruz, line 73: Nit: this is very similar between test/framework/app, consider extracting to a single macro and consuming in all places

thiagohmcruz (approved):
LGTM! Thx for adding the tests too!

rules/test.bzl
jerrymarino: ➕  the main caveat is that we currently feed it `test_deps` to govern the bundling of a test bundle. I've been considering this as well: making an `apple_bundling_rule` factory where we can build out rules or macros for that. This might be something for a future effort where we add support for all the rules like extensions, widgets, etc.

thiagohmcruz: SGTM!

PR #438, opened 2022-03-24
Add xcodeproj documentation

luispadron:
Noticed we were missing the `xcodeproj` documentation which made me miss this rule. The doc was generated with `bazel build //docs:xcodeproj-docs` after adding `xcodeproj` to the `_RULES` define.

There are some missing docs for some of the parameters but I don't feel like I have enough context yet to write them, but feel free to suggest them in the comments and I can get them in this PR

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -14,6 +14,7 @@ _RULES = [
     "substitute_build_settings",
     "test",
     "transition_support",
+    "xcodeproj",
 ]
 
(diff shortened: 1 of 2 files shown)


jerrymarino:
cc @thiagohmcruz do you think we should have a doc for this while the rule is in flux?

luispadron:
Going to close this since #443 changes the API, we can revisit when the API is stable.

PR #439, opened 2022-03-24
Scope an imported xcframework's modulemap to Modules/

jerrymarino:
Prevously it pulled the first modulemap which may not match. This
specifies it down to modules to avoid mixing up nested module maps.

It doesn't currently traverse nested frameworks in Frameworks, however
they will still link. Ideally it will loop Frameworks/ but that is a
larger change

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -277,12 +277,13 @@ def _xcframework_slice(*, xcframework_name, identifier, platform, platform_varia
             xcframework_name,
         ))
 
-    # Collecting interface files for VFS
+    # Collecting interface files for VFS. This doesn't currently traverse nested
+    # frameworks in Frameworks, however they will still link.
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):

congt (approved):

luispadron:
Do we still want this @jerrymarino?

jerrymarino:
@luispadron do you have other issues that are fixed by this one? After writing it out, I was thinking we will need refactor how it works to support a few different use cases. However, that implicates a larger fundamental change. I'm not opposed to merging this, but there is a rabbit hole of issues depending what you're after.

luispadron:
@jerrymarino Nope! This was just one of the PRs I noticed was ready to merge but hadn't been. Was just curious, thanks for explaining!

PR #440, opened 2022-03-29
remove duplicated files in resource_bundles

taowang487:
After this rules-apple PR [Use clonefile directly on macOS](https://github.com/bazelbuild/rules_apple/commit/5401fe26c813c0ec0ae8b0e0a00759215c90be43), our app failed to build because the file already exists when creating a bundle, so `clonefile`.

Trying to fix it in rules-apple, [PR: in bundletool, remove destination if needed before clonefile](https://github.com/bazelbuild/rules_apple/pull/1399). But during the review we think the proper fix should be added to rules-ios.

Without this fix, the unit tests added in this PR will fail with
```
tawang@tawang-md1  ~/Developer/external_repos/rules_ios_forked/tests/ios/frameworks/objc  ➦ e737e68  bazel build ObjcFrameworkTests --spawn_strategy=standalone
INFO: Analyzed target //tests/ios/frameworks/objc:ObjcFrameworkTests (2 packages loaded, 58 targets configured).
INFO: Found 1 target...
ERROR: /Users/tawang/Developer/external_repos/rules_ios_forked/tests/ios/frameworks/objc/BUILD.bazel:4:16: Bundling Precompiled Resource Bundle FooResources failed: (Exit 1): bundletool_experimental failed: error executing command 
  (cd /private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios && \
  exec env - \
  bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/frameworks/objc/ObjcFramework-FooResources-intermediates/bundletool_actions.json)
# Configuration: a19de01546f771c6fb5133c3624f1772775fe90e50f26db36762b52c0ff1abce
  File "/private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios/bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.py", line 303, in <module>
    _main(sys.argv[1])
  File "/private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios/bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.py", line 295, in _main
    bundler.run()
  File "/private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios/bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.py", line 136, in run
    self._add_files(f['src'], f['dest'], f.get('executable', False),
  File "/private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios/bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.py", line 168, in _add_files
    self._copy_file(fsrc, fdest, executable, bundle_root)
  File "/private/var/tmp/_bazel_tawang/c180ded8e4c6974890d9ade80503bb20/execroot/build_bazel_rules_ios/bazel-out/darwin-opt-exec-2B5CBBC6-ST-43530392d69d/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool_experimental.py", line 219, in _copy_file
    raise Exception(f"failed to clonefile {src} to {full_dest}")
Exception: failed to clonefile tests/ios/frameworks/objc/Resources/en.lproj/foo.strings to bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/frameworks/objc/ObjcFramework-FooResources/FooResources.bundle/en.lproj/foo.strings
Target //tests/ios/frameworks/objc:ObjcFrameworkTests failed to build
INFO: Elapsed time: 0.915s, Critical Path: 0.44s
INFO: 6 processes: 6 internal.
```

Diffs for xcodeproj:
- `tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/project.pbxproj`: new resources are added
- `tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj/project.pbxproj`: duplicated `Images.xcassets` is removed
- `tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/project.pbxproj`: duplicated `Images.xcassets` is removed


tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj/project.pbxproj
```diff
@@ -106,14 +104,6 @@
 			path = "test-imports-app";
 			sourceTree = "<group>";
 		};
-		49ACE10FF598298C0C4EA7D9 /* Resources2 */ = {
```
gyfelton, line 109: the for the contribution! Is this intentional? We added this probably to test the situation when there are two image assets under one project...

gyfelton:
Adding second review since this might break downstream projects once picked up

rules/library/resources.bzl
```diff
@@ -4,6 +4,23 @@ def _is_apple_resource_file(file, extensions_to_filter):
 
     return True
 
+# Remvoe files whose parent folders are already in the list. E.g., when both "foo/" and "foo/bar"
+# are in the files list, we should only return "foo/". This de-duplication is needed after this
+# rules-apple PR: https://github.com/bazelbuild/rules_apple/pull/1311. Otherwise 'clonefile' will
+# raise exceptions when both "foo/" and "foo/bar" are copied.
+def _deduplicate_paths(files):
```
gyfelton, line 11: the name does not match with implementation in my opinion... maybe sth like `_most_common_parent_dirs` for files which then means in extreme case, return just one dir if all files under same root dir?

tests/ios/frameworks/objc/BUILD.bazel
```diff
@@ -9,6 +9,10 @@ apple_framework(
         "main.m",
     ],
     platforms = {"ios": "10.0"},
+    resource_bundles = {"FooResources": glob(
+        ["Resources/**/*"],
```
gyfelton, line 13: can we add another Resources dir (Resources2) to verify that both dirs get cloned over?

tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj/project.pbxproj
larrytaowang: oh thanks for cathing this! you are right, this is a bug in my PR. `Resources2` are ignored because it also starts with `Resources`. Let me fix this bug

rules/library/resources.bzl
larrytaowang: make sense, let me do the renaming

tests/ios/frameworks/objc/BUILD.bazel
```diff
@@ -9,6 +9,10 @@ apple_framework(
...
+        exclude_directories = 0,
```
ob, line 14: I'm a bit confused about the use case here... why do we need `exclude_directories = 0` here? Is this a usage that `rules_apple` works with and `rules_ios` does not?

larrytaowang: In `cocoapods-bazel`, [this line](https://github.com/bazel-ios/cocoapods-bazel/blob/master/lib/cocoapods/bazel/target.rb#L355) explicitly set `exclude_directories = 0`. I will check if we can set this to `1` without losing any bundle file.

tests/ios/xcodeproj/Test-Imports-App-Project.xcodeproj/project.pbxproj
larrytaowang: Fixed

tests/ios/frameworks/objc/BUILD.bazel
larrytaowang: Added

rules/library/resources.bzl
larrytaowang: fixed.

larrytaowang:
Fixed bugs in my previous commits. I should not do the filter in `resources.bzl`, otherwise xcodeproj will not able to generate the correct resources. xcodeproj requires both `foo/` and `foo/bar` provided as resources path. So in the latest commit, I do the filter in `precompiled_apple_resource_bundle.bzl`.

I think the proper fix should be in rules-apple, to not copy the contents of a directory.

larrytaowang:
We have a ongoing disuccsionin rules-apple's PR about where the bundltool behaviors are expected or not. https://github.com/bazelbuild/rules_apple/pull/1399 

In the meantime, I am not sure why in the xcodeproj, the resources are included as temp files, whose name changed every time so the xcodeporj tests failed. let me check why this happened and how to fix it.

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -148,13 +179,16 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             # the need to end up
             fail("Got unexpected target location '{}' for '{}'"
                 .format(target_location, sources_list))
-        input_files.extend(sources_list)
         for source in sources_list:
+            if not most_common_parent_resources_set.get(source):
+                continue
```
gyfelton, line 184: will it be bad if we hit continue here? or it's simply not a resource `source` so we skip it?

larrytaowang: `most_common_parent_resources_set ` returns a Dict<Path, Boolean>. For the most common parent directories, the value is true, otherwise, it is false.  For example, if we have a `foo/` and `foo/bar` path, we don't want to include the `foo/bar` resources, so the dict will be 
```
foo/ -> True
foo/bar -> False
```
When this if statement hits `foo/bar`, we want to continue so this `foo/bar` is not included in the bundle.

larrytaowang:
busy this week. will come back to this PR later, maybe next week.

larrytaowang:
@gyfelton  I was doing experiment in this project: [xcodegen-localized.zip](https://github.com/bazel-ios/rules_ios/files/8498924/xcodegen-localized.zip).
- for `Resources2` folder, I only include the folder path, e.g,
```
- path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/Resources2"
        optional: true
```
- for `Resources` folder, i include the folder and every file, e.g.,
```
- path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/Resources/en-CA.lproj"
        optional: true     
      - path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/Resources/en-CA.lproj/Localized.strings"
        optional: true
      - path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/Resources/en-CA.lproj/empty.json"
...
```

As for the generated xcodeporj, `Resources` include everything correctly, but `Resources2` does not.
Screen Shot 2022-04-15 at 4 59 58 PM

So I think right now **we have to provide both directory and file path to Xcodegen**. However, the issue is that the reference is a temporary value, which changes in every xcodeproj and will fails our CI tests every time. E.g., in `project.pbxproj`, for `Resources` the generated reference values are 
```
A3B91F68A2C75A2768B83823 /* empty.json in Resources */ = {isa = PBXBuildFile; fileRef = "TEMP_4814A821-4434-4B43-8472-492642E3B567" /* empty.json */; };
A924B0907ABAEB6A0645F235 /* empty.json in Resources */ = {isa = PBXBuildFile; fileRef = "TEMP_2573875C-3416-4EBA-B7AD-94BC7A2A028F" /* empty.json */; };
```

I think the long term solution is that we create an issue for XcodeGen, either 
-  no long use temp value like `TEMP_4814A821-4434-4B43-8472-492642E3B567` for refence
- when we provide a directory path like `rules-ios/objc/Resources2`, xcodegen should include the resources correctly

Before this issue is fixed, can I remove the tests I added so this PR can get merged? Or do you see a better way to solve this problem? Thanks!

larrytaowang:
Remove the `Resources` added in previous commits in this PR, so it could pass the xcodeproj tests now.
Also create an issue for xcode generator: https://github.com/yonaskolb/XcodeGen/issues/1210

gyfelton:
Btw I did a test on our project that uses rules-ios and it did break tests by crashing the tests because of resouces not found... so unfortiantely we cannot merge this until our side has bandwidth to deal with this change and get a green CI build

larrytaowang:
@gyfelton, thanks for fixing this in rules-ios! This issue only happens if we write our rules like this
```
s.resource_bundle = { 'XXXResources' => [ '**/*' ] }
```
We got it around by changing our BUILD files to list the extensions explicitly.
```
s.resource_bundle = { 'XXXResources' => [ '**/*.lproj', '**/*.xib, ... ] }
```

PR #443, opened 2022-04-07
rules_ios <= XCHammer (Pt. 1)

thiagohmcruz:
#### Summary

As part of the plan to asses moving the Xcode project generator to XCHammer we'll use `rules-ios-xchammer` as a base in both the Tulsi and XCHammer forks for some time while we iterate on this. This PR requires:

* https://github.com/bazel-ios/tulsi/pull/1
* https://github.com/bazel-ios/xchammer/pull/1
* - [x] Bump XCHammer to lastest tag in `rules-ios-xchammer` git rev after the two PRs above land

#### Changes in this PR

* Load XCHammer dependencies
* Introduce `use_xchammer` analysis phase flag to declare the XCHammer rule instead
* Add files to be ignored for now to `.gitignore` (will add tests with golden files to be diff-ed as follow up, need to figure out a way to put the XCHammer proj in a sub-folder)

#### Testing

Added a dummy target just so we can test this during development:
```sh
bazel build --spawn_strategy=standalone //tests/ios/xcodeproj:App-XCHammer
```


luispadron (approved):
Nice excited to try this out, should we add the docs for this rule or is it still very in flux? I generated some older docs previously in #438

thiagohmcruz:
> 

Thx for the review @luispadron! Yeah, we might want to hold the docs for now, this rule is probably going to change soon so it would have to be re-generated anyways. That said this is still in very early stages, this PR just allows us to start iterating so I wouldn't recommend using this yet!

rules/repositories.bzl
```diff
@@ -112,6 +112,20 @@ swift_binary(
         """,
     )
 
+    if not native.existing_rule("xchammer"):
+        git_repository(
+            name = "xchammer",
+            commit = "792e31212bf244c1a4af6e1ecb5806c4f0eb64f2",
```
jerrymarino, line 118: We might need to create a tag for each of the versions we push into `rules_ios` - otherwise if we remove the fork or branches it could cause an error on the devs side

rules/xcodeproj.bzl
```diff
@@ -1255,3 +1257,17 @@ Additional LLDB settings to be added in each target's .lldbinit configuration fi
     },
     executable = True,
 )
+
+def xcodeproj(name, use_xchammer = False, **kwargs):
```
jerrymarino, line 1261: Did you consider to organize this in a way that we can flip it on / off with a configurable variable like other features. Then, we can test our entire CI / transparently migrate people while providing an opt-out:

```
--features rules_ios.disable_xchammer
```

To do so ATM you can probably define both rules:

```
macro
   xchammer_xcodeproj(name=name_xch, project_name=..)
   legacy_xcodeproj(name=name_legacy project_name=..)
   alias(name=name, actual=select(xch: name_xch, default=name_legacy))
````

File: rules/xcodeproj.bzl [hunk 3]
```diff
@@ -1150,7 +1152,7 @@ def _xcodeproj_impl(ctx):
         ),
     ]
 
-xcodeproj = rule(
+_xcodeproj = rule(
```
jerrymarino, line 1155: I'd suggest moving the implementation of the legacy rule to `rules/internal/legacy_xcodeproj.bzl` and trying to leave a minimal file here at some point. Migrating the name to `_xcodeproj` isn't necessary if you do this now load it as:
```
load(rules/internal/legacy_xcodeproj.bzl, legacy_xcodeproj=xcodeproj)
```

jerrymarino (commented):
This looks like it nicely fits in with rules_ios ! Great work @thiagohmcruz !

Can you a PR to `bazel-ios/xchammer`  with the changes we need to get V1 working and in-order prior to merging this? We'll need some strategy in-place to persist any dependent commits in github branches of XCHammer / tulsi in order to keep rules_ios working before upstreaming if we go that route!

rules/xcodeproj.bzl
thiagohmcruz: I didn't consider using an `alias` here! Thx this is much better, I'll use this approach instead

rules/xcodeproj.bzl
thiagohmcruz: Makes sense to me to move to another file, will do!

rules/repositories.bzl
thiagohmcruz: Yeah, I'll have a branch in each fork and we can open PRs into those for now then bump the reference here, not super worried about upstreaming things yet as there's some work to be done first

rules/xcodeproj.bzl
thiagohmcruz: Done!

rules/xcodeproj.bzl
thiagohmcruz: Done!

rules/repositories.bzl
thiagohmcruz: Btw re tags yeah makes sense, once I land this will create a tag in `master`, something like `xchammer-0.0.1`

rules/xcodeproj.bzl
thiagohmcruz: Some problems here though, the tests that run on CI often call `//...` or `//tests/ios/...` and that includes the xchammer targets now that we're adding both rules to the BUILD files, will make some changes to ignore those for now on the CI tests

thiagohmcruz:
So using a feature flag has proven to be problematic, I'm going back to an analysis phase flag here:

* Our test would fail since there's `//...` queries in there and that was picking up the XCHammer targets since they were being added to the BUILD files via a macro. With the analysis phase flag we don't declare those targets unless the flag is specified
* Using an `alias` worked well to specify things during a Bazel invocation but had to add suffixes to each rule's name `_xchammer`, `_legacy` and that would change the file name on disk. That's broken things for us internally since there's asssumptions on the fact that the rule name and the generated file on disk. We could change both rules to specify the filename separately but that could surface more issues, not sure as I didn't go on that route.

My main goal with this was to minimize impact in existing workflows and land something simple that allows us to iterate, so I'm taking the same approach as before here. I really like the idea of moving logic to a different extension file for the legacy stuff though so I did that!

cc @jerrymarino

rules/xcodeproj.bzl
thiagohmcruz: More details here: https://github.com/bazel-ios/rules_ios/pull/443#issuecomment-1105441681

rules/xcodeproj.bzl
thiagohmcruz: Reverted this, more details here: https://github.com/bazel-ios/rules_ios/pull/443#issuecomment-1105441681

rules/xcodeproj.bzl
```diff
@@ -1,1260 +1,30 @@
...
-        srcs = [f for f in srcs if _is_current_project_file(f)]
-        non_arc_srcs = [f for f in non_arc_srcs if _is_current_project_file(f)]
-        asset_srcs = [f for f in asset_srcs if _is_current_project_file(f)]
-        framework_includes = _get_attr_values_for_name(deps, _SrcsInfo, "framework_includes")
-        cc_defines = _get_attr_values_for_name(deps, _SrcsInfo, "cc_defines")
-        swift_defines = _get_attr_values_for_name(deps, _SrcsInfo, "swift_defines")
-        if CcInfo in target:
-            framework_includes.append(target[CcInfo].compilation_context.framework_includes)
-            cc_defines.append(target[CcInfo].compilation_context.defines)
+    # Pop XCHammer specific attributes so these dont' get propagated
```
jerrymarino, line 15: Minor `don't` !

rules/xcodeproj.bzl
```diff
@@ -1,1260 +1,30 @@
...
+    use_xchammer = kwargs.pop("use_xchammer", False)
+    generate_xcode_schemes = kwargs.pop("generate_xcode_schemes", False)
```
jerrymarino, line 17: Nice

jerrymarino (approved):

rules/xcodeproj.bzl
thiagohmcruz: Fixed!

rules/repositories.bzl
thiagohmcruz: * https://github.com/bazel-ios/xchammer/releases/tag/rules_ios-0.0.1
* https://github.com/bazel-ios/tulsi/releases/tag/rules_ios-0.0.1

PR #445, opened 2022-04-11
Fixes import_middleman when ios_multi_cpus=x86_64,sim_arm64

congt:
##  Problem

It crashes with building a dynamic framework with vendored xcframwork with flag --ios_multi_cpus=x86_64,sim_arm64.

## Root cause

We wrongly asserted import_middleman rule to be only analyzed in sim_arm64 configuration. The select method uses the configuration of the current target's configuration. The outgoing edge transition doesn't affect the select method. Therefore, it's possible that an import_middleman target with a non-sim_arm64 configuration is analyzed during the analysis phase, which causes crash.

## Solution

import_middleman should be no op when the configuration is not for sim_arm64. Also uses the right split_attr in Framework to make sure that the right configured deps are used.

## Alternative solution

An alternative solution is to change transition_support.apple_rule_transition to output //command_line_option:ios_multi_cpus with a single cpu. It works because a framework for a specific arch doesn't need deps for other archs. Here is the PR: https://github.com/bazel-ios/rules_ios/pull/450


rules/library.bzl
```diff
@@ -680,13 +680,12 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     for vendored_dynamic_library in kwargs.pop("vendored_dynamic_libraries", []):
         fail("no support for dynamic library: %s" % vendored_dynamic_library)
 
-    # TODO(jmarino)Perhaps it uses a import_middleman here
     if len(vendored_deps):
         import_middleman(name = name + ".import_middleman", deps = vendored_deps, tags = ["manual"])
         deps += select({
             "@build_bazel_rules_ios//:arm64_simulator_use_device_deps": [name + ".import_middleman"],
```
congt, line 686: The current target's configuration determines the select method. As we use the transition `apple_common.multi_arch_split` on deps, it's possible that import_middleman target is analyzed in a non-sim_arm64 configuration.

rules/library.bzl
```diff
@@ -680,7 +680,6 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     for vendored_dynamic_library in kwargs.pop("vendored_dynamic_libraries", []):
         fail("no support for dynamic library: %s" % vendored_dynamic_library)
 
-    # TODO(jmarino)Perhaps it uses a import_middleman here
     if len(vendored_deps):
         import_middleman(name = name + ".import_middleman", deps = vendored_deps, tags = ["manual"])
         deps += select({
```
congt, line 685: The current target's configuration determines the select method. As we use the transition apple_common.multi_arch_split on deps, it's possible that import_middleman target is analyzed in a non-sim_arm64 configuration.

congt:
This PR doesn't solve all problems. It fails when building a dynamic framework depending on another dynamic framework with multiple archs.

Closing in favor for https://github.com/bazel-ios/rules_ios/pull/450

PR #448, opened 2022-04-11
Toggle disable main thread checker in for schemes

ethan-gates:
This PR adds the ability to toggle the disable main thread checker boolean in the diagnostics tab of a scheme.  The previous behavior was using the xcode's default which is false, so the new xcodeproj() attribute default is also false.

xcodegen documentation for disableMainThreadChecker: https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#common-build-action-options


nanwng (approved):
Thanks for looking into this!

jerrymarino (approved):

PR #450, opened 2022-04-12
Transition ios_multi_cpus to to a single cpu in apple_rule_transition

congt:
It fixes the crash when building a dynamic framework with vendored xcframwork with flag --ios_multi_cpus=x86_64,sim_arm64.

## Problem

import_middleman fails when building for non-sim_arm64. See https://github.com/bazel-ios/rules_ios/blob/master/rules/import_middleman.bzl#L174

Previously we don't change  ios_multi_cpus in the apple_rule_transition. As the deps attribute of apple_framework_packaging rule use the apple_common.multi_arch_split transition, it will build import_middleman targets for all archs. Note that the outgoing edge transition doesn't affect the select() method.

## Solution
Transition ios_multi_cpus to to a single cpu in apple_rule_transition. The idea is that a framework doesn't need any artifact from other arch. It will simplify the configured target graph.


thiagohmcruz (approved):
Discussed offline the details, LGTM!

PR #452, opened 2022-04-14
Fix: Remove NSPrincipalClass in rules Info.plist files

luispadron:
## Summary

This fixes an issue in unit test bundles where defining `NSPrincipalClass` within a test bundle's Info.plist would cause a build failure due to merging plists.

This PR fixes this issue by removing the defined `NSPrincipalClass` key from the `Info.plists` used by the rules. This key is already automatically defined by Xcode for iOS and macOS app targets so I don't believe we need to define this key in the rules.

> Xcode sets the default value of this key to NSApplication for macOS apps, and to UIApplication for iOS and tvOS apps. For other types of bundles, you must set this key in The Info.plist File.

Source: https://developer.apple.com/documentation/bundleresources/information_property_list/nsprincipalclass

I've also added a unit test to assert this behavior, this new test fails in the current master branch with error:

```sh
bazel test //tests/ios/unit-test:NSPrincipalClassDefinedInPlist    
...
...                                                  
ERROR: While processing target "//tests/ios/unit-test:NSPrincipalClassDefinedInPlist.__internal__.__test_bundle"; found key "NSPrincipalClass" in two plists with different values: "TestSuiteObserver" != ""
Target //tests/ios/unit-test:NSPrincipalClassDefinedInPlist failed to build
INFO: Elapsed time: 0.349s, Critical Path: 0.20s
INFO: 2 processes: 2 internal.
//tests/ios/unit-test:NSPrincipalClassDefinedInPlist            FAILED TO BUILD

```

With this change the new test should pass successfully.

### DIFF
--- a/rules/test_host_app/Info.plist
+++ b/rules/test_host_app/Info.plist
@@ -16,8 +16,6 @@
     <string>????</string>
     <key>CFBundleVersion</key>
     <string>1.0.0</string>
-    <key>NSPrincipalClass</key>
-    <string></string>
     <key>CFBundleDevelopmentRegion</key>
     <string>en</string>
     <key>CFBundleExecutable</key>
--- a/tests/ios/unit-test/BUILD.bazel
+++ b/tests/ios/unit-test/BUILD.bazel
@@ -43,3 +43,14 @@ ios_unit_test(
     minimum_os_version = "11.0",
     resources = ["resource-file.txt"],
 )
+
+ios_unit_test(
+    name = "NSPrincipalClassDefinedInPlist",
+    srcs = [
+        "TestSuiteObserver.swift",
+        "empty.m",
+        "empty.swift",
+    ],
+    infoplists = [{"NSPrincipalClass": "TestSuiteObserver"}],
+    minimum_os_version = "11.0",
+)
--- a/tests/ios/unit-test/TestSuiteObserver.swift
+++ b/tests/ios/unit-test/TestSuiteObserver.swift
@@ -1,0 +1,20 @@
+import Foundation
+import XCTest
+
+/// Observes significant events in the progress of test runs
+/// Demonstrates defining NSPrincipalClass in unit test bundles.
+@objc(TestSuiteObserver)
+class TestSuiteObserver: NSObject, XCTestObservation {
+
+    // MARK: - Life Cycle
+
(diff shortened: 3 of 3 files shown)


gyfelton (approved):
thx for the contribution! makes sense to me to remove this default field

rules/test_host_app/Info.plist
```diff
@@ -16,8 +16,6 @@
     <string>????</string>
     <key>CFBundleVersion</key>
     <string>1.0.0</string>
-    <key>NSPrincipalClass</key>
```
jerrymarino, line 19: Do you need to remove this both from `resource_bundle.plist` and the default test host app?

tests/ios/unit-test/BUILD.bazel
```diff
@@ -43,3 +43,14 @@ ios_unit_test(
...
 )
+
+ios_unit_test(
+    name = "NSPrincipalClassDefinedInPlist",
+    srcs = [
+        "TestSuiteObserver.swift",
+        "empty.m",
+        "empty.swift",
+    ],
+    infoplists = [{"NSPrincipalClass": "TestSuiteObserver"}],
```
jerrymarino, line 54: Thanks for adding this case, this test doesn't have a test_host - so shouldn't need to remove it from there?

jerrymarino (commented):
Does rules_apple do anything here too?

rules/test_host_app/Info.plist
luispadron: Good point, let me try without this change and report back. Although I wonder if providing less is better here if it works as expected without the key? It doesn't look like we have it defined either unless we inject it somewhere in the rules.

luispadron: So looks like only the `test_host_app` info plist needs the change. Changing resources_bundle.plist doesn't have an effect here so I reverted that change.

jerrymarino: Ok let's run this change on CI and if it passes we can fix fallout as a followup

jerrymarino (approved):

rules/test_host_app/Info.plist
luispadron: Looks like it passed 🚀

PR #455, opened 2022-04-19
Add a basic test for integration with PodToBuild via WORKSPACE (no update pod needed)

gyfelton:
Add a test example that fetches one OSS repo and import it for two targets: one in swift and one in objc.


jerrymarino (approved):
Thanks for the PR @gyfelton LGTM! There might be some edge cases around transitive deps

PR #458, opened 2022-04-25
Add SwiftGen example to tests/ios

luispadron:
Adds an example using the `swiftgen` rule originally from https://github.com/line/rules_apple_line/blob/main/apple/swiftgen.bzl
and modified to work with additional asset types.

The example and test is available at [tests/ios/swiftgen](/tests/ios/swiftgen)


tests/ios/swiftgen/swiftgen.bzl
```diff
@@ -0,0 +1,187 @@
...
+
+    ctx.actions.run(
+        inputs = inputs,
+        outputs = [output],
+        executable = ctx.executable._swiftgen,
+        arguments = [args],
+        mnemonic = "SwiftGen",
+    )
+
+swiftgen = rule(
```
luispadron, line 158: Should we add this rule to the repository? Its from https://github.com/line/rules_apple_line but maybe it would be good for users to have this rule as well without having to pull another repo for it? 

Could also see wanting to keep this repo just focused on Apple specific rules so I'm fine with either direction!

tests/ios/swiftgen/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+
+apple_framework(
+    name = "SwiftGenTest",
+    srcs = glob(["Sources/**/*.swift"]) +
+    # Depend on the swiftgen action created above.
+    # By depending on the swiftgen action Bazel will ensure the generated
+    # sources are up-to-date before building this target.
+    # This also means it will skip generation if there are no changes.
+    [
+        ":SwiftGenTest-BundleResources.swift",
```
jerrymarino, line 23: I feel like we should document this somewhere in case someone hits it again ( that you ATM have to give it a label named `.swift` or `.m` ) until we refactor it!

tests/ios/swiftgen/swiftgen.bzl
jerrymarino: If this is a test rule, I think it'd be good to ensure that we can depend on `rules_apple_line` stuff in general and everything is handled.

Because it is a `test` dependency - it can go into the `WORKSPACE` file: consumers of `rules_ios` shouldn't need to add `rules_apple_line` because it's in the `test` dir I believe. If you checkout my PR to verify proto support - this is what we did for `rules_proto_grpc`

jerrymarino (approved):
Thanks for the PR @luispadron ! Overall  LGTM 🚢   Would you add `rules_apple_line` for the test case to the `WORKSPACE` file prior to landing?

tests/ios/swiftgen/BUILD.bazel
luispadron: Agree, I'll find a doc to up date

luispadron:
> Looks like there's some conflicts to solve here?

Yeah there is a failure with pulling in the rules apple line dependencies. I've opened a PR on that repo.

Issue #460, opened 2022-04-26
Mitigate or document transitive WORKSPACE files


jerrymarino:
I now get errors when I pull rules_ios master.

Recent changes implicated adding these lines to the `WORKSPACE` file after loading `rules_ios`:

```.py
# WORKSPACE
# ...

rules_ios_dependencies()
# ...

load("@xchammer//third_party:repositories.bzl", "xchammer_dependencies")

xchammer_dependencies()

load("@xcbuildkit//third_party:repositories.bzl", xcbuildkit_dependencies = "dependencies")

xcbuildkit_dependencies()
```

If people don't use the project gen rule we might need a solution to either conditionally load it in or inline dependencies unitl `bzlmod` provides us a fix to not compilicate integrating rules_ios 

```
ERROR: /Users/jmarino/Development/ios-register/tools/rules/library/BUILD.bazel:12:8: error loading package '': at /private/var/tmp/_bazel_jmarino/ca0596bbcafdd3d97374bd4573e006cf/external/build_bazel_rules_ios/rules/xcodeproj.bzl:4:6: at /private/var/tmp/_bazel_jmarino/ca0596bbcafdd3d97374bd4573e006cf/external/xchammer/BazelExtensions/xcodeproject.bzl:3:5: at /private/var/tmp/_bazel_jmarino/ca0596bbcafdd3d97374bd4573e006cf/external/xchammer/BazelExtensions/tulsi.bzl:5:5: Unable to find package for @xchammer_tulsi_aspects//:tulsi/tulsi_aspects_paths.bzl: The repository '@xchammer_tulsi_aspects' could not be resolved: Repository '@xchammer_tulsi_aspects' is not defined. and referenced by '//tools/rules/library:cocoapods_atlas_whitelist'
[...]

PR #461, opened 2022-04-26
Load XCHammer and XCBuildKit deps

thiagohmcruz:
#460 

Copying files here is not ideal but until .bzlmod is ready I don't see a better way to handle this and don't force consumers to update their `WORKSPACE` files to add XCHammer specific `load`s


jerrymarino (approved):

jerrymarino:
Thanks for the fix @thiagohmcruz !

PR #466, opened 2022-05-03
Adding xcodeproj() attr to force x86 binaries for sim on M1

ethan-gates:
## Summary

This PR adds the opt-in ability to force simulator builds to use `--ios_multi_cpus=x86_64` only on Apple Silicon.  The goal is to allow simulator builds that use external dependencies that aren't built for M1 yet to run under rosetta, while maintaining their ability to run on device.

## Details
A new boolean attribute was added to `xcodeproj()` called `force_x86_sim_on_apple_silicon` that will trigger this behavior.  The rule now injects a user defined build setting into the generated .xcodeproj that will be picked up by the bazel-wrapper.sh script.  To determine if the host architecture is Apple the script uses `sysctl` looking for cpu branding.  One alternative considered is using `uname` instead, however `uname` will report x86 if its parent process is running under rosetta.

The attribute defaults to false which is a no-op, and realistically only needs to be used during the transition to M1 compatible frameworks.

### DIFF
--- a/tools/xcodeproj_shims/build-wrapper.sh
+++ b/tools/xcodeproj_shims/build-wrapper.sh
@@ -22,6 +22,9 @@ fi
 if [ -n "${TARGET_DEVICE_IDENTIFIER:-}" ] && [ "$PLATFORM_NAME" = "iphoneos" ]; then
     echo "Builds with --ios_multi_cpus=arm64 since the target is an iOS device."
     BAZEL_BUILD_OPTIONS+=("--ios_multi_cpus=arm64")
+elif [ $FORCE_X86_SIM -gt 0 ] && [ "$PLATFORM_NAME" = "iphonesimulator" ] && [[ $(sysctl -n machdep.cpu.brand_string) =~ "Apple" ]]; then
+    echo "Builds with --ios_multi_cpus=x86_64 since the target is sim, host architecture is apple silicon, and xcodeproj attribute force_x86_sim_on_apple_silicon is True"
+    BAZEL_BUILD_OPTIONS+=("--ios_multi_cpus=x86_64")
 fi
 
 # If bazel configs (from .bazelrc file) were specificed and the current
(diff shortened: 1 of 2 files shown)


jerrymarino:
Did you guys try to flip on device deps?
```
--features apple.arm64_simulator_use_device_deps
```

If you're feeding `apple_framework` xcframeworks setting that [should automatically set it up to use the device deps on M1 sim](https://developer.squareup.com/blog/getting-squares-ios-build-ready-for-apple-silicon-with-bazel/). I suspect it might work out of the box depending on what kind of deps you have..

jerrymarino:
_if you just set that in `.bazelrc` it will only take effect for `sim_arm64` only on Bazel 5+_

chenxiao0228:
> 

@jerrymarino  you are right, It works for the most part... We have one pre-compiled dependency AzureCommunications that would crash after the arm-to-sim. binary patch. We have some of these dependencies either not publishing xcframework or blocked by this migration  to bump to newer versions with xcframeworks.

We plana to use this workaround for the interim.

tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -22,6 +22,9 @@ fi
 if [ -n "${TARGET_DEVICE_IDENTIFIER:-}" ] && [ "$PLATFORM_NAME" = "iphoneos" ]; then
     echo "Builds with --ios_multi_cpus=arm64 since the target is an iOS device."
     BAZEL_BUILD_OPTIONS+=("--ios_multi_cpus=arm64")
+elif [ $FORCE_X86_SIM -gt 0 ] && [ "$PLATFORM_NAME" = "iphonesimulator" ] && [[ $(sysctl -n machdep.cpu.brand_string) =~ "Apple" ]]; then
```
chenxiao0228, line 25: We probably do not need the sysctl check.
It wont hurt setting it to x86_64 even on Intel machines.

tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -22,6 +22,9 @@ fi
...
+    echo "Builds with --ios_multi_cpus=x86_64 since the target is sim, host architecture is apple silicon, and xcodeproj attribute force_x86_sim_on_apple_silicon is True"
```
chenxiao0228, line 26: This can also be simplified to be something like:
Using x86_64 arch for simulator build because of the flag

chenxiao0228 (approved):

PR #470, opened 2022-05-04
Add xcode.compile_with_xcode feature

jerrymarino:
In VFS it can't load transitive files from bazel-out


jerrymarino:
Set here: https://github.com/bazel-ios/xchammer/pull/10

thiagohmcruz (approved):

PR #471, opened 2022-05-06
[XCHammer] Bump xcodegen to latest

jerrymarino:
Take the latest build of Xcode gen


thiagohmcruz (approved):
We could also bump XCHammer here now that you landed similar changes there

PR #475, opened 2022-05-12
Set bundle_id in precompiled_apple_resource_bundle

thiagohmcruz:
#### Summary

Picks up:
* https://github.com/bazel-ios/tulsi/pull/3
* https://github.com/bazel-ios/xchammer/pull/15

Also set `bundle_id` in `precompiled_apple_resource_bundle`. This value not being present was causing a runtime crash on XCHammer+`generate_xcode_schemes = True`.

Minor: add `xchammer-includes` to `.gitignore`

#### Testing

To test this apply this diff:
```
diff --git a/tests/ios/xcodeproj/BUILD.bazel b/tests/ios/xcodeproj/BUILD.bazel
index c782d0a..15c70bd 100644
--- a/tests/ios/xcodeproj/BUILD.bazel
+++ b/tests/ios/xcodeproj/BUILD.bazel
@@ -202,6 +202,7 @@ xcodeproj(
 
 xcodeproj(
     name = "App-XCHammer",
+    generate_xcode_schemes = True,
     use_xchammer = True,
     deps = [
         "//tests/ios/app:App",
```
and build this target
```sh
bazel build --spawn_strategy=standalone tests/ios/xcodeproj:App-XCHammer
```
and the note how the bundle identifier now is set:

Screen Shot 2022-05-12 at 7 53 20 AM

#### Note

Note that this is backwards compatible because:
* The `bundle_id` attr is handled in a way that if consumers are not setting that at all it continues to be `None`
* `generate_xcode_schemes = False` is not affected by this change since building in Bazel mode doesn't require that value set in Xcode (you can test that by reverting the diff and generating the same target above and then building the `App` scheme)

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -64,7 +64,10 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         "//conditions:default": None,
     }))
 
+    # We need to put back bundle_id in order to set that attr for 'precompiled_apple_resource_bundle' targets in 'apple_library'
+    kwargs["bundle_id"] = framework_packaging_kwargs.get("bundle_id", None)
     library = apple_library(name = name, **kwargs)
+
     framework_deps = []
     if framework_packaging_kwargs.get("link_dynamic") == True:
         # Setup force loading here - only for direct deps / direct libs
@@ -72,6 +75,7 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         force_load_direct_deps(name = force_load_name, deps = kwargs.get("deps") + library.lib_names, tags = ["manual"])
         framework_deps.append(force_load_name)
     framework_deps += library.lib_names
+
     apple_framework_packaging(
         name = name,
         framework_name = library.namespace,
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -175,13 +175,14 @@ FOUNDATION_EXPORT const unsigned char {module_name}VersionString[];
     )
     return destination
 
-def _generate_resource_bundles(name, library_tools, module_name, resource_bundles, platforms, **kwargs):
+def _generate_resource_bundles(name, library_tools, module_name, resource_bundles, platforms, bundle_id, **kwargs):
     bundle_target_names = []
     for bundle_name in resource_bundles:
         target_name = "%s-%s" % (name, bundle_name)
         precompiled_apple_resource_bundle(
             name = target_name,
             bundle_name = bundle_name,
+            bundle_id = "{}.{}".format(bundle_id, bundle_name) if bundle_id else None,
             resources = [
                 library_tools["wrap_resources_in_filegroup"](name = target_name + "_resources", srcs = resource_bundles[bundle_name]),
             ],
@@ -694,6 +695,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         resource_bundles = kwargs.pop("resource_bundles", {}),
         module_name = module_name,
         platforms = platforms,
+        bundle_id = kwargs.pop("bundle_id", None),
         **kwargs
     )
     deps += resource_bundles
--- a/tests/ios/frameworks/with-bundle-id/BUILD.bazel
+++ b/tests/ios/frameworks/with-bundle-id/BUILD.bazel
@@ -1,0 +1,12 @@
+load("//rules:framework.bzl", "apple_framework")
+
+apple_framework(
+    name = "FrameworkWithBundleId",
+    srcs = [
+        "FrameworkWithBundleId/Foo.h",
+        "FrameworkWithBundleId/Foo.m",
+    ],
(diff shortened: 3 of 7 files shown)


tests/ios/frameworks/with-bundle-id/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+load("//rules:framework.bzl", "apple_framework")
+
+apple_framework(
```
thiagohmcruz, line 3: Actually to properly test this we need to set resource bundles here, the bundle identifier was also missing in the fmw regarding but I'll fix that so this test is testing the right thing (then update the PR description/screenshot)

thiagohmcruz: Done

rules/library.bzl
```diff
@@ -694,6 +695,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         resource_bundles = kwargs.pop("resource_bundles", {}),
         module_name = module_name,
         platforms = platforms,
+        bundle_id = kwargs.pop("bundle_id", None),
```
jerrymarino, line 698: In the change on line 178 `bundle_id` was added to the function arguments, I expected this would no longer be present in `kwargs`. e.g. this line should just access it via the local var `bundle_id`?

See other comment about setting this to the frameworks bundle ID and better constructing the id internally

jerrymarino:
Also thanks for adding a test case for this 🙇‍♂️ !

rules/library.bzl
thiagohmcruz: I think the fact that the definition happens way up is creating the confusion here, `bundle_id` was added to the macro that is set to `library_tools["resource_bundle_generator"]` here. 

Meaning that the param `bundle_id` doesn't exist at this point yet and this `pop` is doing exactly that and setting the `bundle_id` param. Maybe open the complete file to inspect and see what I'm talking about?

rules/framework.bzl
```diff
@@ -64,14 +64,18 @@ def apple_framework(name, apple_library = apple_library, **kwargs):
         "//conditions:default": None,
     }))
 
+    # We need to put back bundle_id in order to set that attr for 'precompiled_apple_resource_bundle' targets in 'apple_library'
```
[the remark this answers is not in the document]
thiagohmcruz: It wouldn't be the same, it would have the same fmw identifier as prefix + the bundle_name for the specific resource, grep this line in my PR:
```py
 bundle_id = "{}.{}".format(bundle_id, bundle_name)
```

Tbh I wasn't super sure how to setup this, what I implemented here gives us a unique value unless I'm missing something. But regarding will also take a look at what cocoapods does and compare, thx for calling it out!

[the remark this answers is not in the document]
thiagohmcruz: @jerrymarino see the [latest commit](https://github.com/bazel-ios/rules_ios/pull/475/commits/e902672e910380d97178c288f479610fab100e56):

[This value was already being set before and I didn't realize it](https://github.com/bazel-ios/rules_ios/blob/76a1551baf12f1784873be4e31372a2a29c5405b/rules/precompiled_apple_resource_bundle.bzl#L106), glad you caught this otherwise I'd be changing the value for the bundle identifier here :)

That said, the only thing we need is to move this assignment up to the rule declaration so aspects can collect it. Added to a utils file since it's used in two places, will make it easier to make future changes to this logic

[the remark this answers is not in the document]
thiagohmcruz: Also, no need to propagate the fmw `bundle_id`, that was just my first attempt to compose an identifier. I'll keep the fmw test though (without setting the bundle identifier) because we will eventually run the XCHammer targets as part of the tests here and we want to cover that test case

rules/utils.bzl
```diff
@@ -0,0 +1,4 @@
+"""Common helper macros that could be used in different BUILD and/or extension files"""
+
+def bundle_identifier_for_bundle(bundle_name):
+    return "com.cocoapods." + bundle_name
```
thiagohmcruz, line 4: CocoaPods applies some validation on this value, a lot of that logic requires regex support so we won't be able to reach parity here easily. This value was already being set before to what you see here, so don't see any harm in continuing to use it for now

rules/framework.bzl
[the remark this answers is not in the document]
thiagohmcruz: I think this is the right path to understand why we ended up having to collect this directly from an attribute, will track this separately and tackle that in a follow up

Also don't have the context of past decisions on that rule but can def investigate

thiagohmcruz:
Quick update: working on some changes we might need on the Tulsi repo side before landing, apparently we're bringing in more dep bundles than necessary causing a build failure. Very likely related to [this comment](https://github.com/bazel-ios/xchammer/pull/15#discussion_r870651976). Holding this until that is fixed.

PR #476, opened 2022-05-12
bug fix for "additional_files" in xcodeproj

larrytaowang:
# Summary
Bug fix for `additional_files` in xcodeproj and add a test case
- the `additional files` are not included as input so when building with sandbox, bazel complains that the `additional_files` are missing with the following error
```
Spec validation error: Invalid file group "../../../../../../tests/ios/frameworks/objc/additional_files/foo.podspec"
``` 
 - wrong xcodeproj JSON file is generated

Currently the generated JSON has the following contents for the additional files, which is not correct. The generated Xcode project will not include the additional files, i.e., the fileGroups correctly.
```
fileGroups:
  - path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/BUILD.bazel"
    optional: true
```
According to the [doc](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#project), `fileGroups` should be `A list of paths to add to the root of the project` like this 
```
fileGroups:
  -  "rules-ios/objc/BUILD.bazel1"
```

# Test
Add a test case and the additional files declared in the `filegroup` are included in the generated Xcode project as expected.

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -1054,7 +1054,7 @@ def _xcodeproj_impl(ctx):
     (xcodeproj_targets_by_name, xcodeproj_schemes_by_name) = _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_transitive_targets)
 
     project_file_groups = [
-        {"path": paths.join(src_dot_dots, f.short_path), "optional": True}
+        paths.join(src_dot_dots, f.short_path)
         for f in ctx.files.additional_files
         if _is_current_project_file(f)
     ]
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -1085,7 +1085,7 @@ def _xcodeproj_impl(ctx):
     ctx.actions.run(
         executable = ctx.executable._xcodegen,
         arguments = ["--quiet", "--no-env", "--spec", xcodegen_jsonfile.path, "--project", project.dirname],
-        inputs = [xcodegen_jsonfile],
+        inputs = [xcodegen_jsonfile] + ctx.files.additional_files,
         outputs = [project],
     )
     install_script = ctx.actions.declare_file(
--- a/tests/ios/frameworks/objc/BUILD.bazel
+++ b/tests/ios/frameworks/objc/BUILD.bazel
@@ -38,3 +38,11 @@ ios_unit_test(
     visibility = ["//visibility:public"],
     deps = [":ObjcFrameworkTestLib"],
 )
+
+filegroup(
+    name = "additional_files",
+    srcs = [
+        "additional_files/foo.podspec",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/tests/ios/frameworks/objc/BUILD.bazel
+++ b/tests/ios/frameworks/objc/BUILD.bazel
@@ -41,8 +41,8 @@ ios_unit_test(
 
 filegroup(
     name = "additional_files",
-    srcs = [
-        "additional_files/foo.podspec",
-    ],
+    srcs = glob([
+        "additional_files/**/*",
+    ]),
     visibility = ["//visibility:public"],
 )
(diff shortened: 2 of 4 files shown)


rules/legacy_xcodeproj.bzl
```diff
@@ -1054,7 +1054,7 @@ def _xcodeproj_impl(ctx):
     (xcodeproj_targets_by_name, xcodeproj_schemes_by_name) = _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_transitive_targets)
 
     project_file_groups = [
-        {"path": paths.join(src_dot_dots, f.short_path), "optional": True}
```
jerrymarino, line 1057: Any downsides to not setting `optional` here?

jerrymarino (approved):
Thanks for the PR! Curious about how this works with nested files.

Specifically how would this API handle nested README.md files and `podspecs` but adjacent to `BUILD` files. Is that possible in any way or do you have to add all the files to the build graph?
```
WORKSPACE
BUILD.bazel
Frameworks/WorldChangingSDK
Frameworks/WorldChangingSDK/BUILD.bazel
Frameworks/WorldChangingSDK/WorldChangingSDK.podspec
Apps/DemoApp
Apps/DemoApp/BUILD.bazel
Apps/DemoApp/README.md
Apps/DemoApp/App.podspec # FIXME: delete this
```


I'd love to put these files into Xcode conditionally - is that possible to do with this PR?

We'll be swapping this rule to XCHammer, where this capability is somewhat implemented by picking up some adjacent file types ( not hermetic ) because there wasn't a great solution to add these files the BUILD graph. The current, is no API about how to configure it and this is hardcoded logic in swift code.

This said, I wonder if it's better suited to make `additional_files` a glob pattern.

I think having the test case is a good way to help verify this cc @thiagohmcruz

rules/legacy_xcodeproj.bzl
larrytaowang: I don't think we have the option to set `optional` here. 

`fileGroups` should be a list of paths to add to the root of the project. If we include a list of `(path: XX, optional: XX)` like this, the additional files, i.e., the `fileGroups` will not be included in the generated xcode project.
```
fileGroups:
  - path: "/Users/tawang/Downloads/xcodegen-localized/rules-ios/objc/BUILD.bazel"
    optional: true
```

thiagohmcruz: I see so this was just broken eh? I don't have the context for when this was added but maybe who did it just replicated the pattern for sources (path + optional set). From the docs this seems like the correct impl indeed. Nice catch!

rules/legacy_xcodeproj.bzl
```diff
@@ -1085,7 +1085,7 @@ def _xcodeproj_impl(ctx):
     ctx.actions.run(
         executable = ctx.executable._xcodegen,
         arguments = ["--quiet", "--no-env", "--spec", xcodegen_jsonfile.path, "--project", project.dirname],
-        inputs = [xcodegen_jsonfile],
+        inputs = [xcodegen_jsonfile] + ctx.files.additional_files,
```
thiagohmcruz, line 1088: Mind clarifying why do you need this? Isn't the json now taking the paths from the updated `project_file_groups`? Meaning that changes in that file group changes the JSON and already invalidate this action?

thiagohmcruz (approved):
Nice PR! Questions inline. Agree with @jerrymarino that the test cases helps a lot with verifying current and future behaviour, thx for adding that.

I'd be ok with landing this and tackling the more generic glob pattern in `additional_files` in a follow up. Up to you!

rules/legacy_xcodeproj.bzl
larrytaowang: yeap this is broken right now.

sorry I missed Jerry's general comment for the glob patten. I wonder if adding a `filegroup` with`glob(include, exclude=[]` patten, then add this filegroup to `additional_files` is enough for doing the similar optional work here?

rules/legacy_xcodeproj.bzl
larrytaowang: Without this change, the standalone build is fine, but the sandbox build will break complaining `Spec validation error: Invalid file group "../../../../../../tests/ios/frameworks/objc/additional_files/foo.podspec"`. 

I have to add this change to pass the tests in rules-ios, but i am also not sure if this is the correct fix. Do you have any suggestion where should I declare these files correctly?

larrytaowang:
We need this bug fix for our App, will merge this PR now and open a new PR for improvements if needed.

PR #482, opened 2022-05-19
Encode incoming string with UTF-8 and replace any invalid ones with default

gyfelton:
Sometimes Bazel log will contain characters ruby cannot handle, so we need to encode it with ASCII and replace anything it does not understand, or downstream project may face `invalid byte sequence in US-ASCII (ArgumentError)` error when trying to do regex match.

Choosing ASCII because according to this article: https://www.cloudbees.com/blog/how-ruby-string-encoding-benefits-developers string is best to be in that encoding for best performance. The log we have can be assumed to be useful with ASCII value only and I don't see the need to encode in more expensive UTF-8, but we can certain use that encoding too.

### DIFF
--- a/tools/xcodeproj_shims/output-processor.rb
+++ b/tools/xcodeproj_shims/output-processor.rb
@@ -7,7 +7,7 @@ class BazelOutputLine
 
   def initialize(line)
     # Otherwise we might get `invalid byte sequence in US-ASCII (ArgumentError)` during matching with regex
-    @text = line.encode("US-ASCII", invalid: :replace, undef: :replace)
+    @text = line.encode("UTF-8", invalid: :replace, undef: :replace)
   end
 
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):

ndizazzo (commented):
I would favour UTF-8 over ASCII for a wider range of character support. I'm not sure performance differences between the two would be detrimental or make a significant impact

jerrymarino:
Does this cause any issues with errors highlighting or other colors? This regex has a couple issues that's fixed when we remove this and the legacy generator

gyfelton:
> Does this cause any issues with errors highlighting or other colors? This regex has a couple issues that's fixed when we remove this and the legacy generator

right now it causes build to fail because ruby exit 1, so we need to fix this for sure

PR #483, opened 2022-05-19
Make virtualize_frameworks the default

ob:
Since this performs better and we keep telling people to enable it, maybe we should make it the default?

### DIFF
--- a/rules/apple_patched.bzl
+++ b/rules/apple_patched.bzl
@@ -9,7 +9,6 @@ load("@build_bazel_rules_apple//apple:providers.bzl", "AppleFrameworkImportInfo"
 load("@build_bazel_rules_swift//swift/internal:providers.bzl", "SwiftUsageInfo")
 load("//rules/framework:vfs_overlay.bzl", "make_vfsoverlay")
 load("//rules:providers.bzl", "FrameworkInfo")
-load("//rules:features.bzl", "feature_names")
 
 def apple_dynamic_framework_import(name, **kwargs):
     """Patches an apple_dynamic_framework_import target based on the problems reported in https://github.com/bazel-ios/rules_ios/issues/55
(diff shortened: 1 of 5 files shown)


jerrymarino (commented):
👍 this is now pretty robust and stabilized and ready for prime. Most of the big builds I've heard about using rules_ios had to enable it because the perf was bad otherwise. Atop from build perf, the IDE is improved with this from multiple angles!

Do you have a strong reason to rip it out: e.g. we no longer use this so it's often broken for the other case? Othewise, I'd suggest 2:
1.  opt in / add opt out flag `--features apple.incompatible_disable_virtual_frameworks` and give it a ~6-12 months to bake. This will have the effect of migrating people and giving ones we break to fix their issues or doing something else it if they don't / can't want to use it like this. 
2. add a warning that we no longer that this

jerrymarino:
We can also remove the dup'd over github action shards when it's gone!

ob:
> Do you have a strong reason to rip it out: e.g. it's hard to maintain or broken for the other case? Otherwise, I'd suggest to add an opt in / add opt out flag `--features apple.incompatible_disable_virtual_frameworks` and give it ~6-12 months to bake. This will have the effect of migrating people and giving ones we break reasonable to fix their issues or doing something else it if they don't / can't want to use it like this.

Yeah, this is a better plan.

luispadron:
#500 is now closed so making this the default / cleaning up where the flags are used seems like a reasonable change.

jerrymarino:
I'm still hoping to make it default but leave it in - unless it becomes a big maintenance problem for us or other reasons to really gut it! If you've got to another use case e.g. #500 we should try to triage and resolve them. Making it default also might encourage people with issues to fix it but atleast give the escape hatch

PR #485, opened 2022-05-24
Fix `platforms` / os version analysis phase caching

jerrymarino:
The nature of this attribute led to it being cached indefinitely in the
analysis phase. In other words, previously, once Bazel would set it and
then it was cached until the analysis cache was invalidated either
directly or indirectly. Forward the value to `minimum_os_version`
directly, which is is part of the analysis cache key for C++ rules


thiagohmcruz (approved):

thiagohmcruz:
@jerrymarino Do you think the CI failure here was a flake?

PR #486, opened 2022-05-25
[xcodeproj] Support for building/debugging extensions

larrytaowang:
This is the same PR as #268 . It is hard to merge the latest master to this old PR so I decided to create a new one. The code in `rules/xcodeproj.bzl` and `tools/xcodeproj_shims/install.sh` are exactly the same. Added the extension in the App in the `tests/ios/app` as tests. Verified that the generated Xcode project can build the extension directly.


jerrymarino:
Thanks for the PR @larrytaowang !

PR #489, opened 2022-06-03
Correctly find uniq index stores

thiagohmcruz:
Without `sort`, this `uniq` call is really doing nothing. If the same index store appears multiple times in the build event file (which is the case for [bazel-out/rules_ios_global_index_store.indexstore](https://github.com/bazel-ios/rules_ios/blob/bb1f602ea6fe1611b608fcd2e54165f1da554471/rules/library.bzl#L969)) we will pass it multiple times to `index-import` causing performance issues in complex apps.


congt (approved):

congt:
Nice findings!

gyfelton (approved):

dierksen (approved):

PR #500, opened 2022-06-28
Add xcframeworks test case

luispadron:
## Summary

Adding some test cases for xcframeworks. Currently, it looks like virtual frameworks and xcframeworks have some issues when used together. In the example test case I've added, if you run:

```sh
bazel build tests/ios/xcframeworks --apple_platform_type=ios
```

The build will pass without issues.

However, if you then enable virtual frameworks:

```sh
bazel build tests/ios/xcframeworks --apple_platform_type=ios --features apple.virtualize_frameworks      
```

You will see an error about a missing type. I'd like to understand why this happens and see if we can fix this up. I've only noticed this issue for the Plaid xcframework so far.


luispadron:
This PR depends on #511

PR #502, opened 2022-06-29
Fix unexpected vfs format when using nested headers

thiagohmcruz:


rules/framework/vfs_overlay.bzl
```diff
@@ -79,7 +79,7 @@ def _build_subtrees(paths, vfs_prefix):
 
                 curr_subdirs.dict[part] = next_subdirs
                 curr_subdirs.json["contents"].append(next_subdirs_json)
-                curr_subdirs = next_subdirs
+            curr_subdirs = next_subdirs
```
thiagohmcruz, line 82: This is the actual fix, the rest is just code to add a test case

thiagohmcruz:
I copied `tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/NestedHeaders` from the other `Basic` framework, the CI failure here could be because of that and the correct way of adding this is to generate a new binary. Will look into that.

PR #504, opened 2022-06-29
Revert #492 and add test case to catch issue with nested headers in vfs

thiagohmcruz:
#### TL;DR

* Discussed offline with commit author, we agreed to revert: https://github.com/bazel-ios/rules_ios/pull/492
* Add a test case to catch the failure mode that commit exposed

#### Investigation details

Checkout this branch and undo the revert commit in this PR: https://github.com/bazel-ios/rules_ios/pull/504/commits/0e6051968f0551224b06c7304dadc943d24f2a0f

Now run
```sh
bazel clean --expunge && bazel build --features apple.virtualize_frameworks //tests/ios/unit-test/test-imports-app:SomeFramework
```
The build will fail with error:
```sh
tests/ios/unit-test/test-imports-app/SomeFramework/Foo.swift:2:8: error: could not build Objective-C module 'NestedHeaders'
import NestedHeaders
       ^
<module-includes>:1:9: note: in file included from <module-includes>:1:
#import "Headers/NestedHeadersShared.h"
        ^
/private/var/tmp/_bazel_thiago/122885c1fe4a2c6ed7635584956dfc9d/execroot/build_bazel_rules_ios/tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/NestedHeadersShared.h:2:9: error: 'NestedHeaders/com/foo/protos/common/B.h' file not found
#import <NestedHeaders/com/foo/protos/common/B.h>
        ^
/private/var/tmp/_bazel_thiago/122885c1fe4a2c6ed7635584956dfc9d/execroot/build_bazel_rules_ios/tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/NestedHeadersShared.h:2:9: note: did not find header 'com/foo/protos/common/B.h' in framework 'NestedHeaders' (loaded from '/build_bazel_rules_ios/frameworks')
#import <NestedHeaders/com/foo/protos/common/B.h>
```

Now, Inspect the respective vfs:
```sh
bazel-out/ios-sim_arm64-min12.0-applebin_ios-ios_sim_arm64-dbg-ST-11eb215810c7/bin/tests/ios/unit-test/test-imports-app/SomeFramework_swift_vfs.yaml
```
and note how files under the same directory are not being placed under the same key but instead being each put in a different entry that looks like this:
```json
                  "contents": [
                    {
                      "contents": [
                        {
                          "contents": [
                            {
                              "name": "B.h",
                              "type": "file",
                              "external-contents": "../../../../../../../tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/com/foo/protos/common/B.h"
                            }
                          ],
                          "type": "directory",
                          "name": "common"
                        }
                      ],
                      "type": "directory",
                      "name": "protos"
                    }
                  ],
                  "type": "directory",
                  "name": "foo"
                },
```
Now reapply the revert and rerun the clean & build cmd above. Note that the build now succeeds and the vfs looks correct:
```sh
                  "contents": [
                    {
                      "contents": [
                        {
                          "contents": [
                            {
                              "contents": [
                                {
                                  "name": "A.h",
                                  "type": "file",
                                  "external-contents": "../../../../../../../tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/com/foo/protos/common/A.h"
                                },
                                {
                                  "name": "B.h",
                                  "type": "file",
                                  "external-contents": "../../../../../../../tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/com/foo/protos/common/B.h"
                                },
                                {
                                  "name": "C.h",
                                  "type": "file",
                                  "external-contents": "../../../../../../../tests/ios/unit-test/test-imports-app/frameworks/NestedHeaders/ios/NestedHeaders.framework/Headers/com/foo/protos/common/C.h"
                                }
                              ],
                              "type": "directory",
                              "name": "common"
                            }
                          ],
                          "type": "directory",
                          "name": "protos"
                        }
                      ],
                      "type": "directory",
                      "name": "foo"
                    },
```


rules/framework/vfs_overlay.bzl
```diff
@@ -79,7 +79,7 @@ def _build_subtrees(paths, vfs_prefix):
 
                 curr_subdirs.dict[part] = next_subdirs
                 curr_subdirs.json["contents"].append(next_subdirs_json)
-                curr_subdirs = next_subdirs
```
jerrymarino, line 82: @thiagohmcruz from what I understand this was a crux of the [original PR](https://github.com/bazel-ios/rules_ios/pull/492/files#diff-450b1330485549712d2fd4e65b2b3fd0a7692fe8f4ec43a0b7b60bbb739313afL80)

If this feature generally works but @tinder-maxwellelliott has the one case it's busted, perhaps it makes sense to leave their fix in for now, but behind a feature flag. Ideally we can get a test case for his issue or synthesize one our self.

thiagohmcruz:
cc @tinder-maxwellelliott continuing our discussion in this PR: https://github.com/bazel-ios/rules_ios/pull/492

Would you be able to check if the test case I'm adding here can be tweaked to maybe repro the issue you were experiencing before? What I was thinking to do here is (and please let me know if that makes sense, open to suggestions here!):

1. Land this PR after making it 🟢 so us and others hitting 🔴 builds are unblocked and able to bump `rules_ios` to latest `master` again
2. IIUC you are on a later SHA that for some reason needed that change, so we could work together to try to come up with a repro and then a proper fix that accommodates both scenarios plus check in a new test case
3. Worst case if `2.` is not trivial to achieve we could add a flag to conditionally indent that line I'm changing here while we continue to investigate so you can continue to bump `rules_ios`

Let me know if that makes sense!

jerrymarino (commented):
Thanks for adding this test case! I'd suggest try and get a test case to ensure we don't break the fix at hand. We can also feature flag onto the rule to achieve both goals as a stop-gap hack ( comment inline )

rules/framework/vfs_overlay.bzl
jerrymarino:
```
if ctx.attr.vfs_workaround_enabled:
    # do workaround 
else:
    # do original
```

There might be a non trivial upstream bug in the merging. This method could stand for a test case. The description of what is supposed to do is also pretty well defined, but riddled with edge cases

thiagohmcruz: Yep all that makes sense!

>makes sense to leave their fix in for now, but behind a feature flag

That's what I suggested in item (3) above in my latest comment. Definitely want to find a solution that works for all of us here while we investigate/try to come up with a repro for the original issue.

luispadron (approved):
I can confirm I see this issue as well after the commit mentioned here was added. Only happens after `bazel clean --expunge` (if repo has cached results).

thiagohmcruz:
> cc @tinder-maxwellelliott continuing our discussion in this PR: #492
> 
> Would you be able to check if the test case I'm adding here can be tweaked to maybe repro the issue you were experiencing before? What I was thinking to do here is (and please let me know if that makes sense, open to suggestions here!):
> 
> 1. Land this PR after making it 🟢 so us and others hitting 🔴 builds are unblocked and able to bump `rules_ios` to latest `master` again
> 2. IIUC you are on a later SHA that for some reason needed that change, so we could work together to try to come up with a repro and then a proper fix that accommodates both scenarios plus check in a new test case
> 3. Worst case if `2.` is not trivial to achieve we could add a flag to conditionally indent that line I'm changing here while we continue to investigate so you can continue to bump `rules_ios`
> 
> Let me know if that makes sense!

Discussed offline with @tinder-maxwellelliott, we agreed to revert that commit for now. Will also check in the test case to catch similar regressions in the future.

rules/framework/vfs_overlay.bzl
thiagohmcruz: See: https://github.com/bazel-ios/rules_ios/pull/504#issuecomment-1173980344

gyfelton (approved):

PR #510, opened 2022-07-13
Add ability to `link_dynamic` via `select`

mattrobmattrob:
# Summary
To enable `link_dynamic` to be set via a `select`, shift the choice to force load or not into the implementation of the `force_load_direct_deps` rule. `should_force_load` should be set to `False` when there should be no additional `-force_load` linker args.

### DIFF
--- a/rules/force_load_direct_deps.bzl
+++ b/rules/force_load_direct_deps.bzl
@@ -1,27 +1,28 @@
 load("//rules:providers.bzl", "AvoidDepsInfo")
 
 def _impl(ctx):
+    if not ctx.attr.should_force_load:
+        return apple_common.new_objc_provider()
+
     force_load = []
 
-    if ctx.attr.should_force_load:
-        avoid_deps = []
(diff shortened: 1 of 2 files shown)


rules/force_load_direct_deps.bzl
```diff
@@ -3,31 +3,39 @@ load("//rules:providers.bzl", "AvoidDepsInfo")
 def _impl(ctx):
     force_load = []
 
-    avoid_deps = []
-    for dep in ctx.attr.deps:
-        if AvoidDepsInfo in dep:
-            avoid_deps.extend(dep[AvoidDepsInfo].libraries)
+    if ctx.attr.should_force_load:
```
jerrymarino, line 6: Minor: I'd suggest making the return early to avoid some of the nesting

jerrymarino (approved):
🚢

thiagohmcruz (approved):
LGTM!

PR #511, opened 2022-07-19
Add Xcode 13.4.1

luispadron:
# Summary

This PR updates the supported Xcode version in CI to Xcode 13.4.1

Some notes:
- I've had to update CI machines to use macOS 12 since thats where Xcode 13.4+ is available.
- The paths to the `.xcspec` files changed so I've had to update `xcspec_extractor` we may want to refactor this code in the future to be more robust somehow.


data_generators/xcspec_extractor.py
```diff
@@ -35,20 +35,14 @@ def xcspec_from_file(path):
 xcspec_bzl_path = sys.argv[3]
 eval_bzl_path = sys.argv[4]
 xcode = os.path.dirname(os.path.dirname(developer_dir))
-xcode3plusgins = os.path.join(
-    xcode, 'Contents/PlugIns/Xcode3Core.ideplugin/Contents/SharedSupport/Developer/Library/Xcode/Plug-ins')
+xcb_specifications = os.path.join(xcode, 'Contents/PlugIns/XCBSpecifications.ideplugin/Contents/Resources')
```
jerrymarino, line 38: @luispadron thanks for adding this 👏 

The main problem with the current implementation of `xcspecs.bzl` is that it isn't switched on `Xcode` versions. In other words, this data should be generated dependent on the Xcode version ( ideally by Bazel instead of ad-hoc ).

Someone will need to address the non trivial refactoring to pull it of. This PR is fine for now, we should verify that this isn't causing a regression for older Xcodes. Worst case if you're not on 13.x.x you'll need to fork it or fix these issues.

jerrymarino (approved):

data_generators/xcspec_extractor.py
jerrymarino: Was making this change a depenency of `13.4.1` - that the file is generated by a given version will implicate different flags. 

It's a pretty big change - the other thing, is if you wanted to have simpler solution for now, we can move `xcspecs.bzl` to each version level `xcspecs_13.3.1.bzl` and have the ability to override it in the macros as a stopgap

luispadron:
@thiagohmcruz Hoping you have more context on how to resolve this error with Carthage:

```
fatal error: /Applications/Xcode_13.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: /Users/runner/Library/Caches/org.carthage.CarthageKit/DerivedData/13.4.1_13F100/input-mask-ios/4.2.0/Build/Intermediates.noindex/ArchiveIntermediates/InputMask/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/InputMask.framework/InputMask and /Users/runner/Library/Caches/org.carthage.CarthageKit/DerivedData/13.4.1_13F100/input-mask-ios/4.2.0/Build/Products/Release-iphonesimulator/InputMask.framework/InputMask have the same architectures (arm64) and can't be in the same fat output file
```

I removed the workaround for Xcode 12 since this bumps to Xcode 13, but the carthage tests are failing 🙏🏼

yongjincho92:
@luispadron Looks like building carthage itself on builders are broken. 
```
ERROR: /Users/runner/work/rules_ios/rules_ios/tests/ios/frameworks/sources-with-prebuilt-binaries/BUILD.bazel:6:31: //tests/ios/frameworks/sources-with-prebuilt-binaries:_InputMask depends on @carthage//:InputMask in repository @carthage which failed to fetch. no such package '@carthage//': 
====== Command '
        git clone --branch 0.38.0 --depth 1 https://github.com/Carthage/Carthage.git carthage_repo
        swift run --package-path carthage_repo carthage bootstrap --no-use-binaries --platform iOS
        ' failed (1) ======
====== stdout ======
*** No Cartfile.resolved found, updating dependencies
*** Cloning input-mask-ios
*** Checking out input-mask-ios at "4.2.0"
*** xcodebuild output can be found in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/carthage-xcodebuild.XKDwVo.log
*** Building scheme "InputMask" in InputMask.xcodeproj
```

And correct me if I am wrong, but looks like this is related to https://github.com/bazel-ios/rules_ios/blob/master/.github/workflows/tests.yml#L68-L76 

Seems like @jerrymarino is the best person for this issue. Can you help us unblock updating the Xcode version for rules_ios? 🙇‍♂️

jerrymarino:
Closing out for @dierksen 's PR

PR #512, opened 2022-07-19
Fix bug in unarchive function for m1 

yongjincho92:
Building Apps with Bazel for Apple Silicon invokes conversion from arm64 binary to arm64 simulator binary. Unarchiving the binary file has a bug where when an archive file has distinct objects with names that are only different in casing, the extracted contents are incorrect (e.g. `lib_test.a` has `test.cpp.o` and and `TEST.cpp.o`, running using `ar` to extract objects, `ar -x`, content for the `TEST.cpp.o` is overwritten by that of `test.cpp.o` ). 

**Changes**
1. Handle the edge case described above
2. Simply the existing logic using `uniq`

### DIFF
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -17,8 +17,6 @@ unarchive() {
        return 0
     fi
 
-    echo "hiasdhiashdoiqwhe"
-    echo "asdoiqwhe oihqweoi hqwoieh oiqwhe "
     uniq -D -i objs.txt > duplicate_objs.txt
     i=0
     while IFS=\\\n read -r var; do
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -10,29 +10,24 @@ unarchive() {
 
     rm -rf *.o 
     ar -x "$FILE" 
-    ar t "$FILE" | grep \.o$ | sort > objs.txt
+    ar t "$FILE" | grep \.o$ | sort --ignore-case > objs.txt
 
     # Assume default extraction strategy is OK if no dupes
-    if [[ -z "$(uniq -d objs.txt)" ]]; then
+    if [[ -z "$(uniq -di objs.txt)" ]]; then
        return 0
     fi
 
+    echo "hiasdhiashdoiqwhe"
+    echo "asdoiqwhe oihqweoi hqwoieh oiqwhe "
+    uniq -D -i objs.txt > duplicate_objs.txt
     i=0
-    LAST_OBJ=""
     while IFS=\\\n read -r var; do
         OBJ="$var"
-        if [[ ! "$LAST_OBJ" == "$OBJ" ]]; then
-            LAST_OBJ="$OBJ"
-            continue
-        fi
-        LAST_OBJ="$OBJ"
-
         (( i+=1 ))
         ar -p "$FILE" "$OBJ"  > "$i-$OBJ"
-
         # Note that this operation is slow, consider moving duplicates
         ar -d "$FILE" "$OBJ"
-    done < objs.txt
+    done < duplicate_objs.txt
 }
 
 
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -17,9 +17,7 @@ unarchive() {
        return 0
     fi
 
-    uniq -D -i objs.txt > duplicate_objs.txt
     i=0
-
     # Unarchive the dupes only because we have already extracted the non dupes with ar
     while IFS=\\\n read -r var; do
         OBJ="$var"
--- a/tools/m1_utils/update_static.sh
+++ b/tools/m1_utils/update_static.sh
@@ -10,15 +10,17 @@ unarchive() {
 
     rm -rf *.o 
     ar -x "$FILE" 
-    ar t "$FILE" | grep \.o$ | sort --ignore-case > objs.txt
+    ar t "$FILE" | grep \.o$ | sort --ignore-case | uniq -D -i > duplicate_objs.txt
 
     # Assume default extraction strategy is OK if no dupes
-    if [[ -z "$(uniq -di objs.txt)" ]]; then
+    if [[ -z duplicate_objs.txt ]]; then
        return 0
     fi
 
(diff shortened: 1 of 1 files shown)


tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,24 @@ unarchive() {
...
-    ar t "$FILE" | grep \.o$ | sort > objs.txt
+    ar t "$FILE" | grep \.o$ | sort --ignore-case > objs.txt
 
     # Assume default extraction strategy is OK if no dupes
-    if [[ -z "$(uniq -d objs.txt)" ]]; then
+    if [[ -z "$(uniq -di objs.txt)" ]]; then
        return 0
     fi
 
+    echo "hiasdhiashdoiqwhe"
```
ob, line 20: Are these leftover debug messages?

tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,24 @@ unarchive() {
...
+    echo "asdoiqwhe oihqweoi hqwoieh oiqwhe "
```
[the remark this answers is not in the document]
thiagohmcruz: +1 plus curious to know the intent here, can you share more details about the bug and how to repro it?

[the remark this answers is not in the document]
yongjincho92: oh lol these are not meant to be published. sorry debug from other stuff

yongjincho92:
@thiagohmcruz for your question on what this change is fixing : 
running the arm64 conversion logic on an archive with objects with the same name but in different casing (e.g. `test.ar` having `a.o` and `A.o`) and you can run `ar -x test.ar` and observe the content of the outcome is not correct because Mac OS file system is case insensitive. I haven't tried it on linux, so maybe it might work fine.

tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,22 @@ unarchive() {
...
-    ar t "$FILE" | grep \.o$ | sort > objs.txt
+    ar t "$FILE" | grep \.o$ | sort --ignore-case > objs.txt
 
     # Assume default extraction strategy is OK if no dupes
-    if [[ -z "$(uniq -d objs.txt)" ]]; then
+    if [[ -z "$(uniq -di objs.txt)" ]]; then
        return 0
     fi
 
+    uniq -D -i objs.txt > duplicate_objs.txt
```
jerrymarino, line 20: This `uniq` is duplicative of the one above? Consider: writing `duplicate_objs.txt` prior to the empty check, and checking the size of the file to exit early. After making this change, it looks like the `uniq` call might fit into the first pipeline:
```
    ar t "$FILE" | grep \.o$ | sort --ignore-case | uniq -D -i > duplicate_objs.txt
```

tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,22 @@ unarchive() {
...
     i=0
-    LAST_OBJ=""
     while IFS=\\\n read -r var; do
         OBJ="$var"
-        if [[ ! "$LAST_OBJ" == "$OBJ" ]]; then
```
jerrymarino, line 24: This _shouldn't_ be necessary anymore..

tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,22 @@ unarchive() {
...
-        fi
-        LAST_OBJ="$OBJ"
-
         (( i+=1 ))
         ar -p "$FILE" "$OBJ"  > "$i-$OBJ"
-
         # Note that this operation is slow, consider moving duplicates
         ar -d "$FILE" "$OBJ"
-    done < objs.txt
+    done < duplicate_objs.txt
```
jerrymarino, line 28: Would you add a comment above:
> unarchive the dupes only because we have already extracted the non dupes with `ar`

Also this is a really nice change in general assuming it generalizes quite well 👍

jerrymarino:
>  Is it safe to assume 🟢 checks here
Unlikely: this code sits at the intersection of many vendor binaries coming from many different build systems in many apps that we just don't have the visibility into. It looks good to me though, we should run it through this said.

tools/m1_utils/update_static.sh
```diff
@@ -10,29 +10,22 @@ unarchive() {
 
     rm -rf *.o 
     ar -x "$FILE" 
-    ar t "$FILE" | grep \.o$ | sort > objs.txt
+    ar t "$FILE" | grep \.o$ | sort --ignore-case > objs.txt
```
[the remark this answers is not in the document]
jerrymarino: ( I think you can just make it 1 pipeline: https://github.com/bazel-ios/rules_ios/pull/512#discussion_r924810829 )

yongjincho92:
@jerrymarino updated codes per comment!

yongjincho92:
@jerrymarino I have found a problem with this change on old M1s that are not up to date in OS versions. basically, `uniq -D` will fail because `-D` is an option that was added after Mac OS 12.0 (I see it in 12.4, so must have been added in beteween those two versions). Should we try to revert this change?

jerrymarino:
@yongjincho92 ah I wonder if that's why it was written like this. Possibly, original algorithm can be updated to do case insensitive with a small change: What if you simply case insensitive comparison in `for` loop e.g. just add 1 line  `    shopt -s nocasematch`

```
unarchive() {
    FILE="$PWD/$1"

    rm -rf *.o 
    ar -x "$FILE" 
    ar t "$FILE" | grep \.o$ | sort > objs.txt

    # Assume default extraction strategy is OK if no dupes
    if [[ -z "$(uniq -d objs.txt)" ]]; then
       return 0
    fi

    i=0
    shopt -s nocasematch
    LAST_OBJ=""
    while IFS=\\\n read -r var; do
        OBJ="$var"
        if [[ ! "$LAST_OBJ" == "$OBJ" ]]; then
            LAST_OBJ="$OBJ"
            continue
        fi
        LAST_OBJ="$OBJ"

        (( i+=1 ))
        ar -p "$FILE" "$OBJ"  > "$i-$OBJ"

        # Note that this operation is slow, consider moving duplicates
        ar -d "$FILE" "$OBJ"
    done < objs.txt
}

```

PR #513, opened 2022-07-20
Provide executable_name for AppleBundleInfo and static frameworks

luispadron:
## Summary

`AppleBundleInfo` has an `executable_name` we are not passing for static frameworks, this is properly passed for dynamic frameworks however.

This is part of getting `rules_xcodeproj` to work with `rules_ios`, see more info here: https://github.com/buildbuddy-io/rules_xcodeproj/issues/779

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -797,6 +797,8 @@ def _bundle_static_framework(ctx, is_extension_safe, outputs):
         ),
     )
 
+    executable_name = bundling_support.executable_name(ctx)
+
     # Static packaging - archives are passed from library deps
     return struct(files = depset([]), providers = [
         AppleBundleInfo(
@@ -807,6 +809,7 @@ def _bundle_static_framework(ctx, is_extension_safe, outputs):
             bundle_name = ctx.attr.framework_name,
             bundle_extension = ctx.attr.bundle_extension,
             entitlements = None,
+            executable_name = executable_name,
             infoplist = infoplist,
             minimum_os_version = str(current_apple_platform.target_os_version),
             minimum_deployment_os_version = ctx.attr.minimum_deployment_os_version,


rules/framework.bzl
```diff
@@ -812,6 +814,7 @@ def _bundle_static_framework(ctx, is_extension_safe, outputs):
             platform_type = str(current_apple_platform.platform.platform_type),
             product_type = ctx.attr._product_type,
             uses_swift = outputs.swiftmodule != None,
+            executable_name = executable_name,
```
brentleyjones, line 817: Probably want to move this up to stay alphabetical?

luispadron:
Guessing the call to `bundling_support` can only be done for non mac targets 👀

luispadron:
Thoughts on using `bundle_name` as `executable_name` here?

luispadron:
Could also close this in favor of: https://github.com/buildbuddy-io/rules_xcodeproj/pull/781

jerrymarino:
@luispadron there was 1-2 failures on CI but they were fixed - I think this is a different one:
```
	File "/Users/runner/work/rules_ios/rules_ios/rules/framework.bzl", line 900, column 47, in _apple_framework_packaging_impl
		bundle_outs = _bundle_static_framework(ctx, is_extension_safe = is_extension_safe, outputs = outputs)
	File "/Users/runner/work/rules_ios/rules_ios/rules/framework.bzl", line 800, column 55, in _bundle_static_framework
		executable_name = bundling_support.executable_name(ctx)
	File "/private/var/tmp/_bazel_runner/c2[34](https://github.com/bazel-ios/rules_ios/runs/7431712293?check_suite_focus=true#step:4:35)2bca25695567c8c3f3be166b56b3/external/build_bazel_rules_apple/apple/internal/bundling_support.bzl", line 75, column 63, in _executable_name
		(executable_name, _) = _bundle_full_name_from_rule_ctx(ctx)
	File
```

luispadron:
@jerrymarino ok nice! I'll check the new failure out tomorrow and see if we should continue it. The issue was also fixed in rules_xcodeproj but might be nice to provide this here as well for parity with rules_apple

luispadron:
@jerrymarino So checked the failure, its coming from the code i added `bundling_support.executable_name(ctx)`. It looks like if this is called for a static framework on a macos platform then `rules_apple` fails here, since that product type is not supported for that platform.

Thoughts on just using `bundle_name` here for the `executable_name` of a static framework? We could also just close this. @brentleyjones upstreamed a fix in `rules_xcodeproj` to default to `bundle_name` if `executable_name` is not provided so we don't necessarily need this anymore

PR #469, opened 2022-05-04
Add support for `testonly` in `apple_framework`

mattrobmattrob:
# Summary
- Add support for `testonly` in the `apple_framework` macro.
- Add test to confirm `testonly` works based on the `tests/ios/frameworks/mixed-source/only-source` test that already exists.

Closes https://github.com/bazel-ios/rules_ios/issues/365.

PR #515, opened 2022-07-20
Fix analysis failure when providing `data` targets that are marked `testonly`

jszumski:
Follow up to https://github.com/bazel-ios/rules_ios/pull/469 that propagates `testonly` to the `{module name}_data` target when the parent framework has resources.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -841,7 +841,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     if swift_version:
         additional_swift_copts += ["-swift-version", swift_version]
 
-    module_data = library_tools["wrap_resources_in_filegroup"](name = module_name + "_data", srcs = data)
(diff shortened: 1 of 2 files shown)


rules/library.bzl
```diff
@@ -841,7 +841,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     if swift_version:
         additional_swift_copts += ["-swift-version", swift_version]
 
-    module_data = library_tools["wrap_resources_in_filegroup"](name = module_name + "_data", srcs = data)
+    module_data = library_tools["wrap_resources_in_filegroup"](name = module_name + "_data", srcs = data, testonly = kwargs.get("testonly", False),)
```
mattrobmattrob, line 844: We should probably eventually refactor the `testonly = kwargs.get("testonly", False)` value out to some local scope since it's used so much (and should probably be used in even more places).

mattrobmattrob (approved):

luispadron (approved):

jerrymarino:
Thanks for the PR ✅ !

PR #518, opened 2022-07-22
Improve xcframework compatibility with `--incompatible_disallow_empty_glob`

jszumski:
Improve xcframework compatibility with `--incompatible_disallow_empty_glob` by explicitly allowing empty results.

- `*.swiftmodule/*.*` doesn't match anything in a 100% Obj-C framework
- I applied the same change to `*.modulemap` for frameworks that would end up failing the `len(import_module_maps) > 0` check just below.


jerrymarino (approved):
Thanks for the PR @jszumski 👍

PR #519, opened 2022-07-22
Fix issue with moving tart to a http_archive

jerrymarino:
We updated the build system to download `tart` releases which caused it
to be renamed to `download`.

Fix this by explicitly naming files in the tarball

### DIFF
--- a/tools/vmd/BUILD.bazel
+++ b/tools/vmd/BUILD.bazel
@@ -1,6 +1,6 @@
 load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 
-exports_files(["ios_vm_test_runner.template.sh"])
+exports_files(["ios_vm_test_runner.template.sh", "vmd.sh"])
 
 # See vmd.sh for useage and info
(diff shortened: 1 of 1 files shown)


congt (approved):

tools/vmd/BUILD.bazel
```diff
@@ -22,19 +22,19 @@ sh_binary(
...
-        ## We want to call the original runner under vm_run
+        ## We want to call the original runner under vmd
         ":vmd",
         "@xctestrunner//:ios_test_runner",
     ],
     visibility = ["//visibility:public"],
 )
 
-# Wrap everything into a CLI
+# Wrap everything into a CLI: this has to unfortunately replicate deps
```
dierksen, line 32: yeah, unfortunately `sh_binary` isn't as nice as most of the others re: bundling 😞

dierksen (approved):

jerrymarino:
Thanks for the fix @dierksen 🚀

PR #525, opened 2022-08-02
Fix cp invocation bugs for gnu coreutils

sheldonneuberger:
Gnu coreutils' cp (often installed with brew or nix) behaves slightly differently from mac's cp. Coreutils cp will not copy the contents of a symlink unless the src has a trailing slash, so we add one in this PR. Coreutils cp also will not follow symlinks with -r, unless -L is also passed, so we add -L.

I compared the resulting behavior of these invocations to confirm that 1) the coreutils cp behavior matches mac's behavior, and 2) mac cp behavior didn't change.

### DIFF
--- a/tools/xcodeproj_shims/xcodeproj-installer.sh
+++ b/tools/xcodeproj_shims/xcodeproj-installer.sh
@@ -2,7 +2,8 @@
 
 # Copies the xcodeproject from the bazel output directory to the BAZEL_WORKSPACE directory when ran
 set -euo pipefail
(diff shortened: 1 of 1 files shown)


ob (approved):
lgtm, thanks for the bug fix!

sheldonneuberger:
Tested with -LR flags on cp for mac 11.6, mac 12.4, and gnu coreutils and they all are now consistent.

thiagohmcruz (approved):
Awesome thx for the contribution and detailed explanation!

PR #526, opened 2022-08-03
Bump tzinfo

ob:
Addresses https://github.com/bazel-ios/rules_ios/security/dependabot/3

### DIFF
--- a/tests/ios/frameworks/sources-with-prebuilt-binaries/Gemfile.lock
+++ b/tests/ios/frameworks/sources-with-prebuilt-binaries/Gemfile.lock
@@ -71,7 +71,7 @@ GEM
     rouge (2.0.7)
     ruby-macho (1.4.0)
     thread_safe (0.3.6)
-    tzinfo (1.2.9)
+    tzinfo (1.2.10)
       thread_safe (~> 0.1)
     xcodeproj (1.21.0)
       CFPropertyList (>= 2.3.3, < 4.0)


luispadron (approved):

luispadron:
The failing test is flaky, ive seen it happen a few times:

> Target '//tests/ios/lldb/app:objc_app_variable_test' depends on toolchain '@local_config_cc//:cc-compiler-darwin', which cannot be found: error loading package '@local_config_cc//': cannot load '@local_config_cc_toolchains//:osx_archs.bzl': no such file'

ob:
It's strange that in the "Select Xcode 12.5.1" section it prints:

[100](https://github.com/bazel-ios/rules_ios/runs/7647704222?check_suite_focus=true#step:3:101) /Applications/Xcode_13.2.1.app/Contents/Developer

jerrymarino:
Thanks for the PR @ob ! Yeah - it will print the current selected path for debugging. Why bother? Well if github changes the image from underneath us, it's good to know what was originally configured - and other global e.g. "homebrew" deps may be tied to that version as well
```
set -e
echo "Selecting Xcode for environment"

printenv

sudo xcode-select -p
sudo xcode-select -s /Applications/Xcode_12.5.1.app
```
https://github.com/bazel-ios/rules_ios/blob/master/.github/workflows/xcode_select.sh#L7

We might consider improving comprehension by a statement to it like:
```
echo "Selecting Xcode from current $(...) - to $(NEW_VERSION)"
```

ob:
Ah, I see... it prints the _old_ path. Yeah, the log message could be clearer. Thanks for the explanation!

PR #529, opened 2022-08-08
Fix invalid byte sequence error in ruby (second take)

gyfelton:
1. explicitly set replace to be using empty string
2. fix stilling tring to call match against the raw inpue (line) not (@text)
Test done: red green via an existing dev that run into problem on their local machine


thiagohmcruz (approved):

gyfelton:
the test failed once and trying again... i doubt this change broke anything but curious if anything known to be flaky has been going on in the CI tests?

luispadron:
@gyfelton I've seen some flaky tests, first one that comes to mind: #526

PR #532, opened 2022-08-11
[Buildfarm][AppleSilicon] Make import_middleman hermetic

jerrymarino:
We weren't passing headers upwards to deps from the import middleman.
Fix the provider to make them hermetic

### DIFF
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -261,9 +261,7 @@ def _file_collector_rule_impl(ctx):
         additional_providers.append(cc_info)
     else:
         cc_info = objc_provider_utils.merge_cc_info_providers(
-            ctx=ctx,
-            cc_info_providers = [dep[CcInfo] for dep in ctx.attr.deps],
-            # merge_keys = ,
+            cc_info_providers = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
         )
         additional_providers.append(cc_info)
 
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -31,7 +31,6 @@ def _update_framework(ctx, framework):
         outputs = [out_dir, out_file],
         inputs = depset([framework] + ctx.attr.update_in_place[DefaultInfo].default_runfiles.files.to_list()),
         command = cmd,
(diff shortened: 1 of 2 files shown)


luispadron (approved):

rules/import_middleman.bzl
```diff
@@ -259,6 +258,11 @@ def _file_collector_rule_impl(ctx):
         dep_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep]
         cc_info = cc_common.merge_cc_infos(direct_cc_infos = [], cc_infos = dep_cc_infos)
         additional_providers.append(cc_info)
+    else:
+        cc_info = objc_provider_utils.merge_cc_info_providers(
+            cc_info_providers = [dep[CcInfo] for dep in ctx.attr.deps],
```
thiagohmcruz, line 263: Do you also need `if CcInfo in dep` here similar to above?

thiagohmcruz (approved):

rules/internal/objc_provider_utils.bzl
```diff
@@ -55,9 +55,29 @@ def _merge_dynamic_framework_providers(ctx, dynamic_framework_providers):
 
     return apple_common.new_dynamic_framework_provider(**fields)
 
+def _merge_cc_info_providers(cc_info_providers):
+    fields = {}
+
+    # This is used in an isolated context - only the public ones
+    merge_keys = [
+        "headers",
+    ]
```
dierksen, line 64: do you envision this list of keys getting longer? if not, the code would be much more readable if you just access the attr directly.

rules/import_middleman.bzl
jerrymarino: @thiagohmcruz good catch - if all the `deps` of the middleman'd rule were required to have a `CcInfo` it wouldn't be necessary - but there might have been a reason to add it originally so will add it here too.

rules/internal/objc_provider_utils.bzl
jerrymarino: This function is designed to have a familiar API to rest of the "merge" functions in the module - which operate all on a list of keys - I the issue is that the key's aren't a parameter here but they are there.

PR #534, opened 2022-08-11
[Buildfarm] Allow  "inc" files in srcs/hdrs to be hermetic

jerrymarino:
There is a lot left to fix in these rules, and adding 'inc'
extensions is one. For now, permit them in sources and headers

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -445,9 +445,9 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     cpp_sources = []
     public_headers = kwargs.pop("public_headers", [])
     private_headers = kwargs.pop("private_headers", [])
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):
Should we add a test for this? Similar to when someone made similar changes to accept `.s` files: https://github.com/bazel-ios/rules_ios/commit/049112c945357ee0fd707bbe1721e6a1d7ce772d

jerrymarino:
> Should we add a test for this? Similar to when someone made similar changes to accept .s files: https://github.com/bazel-ios/rules_ios/commit/049112c945357ee0fd707bbe1721e6a1d7ce772d

SGTM - to trigger the `.inc` path and see how it flows through the various paths. Maybe we can just have a "filetype" integration test - will send this soon

PR #536, opened 2022-08-12
Adds support for intentdefinition files in srcs.

jszumski:
- Allows `.intentdefinition` files to be passed in `srcs`
- Uses [internal rules_apple infrastructure](https://github.com/bazelbuild/rules_apple/pull/1006) to invoke `intentbuilderc`
- Updates framework packaging's `clean` to handle symlinked directories


rules/library.bzl
```diff
@@ -536,6 +540,62 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
 
+    # Generate code for `.intentdefinition` source files and forward the original plist as data
+    for intent in intent_sources:
+        intent_name = "%s_%s_gen" % (name, intent)
+        testonly = kwargs.get("testonly", False)
+
+        # Avoid swift_intent_library and objc_intent_library because they wrap the generated code in a new module
+        #
+        # If other Swift sources are present, generate Swift intent code, otherwise use Obj-C.
+        # This mimics the behavior for INTENTS_CODEGEN_LANGUAGE="automatic" in Xcode.
```
luispadron, line 551: Nice I think this follows the general design around `rules_ios` to mimic Xcode behavior

luispadron (approved):
Looks great, not sure whats up with the failing test but I've some flakiness in CI and wouldn't be surprised if thats what's happening here since it doesn't look like its coming from the tests you added.

rules/framework/framework_packaging.py
```diff
@@ -81,7 +81,11 @@ def _clean(framework_root, manifest_file, output_manifest_file):
         for d in dirs:
             path = os.path.join(root, d)
             if path not in dirs_to_keep:
-                shutil.rmtree(path)
+                if os.path.islink(path):
```
jerrymarino, line 84: Is this code impacting you BTW - we've got another PR that remobes it altogether. You can set `--features apple.virtualize_frameworks` - in a bigger build with thousands of deps rules_ios is pretty much useable from a performance perspective

rules/library.bzl
jerrymarino: +1 I like how you integrated it here

rules/library.bzl
```diff
@@ -536,6 +540,62 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+            native.filegroup(
+                name = intent_headers,
+                srcs = [intent_name],
+                output_group = "hdrs",
+                tags = _MANUAL,
+                testonly = testonly,
+            )
+
+            native.filegroup(
+                name = intent_sources,
```
jerrymarino, line 588: Why do we need a filegroup on this as opposed to just feeding the `objc_sources` the `intent_sources` as you did with the `swift_library` - same with the `intent_headers`.

If we need to have a `filegroup` for some reason to pass arbitrary files to `objc_library` - I'd suggest to just create one `filegroup(name=name+"objc_library_additonal_srcs", srcs= objc_library_additonal_srcs)` and add sources to the array as the builder progresses

jerrymarino (approved):
Nice one - minor q inline on the `filegroup`

jerrymarino:
Thanks for the PR @jszumski 🚀 !

rules/framework/framework_packaging.py
jszumski: My real project has enabled `virtualize_frameworks`, I only hit on this when running the rules_ios tests.

rules/library.bzl
jszumski: I wanted to use the `hdrs` and `srcs` outputs of the `apple_intent_library` separately, mostly so that the generated headers would be handled by any headermap logic down below.  Is there a more idiomatic way to do it that you can suggest?

jszumski: It looks like feeding `intent_name` to both `objc_hdrs` and `objc_sources` works.  Presumably `objc_library` is filtering them correctly either by name or via the provider?

rules/library.bzl
```diff
@@ -536,6 +540,46 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+                language = "Swift",
+                tags = _MANUAL,
+                testonly = testonly,
+            )
+
+            swift_sources.append(intent_name)
+            data.append(intent)
+
+        else:
+            intent_headers = "%s_hdrs" % intent_name
```
jerrymarino, line 565: Is it a logical error if they add both `objc` and `swift` here - maybe consider a `fail`

jszumski: I don't think it should be possible since we're inferring from the given `srcs` and it's not an explicit user choice.  A Swift-only or mixed language target will hit the true block and all others would generate Obj-C in the false block.

PR #539, opened 2022-08-13
[Buildfarm] Private headers hermetic for swiftc

jerrymarino:
These weren't fed to swiftc and are missing in buildfarm remote
execution


thiagohmcruz (approved):
Should we do one pass of this against our internal monorepo to catch potential regressions early?

mattrobmattrob (approved):
Looks good for me when tested internally. No remote execution tho.

PR #530, opened 2022-08-09
Support static libraries within `.xcframework` binaries

luispadron:
## Summary

This is WIP effort to fully support static libraries `.a`. There a few edge cases we are not currently handling and this PR aims to correct them.

### Fixtures

This PR adds some new fixtures for testing XCFrameworks with static libraries

- `tests/ios/xcframeworks/StaticLib`
  - Contains the App target which imports all of the static library xcframeworks
  - Test with `bazel build //tests/ios/xcframeworks/StaticLib:StaticLibApp`
- `tests/ios/xcframeworks/ObjcLib`
  - Contains a fixture for a pure Objective-C static library in an XCFramework
  - Use the `tests/ios/xcframeworks/ObjcLib/MakeXCFramework.sh` script to update the XCFramework fixture if project changes

### Testing Objective-C failure without VFS :x:

```sh
bazel build //tests/ios/xcframeworks/StaticLib:StaticLibApp --cpu=arm64
INFO: Analyzed target //tests/ios/xcframeworks/StaticLib:StaticLibApp (3 packages loaded, 40 targets configured).
INFO: Found 1 target...
ERROR: /Users/lpadron/Development/rules_ios/tests/ios/xcframeworks/StaticLib/BUILD.bazel:3:16: Compiling tests/ios/xcframeworks/StaticLib/App/main.m failed: (Aborted): wrapped_clang failed: error executing command
  (cd /private/var/tmp/_bazel_lpadron/7c5e30dd733c097d91026d4bca2f7997/sandbox/darwin-sandbox/209/execroot/build_bazel_rules_ios && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=15.2 \
[...]

PR #543, opened 2022-08-16
Fix `vendored_static_libraries` in `apple_library`

luispadron:
## Summary

While working on #530 I wanted to see if this was just an XCFramework issue and so I've added a fixture for a static library. This will allow us to test the `vendored_static_libraries` feature of `apple_library`.

### Fixtures

This PR adds some new fixtures for testing static libraries

- `tests/ios/libraries/StaticLib`
  - Contains the App target which imports all of the static library
  - Test with `bazel build //tests/ios/libraries/StaticLib:StaticLibApp`
- `tests/ios/libraries/StaticLib/ObjcLib`
  - Contains a fixture for a pure Objective-C static library (for the simulator platforms)
  - Test with `bazel build //tests/ios/libraries/StaticLib/ObjcLib`
 
### Testing

From testing with `bazel build //tests/ios/libraries/StaticLib:StaticLibApp` it looks like support for static libraries in general is broken in `rules_ios`. A fix for this simple case should be sufficient to fix issues in #530 as well.

```sh
bazel build //tests/ios/libraries/StaticLib:StaticLibApp                               

INFO: Build options --copt and --features have changed, discarding analysis cache.
INFO: Analyzed target //tests/ios/libraries/StaticLib:StaticLibApp (0 packages loaded, 1434 targets configured).
INFO: Found 1 target...
ERROR: /Users/lpadron/Development/rules_ios/tests/ios/libraries/StaticLib/BUILD.bazel:3:16: Compiling tests/ios/libraries/StaticLib/App/main.m failed: (Aborted): wrapped_clang failed: error executing command 
  (cd /private/var/tmp/_bazel_lpadron/7c5e30dd733c097d91026d4bca2f7997/sandbox/darwin-sandbox/191/execroot/build_bazel_rules_ios && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=15.5 \
    PATH=/Users/lpadron/Library/Caches/bazelisk/downloads/bazelbuild/bazel-5.0.0-darwin-arm64/bin:/Users/lpadron/.pyenv/shims:/opt/homebrew/opt/ansible@2.9/bin:/Users/lpadron/.rvm/gems/ruby-2.7.4/bin:/Users/lpadron/.rvm/gems/ruby-2.7.4@global/bin:/Users/lpadron/.rvm/rubies/ruby-2.7.4/bin:/Users/lpadron/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/local/bin:/usr/local/munki:/Library/Apple/usr/bin:/Users/lpadron/.cargo/bin:/Users/lpadron/Development/go/bin:/Users/lpadron/.rvm/bin:/Users/lpadron/go/bin:/opt/homebrew/opt/go/libexec/bin \
    XCODE_VERSION_OVERRIDE=13.4.1.13F100 \
    ZERO_AR_DATE=1 \
  external/local_config_cc/wrapped_clang -arch x86_64 '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -g 'DEBUG_PREFIX_MAP_PWD=.' -include rules/library/common.pch -Wshorten-64-to-32 -Wbool-conversion -Wconstant-conversion -Wduplicate-method-match -Wempty-body -Wenum-conversion -Wint-conversion -Wunreachable-code -Wmismatched-return-types -Wundeclared-selector -Wuninitialized -Wunused-function -Wunused-variable -iquote . -iquote bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin -iquote bazel-out/ios-x86_64-min9.0-applebin_ios-ios_x86_64-dbg-ST-ea30d59265dc/bin -MD -MF bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/_objs/StaticLibApp_objc/arc/main.d -DOS_IOS -fno-autolink -isysroot __BAZEL_XCODE_SDKROOT__ -F__BAZEL_XCODE_SDKROOT__/System/Library/Frameworks -F__BAZEL_XCODE_DEVELOPER_DIR__/Platforms/iPhoneSimulator.platform/Developer/Library/Frameworks -fobjc-arc '-mios-simulator-version-min=10.0' -fexceptions -fasm-blocks '-fobjc-abi-version=2' -fobjc-legacy-dispatch -Ibazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/StaticLibApp_private_hmap.hmap -Ibazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/StaticLibApp_public_hmap.hmap -Ibazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/StaticLibApp_private_angled_hmap.hmap -iquotebazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/StaticLibApp_private_hmap.hmap -fmodules '-fmodule-name=StaticLibApp' -I. -index-store-path bazel-out/rules_ios_global_index_store.indexstore '-fmodules-cache-path=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/_objc_module_cache' -O0 '-DDEBUG=1' -fstack-protector -fstack-protector-all -g -c tests/ios/libraries/StaticLib/App/main.m -o bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-0f1b0425081f/bin/tests/ios/libraries/StaticLib/_objs/StaticLibApp_objc/arc/main.o)
# Configuration: 01e20dec9cb3563b1606cbfd9dca336219a362383e7ccebc192785e25f1af2d7

Use --sandbox_debug to see verbose messages from the sandbox
tests/ios/libraries/StaticLib/App/main.m:2:9: fatal error: 'ObjcLib/Foo.h' file not found
#import <ObjcLib/Foo.h>
        ^~~~~~~~~~~~~~~
1 error generated.
Error in child process '/usr/bin/xcrun'. 1
Target //tests/ios/libraries/StaticLib:StaticLibApp failed to build
INFO: Elapsed time: 1.739s, Critical Path: 1.44s
INFO: 2 processes: 2 internal.
```


rules/library.bzl
```diff
@@ -640,11 +641,6 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
 
     for vendored_static_library in kwargs.pop("vendored_static_libraries", []):
         import_name = "%s-%s-library-import" % (name, paths.basename(vendored_static_library))
-        native.objc_import(
-            name = import_name,
-            archives = [vendored_static_library],
-            tags = _MANUAL,
-        )
         vendored_deps.append(import_name)
```
jszumski, line 644: Should you move this line so it stays with `native.objc_import`? It'll work at runtime just fine, but I'm worried if someone adds conditionals around the `objc_import` in the future a bad dep would be added.

luispadron: Totally, good catch

luispadron: Fixed!

jszumski (approved):

tests/ios/libraries/StaticLib/ObjcLib/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+load("@build_bazel_rules_ios//rules:framework.bzl", "apple_framework")
+
+apple_framework(
+    name = "ObjcLib",
+    platforms = {"ios": "9.0"},
+    vendored_static_libraries = [
```
jerrymarino, line 6: Did you consider to provide it a glob of headers instead - mainly, headers aren't tied to the notion of a static library. I think you could write more explict code - and we'd prevent injecting the `*/*.h` 

```
    vendored_static_libraries = [
    ]
    hdrs = ["Header1.h"],
    module_name="Some",
```

jerrymarino (commented):
I'm 50/50 on this one but won't invoke the reject button because we already implicitly `glob *` the headers. I'd suggest to consider moving away from the implict behavior. This is contrary to a framework - which has a distinct directory for headers

rules/library.bzl
```diff
@@ -639,13 +640,6 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
 
     for vendored_static_library in kwargs.pop("vendored_static_libraries", []):
-        import_name = "%s-%s-library-import" % (name, paths.basename(vendored_static_library))
-        native.objc_import(
-            name = import_name,
-            archives = [vendored_static_library],
-            tags = _MANUAL,
-        )
-        vendored_deps.append(import_name)
         import_headers = native.glob(
```
jerrymarino, line 643: Is it possible to remove this and use `hdrs` instead - it might break folks, but I think it is more of a declarative behavior: there might be middleground strategy as well - just ignoring this value if you set `hdrs`

tests/ios/libraries/StaticLib/ObjcLib/BUILD.bazel
luispadron: I wasnt sure what the correct way to declare this type of target was. I saw we gather the headers in the `apple_library` as well, so i didn't add `hdrs` to the target. 

I think providing the headers for this pure static lib makes sense since there aren't multiple slices we need to select over. 

Like you said though if we remove the implicit collection of headers from the library macro then it might break assumptions users are making. Im happy to go that route though, let me know what you think is best.

rules/library.bzl
luispadron: Yep that makes sense, see my comment above. Let me know what you think is best and i can update accordingly. 

IMO id remove the implicit glob we do and let folks update to use `hdrs`. Were still implicitly constructing the `includes` property for `objc_import` but this way we dont make assumptions about where the headers are.

luispadron:
@jerrymarino Just re-read your comment and I agree, removing the implicit behavior of collecting the headers for `vendored_static_libraries` makes sense.

I've updated the PR to remove the implicit glob and instead rely on the `public_headers` property being provided to `apple_framework`

rules/library.bzl
```diff
@@ -639,36 +640,29 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+        import_swiftmodules = native.glob(["**/*.swiftmodule/*.*"], allow_empty = True)
+        import_module_map = import_module_maps[0] if import_module_maps else None
+        vfs_imported_framework = _find_imported_framework_name(objc_hdrs)
         vfs_framework_name = vfs_imported_framework if vfs_imported_framework else namespace
         framework_vfs_overlay(
             name = import_name + "_vfs",
             framework_name = vfs_framework_name,
             modulemap = import_module_map,
             swiftmodules = import_swiftmodules,
-            hdrs = import_headers,
```
jerrymarino, line 671: In order to not break people that relied on the original behavior, we can make this conditional to allow graceful migration to specifying headers:
```
hdrs = objc_hdrs if len(objc_hdrs) > 0 else glob(["**/*.h"])
```

rules/library.bzl
```diff
@@ -639,36 +640,29 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
     for vendored_static_library in kwargs.pop("vendored_static_libraries", []):
+        # Collect the include paths which are the top-level directories for where the headers are.
+        # Example: "Headers/ObjcLib/Foo.h" -> includes = ["Headers"]
+        includes = collections.uniq([hdr.split("/")[0] for hdr in objc_hdrs])
         import_name = "%s-%s-library-import" % (name, paths.basename(vendored_static_library))
         native.objc_import(
             name = import_name,
             archives = [vendored_static_library],
+            hdrs = objc_hdrs,
+            includes = includes,
```
jerrymarino, line 651: We don't support `include` in rules_ios - because of all the issues it brings to the table and we have optimizations banking that it isn't used. Do you have an example where this fixes anything? Adding this line looks like a symptom of some other issue

jerrymarino: Hmm - also in practice this might not be as big of an issue assuming it is correctly stripped off elsewhere

tests/ios/libraries/StaticLib/App/main.m
```diff
@@ -0,0 +1,6 @@
+@import UIKit;
+#import <ObjcLib/Foo.h>
```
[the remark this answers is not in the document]
jerrymarino: @luispadron check my other comments - I think this is an issue with the BUILD file: it's not setting up the module correctly - this is potentially tied to the issue of adding "include" as well

rules/library.bzl
luispadron: I don't believe this was working before if people were relying on the implicit header glob. 

The fixture I added originally didn't pass any `public_headers` and it failed to import without the changes I made in this PR. 

This is because the rest of the code relies on `objc_hdrs` being set, if thats empty (as it would be if `public_headers` was empty) then importing fails.

rules/library.bzl
luispadron: The fixture I added in this repository doesn't build without the `includes` & `hdrs` change 
in `objc_import` and if `public_headers` is empty. Which means that the code as it is today in `master` wasn't working as intended (i.e. collecting headers implicitly). 

The main issue this PR and #530 attempt to fix, comes from `apple_library` doing a bunch of stuff (creating hmap, objc_library, etc) based on the `objc_hdrs`. This property is empty when `public_headers` is empty. Which is the case when the `BUILD` file doesn't define `public_headers`, it will also be empty for `vendored_xcframeworks` since headers are in multiple slices and passing a `select` doesn't work with the code as it is today.

I think the original intent of the code as it is today was to separate the vendored libraries and the top level target that defines them into separate targets. For example, create different `framework_vfs_overlay` for the library and the top-level target. 

So my attempt at fixing this was correcting usage of `objc_import` since in the Bazel docs it says using `includes` allows the `#import <>` statements to work without changing too much of the existing code (implicit collection of headers, etc).

If we want to not provide `includes` to `objc_import` then we need to revert #530 as well and rework `apple_library` to somehow support `select` for `objc_hdrs` (since that will be required for the different slices). This will also change the rest of the code since we implicitly collect headers for every single vendored library (see `import_headers`).

This has all been from me trying to understand the original code, so if I'm missing something then please correct me but it was difficult enough to understand where everything was being created and how these things are intended to work since we never had tests for this in the repo.

Issue #497, opened 2022-06-22
index import blocking build for xcode 14


nanwng:
Trying to build our app from generated xcodeproj in Xcode 14 beta and saw that the index-import script is blocking at the end of build. It seems like the script is not running in background anymore (Could be a new change in Xcode 14)

jerrymarino:
Interesting, is background sub processes broken for you altogether in a post action? It seems like a FB/bug if Xcode waits for background jobs, but maybe intentional!

Also there’s an active issue to consider removing index-import, worst case we can background with another mechanism as a gap or just disable it as a stop gap..

nanwng:
I think this seems to be fixed with later Xcode 14 beta. It works for me with Xcode 14 beta 5

thiagohmcruz:
Cool @nanwng I'll close it for now, feel free to open a new issue if anything similar comes up!

Issue #490, opened 2022-06-07
Update xcconfig for driver: SWIFT_USE_INTEGRATED_DRIVER=NO 


jerrymarino:
Currently there is an issue with Xcode where it tries to use the baked in driver. This causes an issue with the python stub.

We'll remove the need to stub out toolchains by running Bazel directly with XCBuildKit longer term. However, it's necessary to add this to the legacy project  generator and XCHammer until then.

keith:
Nice find

jerrymarino:
You got it 👍

PR #544, opened 2022-08-16
add new flag SWIFT_USE_INTEGRATED_DRIVER and chagne python path

ShawnMa16:
As what was released in the Xcode 14 beta, to make `swift_exec` work, we need to add a new flag `SWFT_USE_INTEGRATED_DRIVER` and set it to `NO`. Further more, we need to update from python to python3 to make `swift-stub.sh` work

[Apple's release note](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes) shows that: 

> Added a new build setting, SWIFT_TOOLS_DIR, which replaces SWIFT_EXEC as the recommended mechanism to use a custom swift-frontend executable. SWIFT_TOOLS_DIR should be set to the path of the directory which contains swift-frontend and related tools. Note that because the Swift driver is now integrated in the Xcode build system, SWIFT_TOOLS_DIR and SWIFT_EXEC cannot be used to specify a custom driver binary unless SWIFT_USE_INTEGRATED_DRIVER is set to NO. (88967344)


chenxiao0228:
@ShawnMa16  can you check why it breaks Xcode 12.5.1 ?

@jerrymarino  what's our strategy on supporting new Xcode versions? Shall we add checks that run on xcode beta runners?

ShawnMa16:
> @ShawnMa16 can you check why it breaks Xcode 12.5.1 ?
> 
> @jerrymarino what's our strategy on supporting new Xcode versions? Shall we add checks that run on xcode beta runners?

I'm looking at that locally

jerrymarino:
First off, thanks so much for the PR and getting this ironed out 🙇 ! Is it safe to say,  we can close out https://github.com/bazel-ios/rules_ios/issues/490 when this lands? Do you have any other issues with new build system or Xcode 14? I was imagining we'd need to face XCBuild related issues 

> @jerrymarino what's our strategy on supporting new Xcode versions? Shall we add checks that run on xcode beta runners?

So far, rules_ios has worked with a number of Xcode versions. We worked to add betas and/or at-least identify issues, then make rules_ios backwards compatible within reason. The CI runs on a version of macOS and Xcode in this repo - which ideally represents the lowest common denominator that a majority of the folks use. 

Unless there is a concern - I consider the betas a "nice" to have - also would gating adding more shards on performance. It seems like we hit capacity issues on the free tier in the past as-is

tools/xcodeproj_shims/swiftc-stub.sh
```diff
@@ -8,7 +8,7 @@ if [[ $# -eq 1 && $1 == "-v" ]]; then
 fi
 
 write_output_files() {
-    cat $1 | python $(dirname "$0")/print_json_leaf_nodes | xargs touch
+    cat $1 | python3 $(dirname "$0")/print_json_leaf_nodes | xargs touch
```
jerrymarino, line 11: This probably isn't a concern for older systems

jerrymarino (approved):

nanwng:
> Is it safe to say, we can close out https://github.com/bazel-ios/rules_ios/issues/490 when this lands?

@jerrymarino I think we can close out #490 once this is merged. Another issue we saw is the background processes broken #497 we are going to fix it in another PR. 
Additionally we are seeing a messed up UI when launching our app from the generated xcodeproj, but it worked properly when launching from command line. We will create an issue for that

luispadron:
In terms of updating Xcode, we've started work on moving minimum to Xcode 13 (might make sense to make that move once Xcode 14 ships). 

#511 gets us mostly there but there is some broken parts that I need some help fixing before we can merge it. It updates the minimum Xcode in CI to `13.4.1` which once Xcode 14 releases seems to match with us supporting Xcode 12 today (i.e. N-1 Xcode release supported)

tools/xcodeproj_shims/swiftc-stub.sh
ShawnMa16: Since the Xcode 14 will only support macOS 12.4 + which doesn't come with python2 anymore, I guess it's OK to do so

PR #545, opened 2022-08-17
Add ability to split a test bundle.

dostrander:
This new macro allows for you to specify a way to "split" an identical bundle into multiple rules with different args for each. To do this you'd specify the different arguments you want to in each split. The suffix of each split will be specified by the key in the dictionary, the different args will be the value. These args will be merged with the ios_test_kwargs specified.

For instance if we have:
`split_ios_test(name = "Foo", split_name_to_kwargs {"MyTestHost1": {"test_host" : //MyTestHost1}, "MyTestHost2": {"test_host": //MyTestHost2}})`

Then target //Foo would have 2 test rules created. 1 named `//Foo_MyTestHost1` and `//Foo_MyTestHost2` and will be packaged together in a test suite of `//Foo`.

**NOTE**
Another possible way to get this similar effect without putting it right in rules_ios is to split out the creation of the test apple library into its own macro which folks could use to create the library and pass it into ios_unit_test

### DIFF
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -62,10 +62,10 @@ def rules_ios_dependencies():
     _maybe(
         github_repo,
         name = "build_bazel_rules_apple",
-        ref = "942c24380a6cb001d3521c510fdb9b61ee164585",
+        ref = "c17af58765d4903a8878d3d6c8004679de4fe8ab",
         project = "bazelbuild",
         repo = "rules_apple",
-        sha256 = "2556021318e6250e00039175e07196d9ae479993712f69047fd239f168f811ee",
+        sha256 = "45846e9da2f055065b99e066fd3e6e438085c81909e6203aedc8c2773819fe90",
     )
 
     _maybe(
--- a/rules/test.bzl
+++ b/rules/test.bzl
@@ -20,9 +20,10 @@ _IOS_TEST_KWARGS = [
     "flaky",
     "frameworks",
     "provisioning_profile",
+    "test_filter",
 ]
 
-def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_build_setting = {}, **kwargs):
+def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_build_setting = {}, split_name_to_kwargs = {}, **kwargs):
     """
     Builds and packages iOS Unit/UI Tests.
 
@@ -38,6 +39,7 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
 
                                      If '//conditions:default' is not set the value in 'infoplists'
                                      is set as default.
+        split_name_to_kwargs: A dictionary of suffixes to kwargs that will be passed into the "split" test bundle. The suffix will be appended to the name of the suite.
         **kwargs: Arguments passed to the apple_library and test_rule rules as appropriate.
     """
 
@@ -50,13 +52,6 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
         fail("cannot specify both runner and runners for %s" % name)
 
     runner = kwargs.pop("runner", kwargs.pop("runners", None))
-    rule = test_rule
-    if runner:
-        if types.is_list(runner):
-            ios_test_kwargs["runners"] = runner
-            rule = test_suite_rule
-        else:
-            ios_test_kwargs["runner"] = runner
 
     # Deduplicate against the test deps
     if ios_test_kwargs.get("test_host", None):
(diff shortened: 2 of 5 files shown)


rules/test.bzl
```diff
@@ -80,6 +153,9 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
         **ios_test_kwargs
     )
 
+def split_ios_unit_test(name, split_name_to_kwargs, apple_library = apple_library, **kwargs):
```
congt, line 156: Could we merge it with ios_unit_test? We can check if split_name_to_kwargs exists to decide if it should call _ios_test or _split_ios_test.

Also could you update the document?

rules/test.bzl
```diff
@@ -19,8 +19,81 @@ _IOS_TEST_KWARGS = [
     "shard_count",
     "flaky",
     "frameworks",
+    "test_filter",
 ]
 
+def _split_ios_test(name, test_rule, test_suite_rule, apple_library, split_name_to_kwargs, infoplists_by_build_setting = {}, **kwargs):
```
congt, line 25: Could you add some tests?

rules/test.bzl
```diff
@@ -19,8 +19,81 @@ _IOS_TEST_KWARGS = [
...
+    dep_name = name + ".dep_middleman"
+    dep_middleman(name = dep_name, deps = kwargs.get("deps", []) + library.lib_names, testonly = True, tags = ["manual"], test_deps = host_args)
+
+    tests = []
+    for suffix, split_kwargs in split_name_to_kwargs.items():
+        all_kwargs = {}
+        all_kwargs.update(ios_test_kwargs)
+        all_kwargs.update(split_kwargs)
+
+        if "runner" in kwargs and "runners" in kwargs:
```
jerrymarino, line 72: Any reason to not pass this via: `split_name_to_kwargs` - then we can simplify / move validation logic out of this rule to the place the `split` kwargs are combined

rules/test.bzl
jerrymarino: Unless we can somehow mitigate this - the downside of changing it from what you have here is now you'll end up recompiling an entirely different `apple_library` per `slice` - where the existing code compiles a test 1 time: e.g. 1 `apple_library` per a set of `srcs`

jerrymarino (commented):
Nice PR - a few inline, can't wait to see this one land 🚀

rules/test.bzl
dostrander: @jerrymarino  i don't think so if we still only compile a single apple_library before "splitting" them i would expect it would still act the same as it does today

jerrymarino (approved):
@dostrander this LGTM overall.

It's up to you but I'd noodle on the name a bit `split_name_to_kwargs` -> `shards`. From an API perspective - you could potentially simplify this down something like this:

```py
ios_unit_test(name="x", shards=["A", "B"])
```
You'd have to apply the same `kwargs` to all shards. The downside is you'd have to make a different bundle for each `runner`. I'll defer to you on that 🚢

jerrymarino:
Also thanks again for the PR - this is going to make remotely executing these tests awesome and you've done it in a super general way ❤️

rules/repositories.bzl
```diff
@@ -62,10 +62,10 @@ def rules_ios_dependencies():
     _maybe(
         github_repo,
         name = "build_bazel_rules_apple",
-        ref = "942c24380a6cb001d3521c510fdb9b61ee164585",
-        project = "bazelbuild",
+        ref = "f628e6d33e95dd395b07330d7e191f31d6bd525c",
+        project = "dostrander",
```
jszumski, line 66: Do we need this fork if https://github.com/bazelbuild/rules_apple/pull/1616 is merged?

dostrander: That is merged but it isn't in the release of 5.x which causes an error of:

```
File "/private/var/tmp/_bazel_derko/45d370c48a980ff7ee8dfaac4184ab30/external/build_bazel_rules_apple/apple/internal/transition_support.bzl", line 528, column 46, in <toplevel>
                _apple_platform_split_transition = transition(
Error in transition: Invalid transition input '//command_line_option:incompatible_enable_apple_toolchain_resolution'. Cannot transition on --experimental_* or --incompatible_* options
ERROR: error loading package 'tests/ios/frameworks/sources-with-prebuilt-binaries': at /Users/derko/Development/rules_ios/rules/apple_patched.bzl:4:5: at /private/var/tmp/_bazel_derko/45d370c48a980ff7ee8dfaac4184ab30/external/build_bazel_rules_apple/apple/apple.bzl:18:5: at /private/var/tmp/_bazel_derko/45d370c48a980ff7ee8dfaac4184ab30/external/build_bazel_rules_apple/apple/apple_static_library.bzl:18:5: initialization of module 'apple/internal/transition_support.bzl' failed
```

When using master. The 5.x must have this fix but not master :/

dostrander:
> @dostrander this LGTM overall.
> 
> It's up to you but I'd noodle on the name a bit `split_name_to_kwargs` -> `shards`. From an API perspective - you could potentially simplify this down something like this:
> 
> ```python
> ios_unit_test(name="x", shards=["A", "B"])
> ```
> 
> You'd have to apply the same `kwargs` to all shards. The downside is you'd have to make a different bundle for each `runner`. I'll defer to you on that 🚢

Yeah I'm planning on doing that before merge. This is somewhat of a PoC based on my first go at it before the feedback

dostrander:
> > @dostrander this LGTM overall.
> > It's up to you but I'd noodle on the name a bit `split_name_to_kwargs` -> `shards`. From an API perspective - you could potentially simplify this down something like this:
> > ```python
> > ios_unit_test(name="x", shards=["A", "B"])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You'd have to apply the same `kwargs` to all shards. The downside is you'd have to make a different bundle for each `runner`. I'll defer to you on that 🚢
> 
> Yeah I'm planning on doing that before merge. This is somewhat of a PoC based on my first go at it before the feedback

That would make it a lot harder to do sharding on this as we will be applying the same `test_filter` kwarg to all of them. The dictionary allows us to specify different kwargs

rules/test.bzl
dostrander: Added

rules/library.bzl
```diff
@@ -97,7 +97,7 @@ extend_modulemap = rule(
     doc = "Extends a modulemap with a Swift submodule",
 )
 
-def _write_modulemap(name, umbrella_header = None, module_name = None, framework = False):
+def _write_modulemap(name, umbrella_header = None, public_headers = [], private_headers = [], module_name = None, framework = False):
```
dostrander, line 100: @dierksen FYI added these back in for use of custom module maps that need to know the public/private header

jerrymarino: Why did this PR end up having to change the module maps? Overall I think this should probably be removed - it would just set the `modulemap` attribute of `apple_framework` better

jerrymarino: Agh I see what happened here - sorry for the breakage
https://github.com/bazel-ios/rules_ios/commit/0f024b4bea57bbfa6c13e5c2c53c07a1320358ff#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478L811

TBH I wonder how this got through rules_ios CI - it's something we can follow up with

rules/test.bzl
```diff
@@ -73,13 +68,57 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
...
+        tests = []
+        for suffix, split_kwargs in split_name_to_kwargs.items():
+            test_name = "{}_{}".format(name, suffix)
+
+            all_kwargs = {}
+            all_kwargs.update(ios_test_kwargs)
+            all_kwargs.update(split_kwargs)
+
+            if runner and ("runner" in all_kwargs or "runners" in all_kwargs):
+                fail("cannot runner/s in both split and top level kwargs for %s" % test_name)
```
luispadron, line 81: nit:
```suggestion
                fail("cannot use runner/s attribute in both split and top level kwargs for %s" % test_name)
```

luispadron (approved):
I'd love to not have to fork if we can mirror the change to the `bazel/5.x` branch. Change looks good though!

tests/test-tests.sh
```diff
@@ -0,0 +1,16 @@
...
+set -euo pipefail
+
+## split tests
+test_suite_count=$(bazel query 'kind("test_suite", //tests/ios/unit-test:SplitTests)' | wc -l | xargs)
+if [[ "$test_suite_count" != "1" ]]; then
+  echo "Expected 1 test suite for split tests got $test_suite_count"
+  exit 1
+fi
+
+tests_count=$(bazel query 'tests(kind("test_suite", //tests/ios/unit-test:SplitTests))' | wc -l | xargs)
```
thiagohmcruz, line 12: `tests/xcodeproj-tests.sh` invokes `bazelisk`. Should we do the same here just for consistency?

thiagohmcruz (approved):

tests/test-tests.sh
dostrander: Updated

PR #547, opened 2022-08-18
set index-import priority to lowest

larrytaowang:
Sometimes when indexing, `index-import` eats all CPU and network resources, then the internet is cut out and the laptop is frozen. 

Setting `index-import` to the lowest priority can help with this problem. The CPU usage is still high but the laptop is responsive.

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_indexstore.sh
@@ -57,7 +57,7 @@ fi
 
 echo "Running index-import for arch $ARCH and parallel-stride set to $PARALLEL_STRIDE"
 
-$BAZEL_INSTALLERS_DIR/index-import \
+nice -n 20 $BAZEL_INSTALLERS_DIR/index-import \
     -parallel-stride $PARALLEL_STRIDE \
     -incremental \
     -remap "$remote_developer_dir=$local_developer_dir" \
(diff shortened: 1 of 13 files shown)


jerrymarino:
Minor suggestion inline - I think suspect there are a number of other performance improvements ranging from bumping index-import to the arm64 cut. @thiagohmcruz might have some other thoughts or opinions on some of this too

thiagohmcruz:
@larrytaowang just wanted to confirm, are you running a version of rules_ios after this change https://github.com/bazel-ios/rules_ios/pull/489?

It's something that improved things a bit, we were passing bad input to index-import because of that, that plus the arm64 bump has improved things a bit but there's def more testing/optimizations to do here.

I'm fine with landing this, just wanted to know if the fix above has helped others at all

thiagohmcruz (approved):

jerrymarino (approved):
On a busy system running threads like `mds`, it software, iOS apps, are you concerned about total priority inversion or it just never completing / regressing the code editing experience? I haven't tried to pull this PR but I assumed you did 

Yep - I hope we arrive at a solution soon [remove much unnecessary data processing and make it "real time"](https://github.com/bazel-ios/rules_ios/issues/227) - that is much a simpler solution from than what we do right now from a schduler perspective

larrytaowang:
> On a busy system running threads like `mds`, it software, iOS apps, are you concerned about total priority inversion or it just never completing / regressing the code editing experience?

sorry I totally lost track of this PR. yes, I think we already have this commit, but sometimes the indexing process still eats all the CPU resouces.

larrytaowang:
> On a busy system running threads like `mds`, it software, iOS apps, are you concerned about total priority inversion or it just never completing / regressing the code editing experience?

Usually it is fine, but sometimes it can be really bad. The indexing process eats all the CPU and network resources, and devs cannot even use chrome.

jerrymarino:
Ah makes sense, yeah we've still got a few changes to make to get rid of `index-import` how it runs here - but it's nearly done AFAIK!

PR #548, opened 2022-08-22
Remove extra filegroups

jerrymarino:
This creates a number of extra rules that can go away - it's effectivly
unused

This change is done for simplify - not heap size. This said, it cuts an
entire category from the target graph: Because this removes 3 for each
rule,  we'll eliminate this many pointers from memory:
```
O((Napple_framework + Kheaders) * 3)
```

After this change - you'll end up with O(Nresource groups) filegroups.
For me, in larger build - the measured impact is a reduction of ~10k
filegroups -> 89


dierksen (approved):

luispadron (approved):

thiagohmcruz (approved):

ob (approved):
lol, so inefficient before :-) 

Thanks for cleaning this up

mattrobmattrob (approved):

Issue #537, opened 2022-08-12
framework macro does not seem to include cpp source files


gyfelton:
How to repro:
1. modify tests/ios/app/Cpp/hello.cpp by adding some bad text which should fail the compilation
2. run `bazel build //tests/ios/app:CppLib`, expect it to fail but it is green
3. run `bazel build //tests/ios/AppWithSelectableCopts`, expect it to fail since it depends on CppLib but it is green still
4. add `hello.h` declaring the hello method, run above steps, expect it to fail but it is still successful

jszumski:
C++ sources are intentionally ignored: https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl#L955

I don't believe there is a technical reason however, it's just waiting for a contribution.

jerrymarino:
Yep it looks that test case is broken - and doesn't even compile `tests/ios/app/Cpp/hello.cpp` ATM. There's an open a PR about this: https://github.com/bazel-ios/rules_ios/pull/304 but a few pending questions on it I last checked.

I think you'd need to ensure it dedupes. PodToBUILD splits out different `objc_library` in order to give it different copts for c++ and objc - we could probably do a similar thing here as well

gyfelton:
Gotcha! Thx for quick reply! I think we disable it for VFS, but yeah a distinct objc_library is def. the next step to try

PR #550, opened 2022-08-24
Enables C++ support in srcs

jszumski:
An alternative to https://github.com/bazel-ios/rules_ios/pull/304 that enables the existing C++ target.  Resolves https://github.com/bazel-ios/rules_ios/issues/537.

- Sends C++ sources into a separate target instead of the existing Obj-C target to enable different `copt` values for each.
- Re-uses the same mixed Obj-C and C++ test fixture, thanks @ivan-golub!

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -1022,14 +1022,15 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         _append_headermap_copts(swift_angle_bracket_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
 
     # Note: this line is intentionally disabled
-    if cpp_sources and False:
+    if cpp_sources:
         additional_cc_copts.append("-I.")
-        native.cc_library(
+        native.objc_library(
             name = cpp_libname,
-            srcs = cpp_sources + objc_private_hdrs,
+            srcs = cpp_sources + objc_private_hdrs + objc_non_exported_hdrs,
             hdrs = objc_hdrs,
             copts = copts_by_build_setting.cc_copts + cc_copts + additional_cc_copts,
-            deps = deps,
+            deps = deps + private_deps,
+            defines = defines,
             tags = tags_manual,
             testonly = kwargs.get("testonly", False),
         )
--- a/tests/ios/frameworks/objcpp/BUILD.bazel
+++ b/tests/ios/frameworks/objcpp/BUILD.bazel
@@ -1,0 +1,56 @@
+load("//rules:framework.bzl", "apple_framework")
+load("//rules:test.bzl", "ios_unit_test")
+
+apple_framework(
+    name = "ObjCppMixedLib",
+    srcs = glob(
+        [
+            "ObjCppMixedLib/*.mm",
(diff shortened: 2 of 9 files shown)


rules/library.bzl
```diff
@@ -1022,14 +1022,15 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         _append_headermap_copts(swift_angle_bracket_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
 
     # Note: this line is intentionally disabled
-    if cpp_sources and False:
+    if cpp_sources:
         additional_cc_copts.append("-I.")
-        native.cc_library(
+        native.objc_library(
```
jszumski, line 1027: This had to be an `objc_library` to avoid an error in `framework_packaging_impl` where it expected all deps would have an `Objc` provider

luispadron (approved):
Believe this closes #537

rules/library.bzl
```diff
@@ -1022,14 +1022,15 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
             name = cpp_libname,
-            srcs = cpp_sources + objc_private_hdrs,
+            srcs = cpp_sources + objc_private_hdrs + objc_non_exported_hdrs,
             hdrs = objc_hdrs,
             copts = copts_by_build_setting.cc_copts + cc_copts + additional_cc_copts,
```
jerrymarino, line 1031: This is missing a number of flags: perhaps you can add as a followup.
```

    additional_objc_vfs_deps = select({
        "@build_bazel_rules_ios//:virtualize_frameworks": [framework_vfs_overlay_name_swift] + [framework_vfs_overlay_name],
        "//conditions:default": [framework_vfs_overlay_name_swift] + [framework_vfs_overlay_name] if enable_framework_vfs else [],
    })
    additional_objc_vfs_copts = select({
        "@build_bazel_rules_ios//:virtualize_frameworks": framework_vfs_objc_copts,
        "//conditions:default": framework_vfs_objc_copts if enable_framework_vfs else [],
    })

```

jerrymarino (approved):
@jszumski - thanks for the PR and adding the test case  overall, this LGTM! The one question I had for you is around the `.mm` files. Addressing that may be better as a followup if it's going to change behavior for people. Historically, these needed to be compiled adjacent to the C++ in Bazel. In PodToBUILD `ObjcLibrary.swift` I believe you'll find it this way or similar:  to allow mirror cocoapods and apply C++ specific flags ( e.g. from CXX cconfig group ) to the files

mattrobmattrob (approved):
FWIW, this works in our build which has C++ sources for a couple CocoaPods. We were previously rewriting these into `.mm` files and now we don't have to! 🎉 

```Starlark
    if not name.endswith(".cpp"):
        return name

    return _copy_genrule(
        name.replace(".cpp", "_transmmogrified.mm"),
        name,
        name.replace(".cpp", ".mm"),
    )
```

rules/library.bzl
jszumski: A follow up is a good idea 👍

luispadron (approved):

PR #553, opened 2022-08-30
Add support for project options in `legacy_xcodeproj`

mattrobmattrob:
# Summary
We override the `Text Settings` in our non-`legacy_xcodeproj` XcodeGen config:
```YML
options:
  # Text indent
  usesTabs: false
  indentWidth: 2
  tabWidth: 2
```
Add support for `options` to be overridden via the `project_options_overrides` in `legacy_xcodeproj`. It might seem weird that this requires an entire new `rule` impl but this is done since XcodeGen won’t process `”2”` correctly from a `string_dict` for arguments like `indentWidth` or `tabWidth`. `_ProjectOptionsInfo `, a new provider, allows the values to be forwarded to the JSON payload as the properly encoded `int` and `bool` values as expected by the underlying project gen tool:
```JSON
{
    "options": {
        "createIntermediateGroups": true,
        "defaultConfig": "Debug",
        "groupSortPosition": "none",
        "settingPresets": "none",
        "usesTabs": false,
        "indentWidth": 2,
        "tabWidth": 2
    },
}
```

# Testing
`BUILD.bazel`:
```Starlark
load("@build_bazel_rules_ios//rules:xcodeproj.bzl", "xcodeproj")
load("@build_bazel_rules_ios//rules:legacy_xcodeproj.bzl", "xcodeproj_project_options")

xcodeproj_project_options(
    name = "ProjectOptions",
    uses_tabs = False,
    indent_width = 2,
    tab_width = 2,
)

xcodeproj(
    name = "Project",
    testonly = True,
    generate_schemes_for_product_types = [
        "application",
        "bundle.unit-test",
        "test_suite",
    ],
    project_options_overrides = ":ProjectOptions",
    include_transitive_targets = True,
    deps = [":Foo"],
)
```

|Xcode 13.4.1|
|---|
|Screen Shot 2022-08-30 at 2 06 48 PM|

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -475,7 +475,8 @@ def _xcodeproj_project_options_impl(ctx):
     # Accumulate all provided options into a Starlark dict.
     options = {}
     for attr, option in options_map.items():
-        if hasattr(ctx.attr, attr):
+        attr_value = getattr(ctx.attr, attr, None)
+        if attr_value:
             options[option] = getattr(ctx.attr, attr)
     return [_ProjectOptionsInfo(options = options)]
 
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -486,7 +486,7 @@ xcodeproj_project_options = rule(
         "indent_width": attr.int(mandatory = False),
         "tab_width": attr.int(mandatory = False),
     },
-    doc = "Rule to change project options",
+    doc = "Rule to change project options, values will fallback to Xcode default if not provided",
 )
 
 def _collect_swift_defines(modules):
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -16,6 +16,7 @@ def _get_attr_values_for_name(deps, provider, field):
         if dep and provider in dep
     ]
 
+_ProjectOptionsInfo = provider()
 _TargetInfo = provider()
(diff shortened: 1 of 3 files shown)


mattrobmattrob:
I'm not very familiar with the organization of the project gen code since this is my first time looking at it so let me know if a different location for the `rule` or provider or anything makes more sense.

rules/legacy_xcodeproj.bzl
```diff
@@ -463,6 +464,31 @@ xcodeproj_lldbinit = rule(
...
+    return [_ProjectOptionsInfo(options = options)]
+
+xcodeproj_project_options = rule(
+    implementation = _xcodeproj_project_options_impl,
+    attrs = {
+        "uses_tabs": attr.bool(mandatory = False),
+        "indent_width": attr.int(mandatory = False),
+        "tab_width": attr.int(mandatory = False),
+    },
+    doc = "Rule to change project options",
```
jerrymarino, line 489: Minor: maybe add a note to cover what happens if you _don't_ provide this option: e.g. if you don't set it then it will fall back to xcode's defaults ( assuming xcodegen doesn't have their own flavor of default value )

jerrymarino (approved):
Nice one @mattrobmattrob 👍

rules/legacy_xcodeproj.bzl
mattrobmattrob:
```suggestion
    doc = "Rule to change project options, values will fallback to Xcode default if not provided",
```

Issue #484, opened 2022-05-19
Prototype performance of remote compiled clang framework modules 


jerrymarino:
We'd like a prototype to demonstrate the performance implications and identify unknowns of IDE integration issues when compiling framework modules with Bazel. 

Consider how these are loaded into clang and swift, how to cache them, and how it works with the IDE ( lldb, etc ). Currently a remote compilation action pays the worst case cost `O(NDeps)` implicit module compilation and is riddled with other cache invalidation issues. If compiling them with Bazel, we'll be able to reliably cache / load state onto workers.

jerrymarino:
Well it looks like explicit module compilation support has recently been accelerating in [`swift-driver`](https://github.com/apple/swift-driver/blob/e1a60b010a67b1c4705d65cbdc7bcff283fc914c/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift) - the rewrite of the original C++ driver baked into the swift compiler - it seems like that might help validate some of this end to end - or even be a nod of how to do it in a supported way.

The other pontentially helpful bit is the [C++ module compilation action inside of Bazel](https://github.com/bazelbuild/bazel/blob/fb3110aa168d9f04e9195d4964d53d3fde2f4328/src/main/java/com/google/devtools/build/lib/rules/cpp/CppRuleClasses.java#L193). What has been  unclear how to pull the actual compilation  off in starlark: e.g. if we'll need header pruning for this feature, or how we'd invoke compilation with `cc_common` 🤔

jerrymarino:
For system modules - it looks like the bit you'll need to do is now open sourced - and easy to digest in [part of buck](https://github.com/facebook/buck/blob/dev/src/com/facebook/buck/swift/SwiftModuleMapCompile.java#L195) ❤️

jerrymarino:
Yesterday I looked into this a bit more and trying to reason about how to fit this into Bazel's paradigms of loading, analysis, and execution phases: both from an Xcode side and intra-build side. It looks like rules_swift has the "precompile module" clang invocation mostly there - so my theoretical pseudo was focused on the Xcode side

[...]

PR #554, opened 2022-08-31
[WIP] First pass at rules_ios explicit PCM test case 

jerrymarino:
The first pass compiles them by calling `cc_common` via a helper they added into `rules_swift` that creates a provider with `cc_common`. Longer term, we may remove the rules_swift dependency here: calling `cc_common` or `SpawnAction` directly 

Putting this into the bazel-ios org so we can cleanly integrate it with our providers, own the code, and test it end to end.

cc @dierksen I've taken a stab at PRing some of the stuff you've added into swift_scratch so we can iterate on it as a team.

### DIFF
--- a/rules/pcm/apple_framework_pcm.bzl
+++ b/rules/pcm/apple_framework_pcm.bzl
@@ -1,0 +1,64 @@
+load("@build_bazel_rules_swift//swift/internal:swift_common.bzl", "swift_common")
+load("@build_bazel_rules_swift//swift/internal:attrs.bzl", "swift_library_rule_attrs")
+load("@build_bazel_rules_swift//swift/internal:providers.bzl", "SwiftInfo", "SwiftToolchainInfo")
+load("@bazel_skylib//lib:paths.bzl", "paths")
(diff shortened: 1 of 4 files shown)


rules/pcm/apple_framework_pcm.bzl
```diff
@@ -0,0 +1,64 @@
+load("@build_bazel_rules_swift//swift/internal:swift_common.bzl", "swift_common")
```
jerrymarino, line 1: @dierksen I've taken a look through the code last night and this morning to get a better grip on it - I've got some good news!

I found all of these APIs to be [officially supported public APIs](https://github.com/bazelbuild/rules_swift/issues/891#issue-1357849822) because of this - we can leverage the and don't need to tie our development of PCM compilation to rules_swift in any way

jerrymarino:
@dierksen - I'm going to close this out and defer to you on what to do with this, but atleast it exemplifies we potentially pull this off without calling some private APIs

jerrymarino:
This is in reference to the rules_ios issue and context here BTW: https://github.com/bazel-ios/rules_ios/issues/484

PR #555, opened 2022-08-31
[WIP] Cut target graph/memory usage down / simplify compiling

jerrymarino:
This PR cuts down the target graph and rule instance usage signficantly.
Today we create sevearl rule instances in memory and add complexity by
calling `swift_library` and `objc_library`. This is a paradigm shift
that allows us to more easily iterate on `rules_ios` and dramatically
simplify PCM compilation and Xcode integration. e.g.
https://github.com/bazel-ios/rules_ios/pull/554

For instance, we can create explict PCMs, headermaps, vfsoverlays
without ballooning the target graph - and more deliberately integrate
these with our own providers

This PR is mainly a WIP PR - there are a couple small things missing.
TODO:
1. fix non_arc_srcs integration
2. xcode integration

### DIFF
--- a/rules/apple_library.bzl
+++ b/rules/apple_library.bzl
@@ -1,0 +1,673 @@
+"""Implementation of the `apple_library` rule."""
+
+load("//rules/framework:vfs_overlay.bzl", "framework_vfs_overlay", VFS_OVERLAY_FRAMEWORK_SEARCH_PATH = "FRAMEWORK_SEARCH_PATH")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("//rules/internal:objc_provider_utils.bzl", "objc_provider_utils")
+load("//rules:providers.bzl", "AvoidDepsInfo", "FrameworkInfo")
+load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", "make_vfsoverlay")
+load("@build_bazel_rules_swift//swift/internal:attrs.bzl", "swift_common_rule_attrs", "swift_deps_attr", "swift_toolchain_attrs")
+load(
+    "@build_bazel_rules_swift//swift/internal:build_settings.bzl",
+    "PerModuleSwiftCoptSettingInfo",
+    "additional_per_module_swiftcopts",
+)
+load(
+    "@build_bazel_rules_swift//swift/internal:compiling.bzl",
+    "apple_library_output_map",
+    "new_objc_provider",
+    "output_groups_from_other_compilation_outputs",
+)
+load(
+    "@build_bazel_rules_swift//swift/internal:feature_names.bzl",
+    "SWIFT_FEATURE_EMIT_SWIFTINTERFACE",
+    "SWIFT_FEATURE_ENABLE_LIBRARY_EVOLUTION",
+)
+
+#load("@build_bazel_rules_swift//swift/internal:linking.bzl", "create_linker_input")
+load("@build_bazel_rules_swift//swift/internal:providers.bzl", "SwiftInfo", "SwiftToolchainInfo")
+load("@build_bazel_rules_swift//swift/internal:swift_clang_module_aspect.bzl", "swift_clang_module_aspect")
+load("@build_bazel_rules_swift//swift/internal:swift_common.bzl", "swift_common")
+load(
+    "@build_bazel_rules_swift//swift/internal:utils.bzl",
+    "compact",
+    #"create_cc_info",
+    "expand_locations",
+    "expand_make_variables",
+    "get_providers",
+)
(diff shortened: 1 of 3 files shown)


rules/apple_library.bzl
```diff
@@ -0,0 +1,673 @@
...
+)
+load(
+    "@build_bazel_rules_swift//swift/internal:feature_names.bzl",
+    "SWIFT_FEATURE_EMIT_SWIFTINTERFACE",
+    "SWIFT_FEATURE_ENABLE_LIBRARY_EVOLUTION",
+)
+
+#load("@build_bazel_rules_swift//swift/internal:linking.bzl", "create_linker_input")
+load("@build_bazel_rules_swift//swift/internal:providers.bzl", "SwiftInfo", "SwiftToolchainInfo")
+load("@build_bazel_rules_swift//swift/internal:swift_clang_module_aspect.bzl", "swift_clang_module_aspect")
```
jerrymarino, line 28: I think we can remove this once we get have a better story and approach to propagate PCMs. By nixing these aspects, we'll further reduce memory and speed up the analysis phase.

rules/apple_library.bzl
```diff
@@ -0,0 +1,673 @@
...
+    objc_non_arc_srcs = []
+
+    # TODO: these are smashed as 1 target
+    for h in ctx.attr.non_arc_srcs:
+        if DefaultInfo in h:
+            objc_non_arc_srcs.extend(h[DefaultInfo].files.to_list())
+        else:
+            objc_non_arc_srcs.append(h)
+
+    objc_cc_info = None
```
jerrymarino, line 422: The ideas are here - but is should be split into a number of helper functions instead of 1 massive one 🔨

jerrymarino:
This was part of my hackweek project to further optimize rules_ios for a big app. For a t-shirt sized estimate - I hit 1.5gb of RAM usage and ~45secs for a clean analysis. In early testing, this got it down to 700mb and ~30secs. There is still a lot of improvements left on the table by simplifying our providers

This is highly WIP: but want to put this on the table as the next direction for rules_ios: moving the loading phase actions of `library.bzl` into discrete analysis level logic and completing the transition to our own simplified providers: e.g. `FrameworkInfo`

rules/apple_library.bzl
```diff
@@ -0,0 +1,673 @@
+"""Implementation of the `apple_library` rule."""
+
+load("//rules/framework:vfs_overlay.bzl", "framework_vfs_overlay", VFS_OVERLAY_FRAMEWORK_SEARCH_PATH = "FRAMEWORK_SEARCH_PATH")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("//rules/internal:objc_provider_utils.bzl", "objc_provider_utils")
+load("//rules:providers.bzl", "AvoidDepsInfo", "FrameworkInfo")
+load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", "make_vfsoverlay")
+load("@build_bazel_rules_swift//swift/internal:attrs.bzl", "swift_common_rule_attrs", "swift_deps_attr", "swift_toolchain_attrs")
```
jerrymarino, line 8: Most of this has migrated to a public API now - move from `internal`

ob:
Nice, so moving `apple_library()` to a macro and calling the swift/apple rules functions directly saves a bunch of memory eh? I wonder if we can clean up the header propagation story with a custom provider that is cleaner than passing around partial providers for objc/swift... I guess that would be the purpose of `FrameworkInfo` provider... 🤔

jerrymarino:
> Nice, so moving apple_library() to a macro and calling the swift/apple rules functions directly saves a bunch of memory eh? 

Yep, the memory savings were pretty decent but I'll have to complete it to get the full picture. The other one is speeding up the analysis phase - I want to get that to a very low number 📈 

> I wonder if we can clean up the header propagation story with a custom provider that is cleaner than passing around partial providers for objc/swift... I guess that would be the purpose of FrameworkInfo provider... 🤔

My other goal is to reduce looping by exclusively using our own data structures: e.g. `FrameworkInfo` to represent the _interface_ amongst compilation dependencies - and have another performant one for the linking deps / concerns. We can then cleanly interface with primitives from external rules (  e.g.`bazelbuild` / `PodToBUILD` / `rules_lime` ) by ingesting said dependencies at a single point. Then, cleanup `framework.bzl` to reduce rules_ios thrash on Bazel updates

jerrymarino:
We can sync more about the direction on this soon, going to close it up now - and wanted to exemplify how we can process and inject PCMs at this level - and possibly use this to bootstreap a custom `swift_library` in the mid-term with public apis: e.g. `swift_common.compile` - if it comes to this! cc @dierksen

PR #557, opened 2022-09-08
Supports exported_symbols_lists in apple_framework

congt:
It's already supported in the apple_framework_packaging rule. We need to pass it from apple_framework to apple_framework_packaging.


jerrymarino (approved):
Thanks for the PR @congt 🚀

PR #561, opened 2022-09-14
Add provisioning_profile to test rules

thiagohmcruz:
Required for folks trying to run tests on device

### DIFF
--- a/tests/ios/app/BUILD.bazel
+++ b/tests/ios/app/BUILD.bazel
@@ -112,6 +112,14 @@ ios_application(
     ],
 )
 
+ios_unit_test(
+    name = "TestAppWithProvisioningProfile",
+    srcs = ["empty_tests.m"],
+    minimum_os_version = "10.0",
+    provisioning_profile = "@build_bazel_rules_ios//tests/ios:integration-test.mobileprovision",
+    test_host = ":AppWithProvisioningProfile",
+)
+
 ios_application(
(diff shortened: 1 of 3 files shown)


tests/ios/app/BUILD.bazel
```diff
@@ -112,6 +112,14 @@ ios_application(
     ],
 )
 
+ios_unit_test(
+    name = "TestAppWithProvisioningProfile",
+    srcs = ["empty_tests.m"],
+    minimum_os_version = "10.0",
+    provisioning_profile = "@build_bazel_rules_ios//tests/ios:integration-test.mobileprovision",
```
thiagohmcruz, line 119: We can think of more elaborate test for the "on device" use case, just adding a way to catch some possible failure modes when the attr is set

I believe this use case as whole is missing more test cases and could receive some love but not planning to tackle it in this PR

luispadron (approved):

tests/ios/app/BUILD.bazel
jerrymarino: What'd be cool is to add some signing tests with an ad-hoc identity - we might end up needing to bake that into the repo. Longer term we can probably have some end to end integration test with the Xcode GUI stuff as well 👍

jerrymarino (approved):
Thanks for the PR @thiagohmcruz!

PR #562, opened 2022-09-15
Add rules for building SDK frameworks for explicit module compilation

dierksen:


tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
```diff
@@ -0,0 +1,141 @@
...
+def _rule_for_pkg(developer_dir, sdk_path, pkg):
+    module_path = pkg.get("modulePath", "")
+    if not module_path or IMPORTS_FILE in module_path:
+        return ""
+    module_name = module_path.replace(".pcm", "").replace(".swiftmodule", "")
+
+    pkg_deps = pkg.get("directDependencies", [])
+    deps = ["{}_c".format(d["clang"]) for d in pkg_deps if "clang" in d]
+    deps += ["{}_swift".format(d["swift"]) for d in pkg_deps if "swift" in d]
+    deps_string = "\n".join(sorted(["        \":{}\",".format(d) for d in deps]))
```
thiagohmcruz, line 95: Sorry if this is a dumb Q but is it expected for all deps be propagated to both `swift_c_module` and `swift_module_alias`? Or should the `_swift` ones be passed only to `swift_module_alias`?

thiagohmcruz (commented):
Just a few questions/nits for now! Can take a second pass once it's ready for review

tools/toolchains/xcode_configure/xcode_configure.bzl
```diff
@@ -275,6 +286,7 @@ xcode_autoconf = repository_rule(
     attrs = {
         "xcode_locator": attr.string(),
         "remote_xcode": attr.string(),
+        "explicit_modules": attr.bool(),
```
jerrymarino, line 289: How long does it take to run this one? This will run many times per day having the requirement of being a `local` rule.  Longer term, maybe we'd breaking the `xcode_locator` into it's [own repo rule](https://github.com/bazel-ios/rules_ios/issues/484#issuecomment-1221169347) which will allow us to add non local cache semantics on the pcm gem::
- `xcode_locator(local=True)`
- `pcm_gem(sdk=X, xcode_locator, local=False)`

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
```diff
@@ -0,0 +1,141 @@
...
+    platforms_dir = developer_dir_path.get_child("Platforms")
+    for platform_name, target_name in PLATFORM_TUPLES:
+        platform_dir = platforms_dir.get_child(platform_name + ".platform")
+        platform_version = versions[target_name]
+
+        sdk_dir = platform_dir.get_child("Developer").get_child("SDKs").get_child("{}.sdk".format(platform_name))
+        target_triple = _platform_target_triple(platform_name, target_name, platform_version)
+
+        output_folder = repository_ctx.path(xcode_version_name).get_child(platform_name)
+        _create_build_file_for_sdk(repository_ctx, developer_dir_path, sdk_dir, output_folder, target_triple)
```
jerrymarino, line 141: This structure is nice and makes for an easy partitioning to Xcode 14 decomposed `watchos` `tvos` SDKs or we have to load SDKs from elsewhere, or different rules 👍

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
```diff
@@ -0,0 +1,141 @@
...
+"""
+
+def _sub_bazel_path_vars(path, sdk_path, developer_dir):
+    path = path.replace(str(sdk_path), "__BAZEL_XCODE_SDKROOT__")
+    path = path.replace(str(developer_dir), "__BAZEL_XCODE_DEVELOPER_DIR__")
+    return path
+
+def _rule_for_pkg(developer_dir, sdk_path, pkg):
+    module_path = pkg.get("modulePath", "")
+    if not module_path or IMPORTS_FILE in module_path:
```
jerrymarino, line 88: Should this be a `fail` ? It looks like it detects a module we didn't load in?

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
```diff
@@ -0,0 +1,141 @@
...
+        suffix = "-simulator" if platform.endswith("Simulator") else "",
+    )
+
+def create_xcode_framework_targets(repository_ctx, xcode_version_name, developer_dir, versions):
+    """Creates a BUILD file for all the SDK frameworks contained in a given Xcode version."""
+    developer_dir_path = repository_ctx.path(developer_dir)
+    platforms_dir = developer_dir_path.get_child("Platforms")
+    for platform_name, target_name in PLATFORM_TUPLES:
+        platform_dir = platforms_dir.get_child(platform_name + ".platform")
+        platform_version = versions[target_name]
```
jerrymarino, line 135: For `Xcode 14` - it looks like if a `PLATFORM` on disk e.g. `tvos` - this will throw an error - potentially as a stopgap to make it work for iOS to buy us more time to fix it:
```
if not platform_dir.exists():
   print("Warning: skipping SDK for Xcode 14.0")
   continue
```

jerrymarino (approved):
@dierksen Nice - the comments I've got are mostly minor - if we want to iterate on this incrementally in smaller bits that seems fine 👍

.vscode/settings.json
```diff
@@ -1,6 +1,3 @@
 {
-    "files.exclude": {
```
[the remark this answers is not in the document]
dierksen: Not sure what you mean here, but I was constantly checking stuff in `bazel-out` for sanity-checking, but I'm also not ready to land this yet.

In general, both of these seem like settings that would make more sense in a personal setting and not on a repo level.

tools/toolchains/xcode_configure/xcode_configure.bzl
```diff
@@ -73,6 +75,12 @@ def _xcode_version_output(repository_ctx, name, version, aliases, developer_dir)
     tvos_sdk_version = _search_sdk_output(xcodebuild_result.stdout, "appletvos")
     macos_sdk_version = _search_sdk_output(xcodebuild_result.stdout, "macosx")
     watchos_sdk_version = _search_sdk_output(xcodebuild_result.stdout, "watchos")
+    versions_dict = {
```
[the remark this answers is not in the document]
dierksen: Pretty sure it'll wind up the same at the end; it'll just fail if it can't find the right setting (since I'm not using `get`). Willing to revisit this though.

tools/toolchains/xcode_configure/xcode_configure.bzl
dierksen: It takes about a minute to run through all the SDKs in 3 Xcode versions from a clean/expunged repo.

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
dierksen: This is just parsing out stuff in the `swiftc` outputs that doesn't contain useful info (or like every other item looks like ```{"clang": "moduleName"}```)

dierksen: oh, and the imports file itself, which might be helpful for creating an umbrella header, but I'll follow up on that later.

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
dierksen: I think the optional installs might just be for simulators? I don't have either of the optional platforms installed on my machine, but it generates the BUILD files just fine 🤔

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
dierksen: Some swift modules depend on c modules; the reverse probably isn't true, but this is really just taking all the dependencies that `swiftc` found and connecting them (by adding the necessary suffix for collisions).

tools/toolchains/xcode_configure/xcode_configure.bzl
```diff
@@ -284,4 +296,5 @@ def xcode_configure(xcode_locator_label, remote_xcode_label = None):
         name = "local_config_xcode",
         xcode_locator = xcode_locator_label,
         remote_xcode = remote_xcode_label,
+        explicit_modules = True,
```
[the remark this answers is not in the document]
dierksen: No, this is just an on/off switch while I'm testing it out. 😅

tools/toolchains/xcode_configure/xcode_configure.bzl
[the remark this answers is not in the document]
jerrymarino: Cool, yeah I think the dict might make more sense: if we end up with dynamic access patterns!

tools/toolchains/xcode_configure/xcode_configure.bzl
jerrymarino: That's pretty good perf! Maybe we can decompose it for Xcode 14 stuff. They wrote a nice description about [recommending local=False](https://blog.bazel.build/2017/02/22/repository-invalidation.html#declare-your-repository-as-local-very-carefully) a few years ago BTW

.vscode/settings.json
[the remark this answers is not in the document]
jerrymarino: > In general, both of these seem like settings that would make more sense in a personal setting and not on a repo level.

Yeah well put - these both seem like GUI preferences

tools/toolchains/xcode_configure/xcode_sdk_frameworks.bzl
thiagohmcruz: k thx for clarifying!

.vscode/settings.json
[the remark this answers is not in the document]
ob: I'm not opposed to removing it. I think it was there because when you search for stuff you keep getting hits in the `bazel-out` directory but like @dierksen said, it can be a personal preference.

tools/toolchains/xcode_configure/xcode_configure.bzl
dierksen: I'll plan on splitting this up a bit more to avoid the issues with `local`. If we have to run the xcode locator twice, so be it, but it'd be nice to kill two birds with one stone. If nothing else, I can at least refactor this to reuse the same code.

On top of all this, I'm still not 100% married to the idea of keeping this as something we generate on the fly every time. Assuming the SDKs are fairly stable within major Xcode versions or across major OS versions for the respective platforms, it might be as simple as generating _enough_ of them to get good coverage and making it easy to update for future Xcode/platform releases. I'd need to download a few more versions of Xcode to test though ⏳

jerrymarino: > Assuming the SDKs are fairly stable within major Xcode versions or across major OS versions for the respective platforms, it might be as simple as generating enough of them 

Recently we've seen updates moving fast: shipping later updates in minor releases add new frameworks or dependencies quite often - or even from `13.3` -> `13.3.1` - with the M1 perf it should be a fair tradeoff 😅 

We can always circle back to it at a much later point as well - or do it orthogonally! I imagined the perf of genning these repos on the fly is a drop in the bucket compared to the wins we get

dierksen:
Happy 2-month birthday to this PR 🎂 

Remaining issues that I'm pondering:
- still not completely sure on `-import-underlying-module`: I don't completely understand the behavior for this flag, and it seems weirdly inconsistent between mixed-module and swift-only libraries
- need to mark sdk frameworks as extension-safe (assuming all of them are? I really don't know)
- all of the `xchammer-`-namespaced libraries need the frameworks added

jerrymarino (changes requested):

rules/library.bzl
```diff
@@ -975,7 +975,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             deps = deps + private_deps + lib_names + select({
                 "@build_bazel_rules_ios//:virtualize_frameworks": [framework_vfs_overlay_name_swift],
                 "//conditions:default": [framework_vfs_overlay_name_swift] if enable_framework_vfs else [],
-            }),
+            }) + (["@xcode_sdk_frameworks//:XCTest"] if testonly else []),
```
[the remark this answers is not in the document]
dierksen: `sdk_frameworks` only affects linking and it's only on `objc_*` AFAICT.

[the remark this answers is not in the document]
jerrymarino: Yeah today it's used for linking. However, the users already give us the SDK frameworks via `sdk_frameworks` so I imaged we might need / want to leverage [that API for explicit PCMs](https://github.com/bazel-ios/rules_ios/issues/484#issuecomment-1217157512)

Without such a switch it looks like the PR has to injects every SDK to the swift compile action? I imagined that wouldn't be ideal from a performance angle but will have to dig into it a bit more.

PR #566, opened 2022-09-21
Fixing git_repository assumption in xcbuildkit dep

ethan-gates:
The xcbuildkit_version.bzl file contains an assumption that the xcbuildkit repo was pulled via git_repository.  We pull it in via http_archive and as a result the rule has no commit field and bazel crashes during loading phase.

Error message:
```
File "/private/var/tmp/_bazel_egates/c4850d11fbeadd6bd3db71db3a1c0e4d/external/build_bazel_rules_ios/rules/third_party/xcbuildkit_version.bzl", line 26, column 12, in _get_ref
     if rule["commit"]:
Error: key "commit" not found in dictionary

### DIFF
--- a/rules/third_party/xcbuildkit_version.bzl
+++ b/rules/third_party/xcbuildkit_version.bzl
@@ -23,6 +23,8 @@ _repo_info = repository_rule(
 def _get_ref(rule):
     if not rule:
         return None
+    if rule["kind"] != "git_repository":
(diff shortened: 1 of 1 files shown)


chenxiao0228 (approved):

thiagohmcruz (approved):
Thx for the fix @ethan-gates!

PR #567, opened 2022-09-21
Adding application env var in xcodeproj

ethan-gates:
Currently ios_unit_test and ios_ui_test accept the `env` attribute [in rules_apple](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_ui_test-env) but ios_application does not.  In order to add this functionality, this PR adds an attribute to the xcodeproj rule specifically for application type products.  This allows the env var dict given to be propagated to the env var list in the app generated schemes

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -905,6 +905,9 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
 
         scheme_action_details = {"targets": [target_name]}
         env_vars_dict = {}
+        if product_type == "application":
+            env_vars_dict = ctx.attr.application_env_vars
+
         for (k, v) in getattr(target_info, "env_vars", ()):
             # Specific scheme can override the ones defined under env_vars here:
             if ctx.attr.scheme_existing_envvar_overrides.get(k, None):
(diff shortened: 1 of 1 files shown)


rules/legacy_xcodeproj.bzl
```diff
@@ -1312,6 +1315,7 @@ Additional LLDB settings to be added in each target's .lldbinit configuration fi
         "bazel_profile_enabled": attr.bool(default = False, mandatory = False),
         "disable_main_thread_checker": attr.bool(default = False, mandatory = False),
         "force_x86_sim": attr.bool(default = False, mandatory = False),
+        "application_env_vars": attr.string_dict(mandatory = False, default = {}, doc = "Add env vars to launch actions for app schemes")
```
jerrymarino, line 1318: Any reason not to use `scheme_existing_envvar_overrides` - it seems like if we add it for `application` it'll be a slippery slope - for every potential kind of bundle. 

FWIW XCHammer will allow you configure such options at the _target level_ - and have such [scheme options applying to a given target](https://github.com/pinterest/xchammer/blob/master/BazelExtensions/xchammerconfig.bzl#L47)

jerrymarino: Also, I'm wondering if it'd be better to add it to the point which allows _target_ config:
https://github.com/bazel-ios/rules_ios/blob/master/tests/ios/xcodeproj/BUILD.bazel#L32

ethan-gates: I think the scheme_existing_envvar_overrides requires there to be an env_vars kwarg in the target already.  I found including scheme_existing_envvar_overrides yields nothing by itself, and then when adding env_vars to ios_application the attribute is passed down to _objc and _swift rules which don't handle it.

I'll check out the target level config.  Are you thinking it would be another attribute like `app_environment` analogous to the existing `test_environment`?

mattrobmattrob:
@ethan-gates, is there still interest in getting this merged? Can you rebase it onto the latest, if so?

ethan-gates:
We're no longer blocked by this so I'll close

Issue #569, opened 2022-09-23
Transitive framework import causes release build fail


qyang-nj:
If we have an `apple_framework` transitively depends on `framework_import`, the build will fail with `--compilation_mode=opt`. In the following code, Module2 is not able to build. The same error will occur with `--features=swift.cacheable_swiftmodules`.

```
apple_dynamic_framework_import(
    name = "PrebuiltFramework",
)

apple_framework(
    name = "Module1",
    deps = [":PrebuiltFramework"],
)

apple_framework(
    name = "Module2",
    deps = [":Module1"],
)
```

With either `--compilation_mode=opt` or `--features=swift.cacheable_swiftmodules`, swift_library will stop passing `-serialize-debugging-options` to swiftc. Unfortunately, it seems we are relying on the search paths embedded in .swiftmodule for transitive framework imports. We should explicitly provide those search paths to the compiler.

**I have sample repro code  and steps [here](https://github.com/qyang-nj/BazelPlayground/tree/main/Issue0).**

qyang-nj:
I have a fix in mind and will validate it and submit a PR.

thiagohmcruz:
I guess we should re-open this give https://github.com/bazel-ios/rules_ios/pull/715

cc @mattrobmattrob @qyang-nj

qyang-nj:
Thanks for adding the tests. I'll take a look at this again.

PR #570, opened 2022-09-26
Fix transitive framework import in opt build

qyang-nj:
This PR fixes https://github.com/bazel-ios/rules_ios/issues/569.

`apple_*_(xc)framework_import` provides `_SwiftInteropInfo` instead of `SwiftInfo`. In `apple_framework_packaging`, when [we merge the SwiftInfo](https://github.com/bazel-ios/rules_ios/blob/b1499fc08a7e4b7a7ba238fb4c9b089e6648ecbd/rules/framework.bzl#L901), the info from the framework import get lost there.

To fix the problem, we can add `swift_clang_module_aspect` to the `transitive_deps` attribute of `apple_framework_packaging`, which converts `_SwiftInteropInfo` to `SwiftInfo`. `swift_library` does this for `deps` as well.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -24,7 +24,7 @@ load("@build_bazel_rules_apple//apple/internal:rule_support.bzl", "rule_support"
 load("@build_bazel_rules_apple//apple/internal:apple_toolchains.bzl", "AppleMacToolsToolchainInfo", "AppleXPlatToolsToolchainInfo")
 load("@build_bazel_rules_apple//apple/internal/utils:clang_rt_dylibs.bzl", "clang_rt_dylibs")
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "IosFrameworkBundleInfo")
-load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_common")
+load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_clang_module_aspect", "swift_common")
 load(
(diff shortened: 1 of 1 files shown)


luispadron:
One suggestion would be to add a fixture which tests this change.

qyang-nj:
The CI [fails building `//tests/ios/unit-test/test-imports-app:SomeFramework_swift`](https://github.com/bazel-ios/rules_ios/actions/runs/3130537379/jobs/5081225979#step:4:225), but I couldn't reproduce locally. I'm not sure if it's because of Xcode version, as I can't run Xcode 12.5.1 because of [this](https://github.com/bazel-ios/rules_ios/issues/571).

luispadron:
@qyang-nj this should possibly work now that 13.4.1 is the min Xcode

qyang-nj:
Great! I'll rebase my PR to the latest master later.

thiagohmcruz:
> One suggestion would be to add a fixture which tests this change.

+1 for this, can we add a test that exercises this? Maybe a structure similar to what is the issue or alternatively tweak/add to one of the existing tests?

thiagohmcruz (approved):
Thx for the fix!

qyang-nj:
@luispadron @thiagohmcruz  Thanks for reviewing. The issue this change fixes is exposed by `--compilation_mode=opt`. I just realized we don't have any tests [here](https://github.com/bazel-ios/rules_ios/blob/master/.github/workflows/tests.yml) for that flag. Do you think it's worth adding another tests to cover `--compilation_mode=opt`. If so, I can add it in a separate PR.

PR #572, opened 2022-10-04
Run buildifier fix, clean up unused variables/functions

dierksen:
Too many squiggles when I'm working on stuff, so took a quick cleanup pass.

I'd love to figure out how to suppress errors on the `xcspecs` file and add a bunch more docstrings, but this is a baby step 😄

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -180,10 +180,6 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
 
     return outputs
 
-def _add_to_dict_if_present(dict, key, value):
-    if value:
-        dict[key] = value
-
 def _concat(*args):
     arr = []
     for x in args:
@@ -843,7 +839,6 @@ def _apple_framework_packaging_impl(ctx):
 
     framework_files = _get_framework_files(ctx, deps)
     outputs = framework_files.outputs
-    inputs = framework_files.inputs
 
     # Perform a basic merging of compilation context fields
     compilation_context_fields = {}
@@ -1004,7 +999,7 @@ Valid values are:
             """,
         ),
         "_framework_packaging": attr.label(
-            cfg = "host",
+            cfg = "exec",
             default = Label(
                 "//rules/framework:framework_packaging",
             ),
@@ -1021,7 +1016,7 @@ that this target depends on.
         ),
         "_headermap_builder": attr.label(
             executable = True,
-            cfg = "host",
+            cfg = "exec",
             default = Label(
                 "//rules/hmap:hmaptool",
             ),
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -97,7 +97,7 @@ extend_modulemap = rule(
     doc = "Extends a modulemap with a Swift submodule",
 )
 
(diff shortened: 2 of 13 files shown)


rules/library.bzl
```diff
@@ -126,12 +126,10 @@ module {module_name} {{
 
 def _write_umbrella_header(
         name,
-        library_tools,
         generate_default_umbrella_header,
         public_headers = [],
-        private_headers = [],
         module_name = None,
-        **kwargs):
+        **_kwargs):
```
jerrymarino, line 132: Remove?

rules/library.bzl
```diff
@@ -177,7 +175,7 @@ FOUNDATION_EXPORT const unsigned char {module_name}VersionString[];
     )
     return destination
 
-def _generate_resource_bundles(name, library_tools, module_name, resource_bundles, platforms, **kwargs):
+def _generate_resource_bundles(name, library_tools, resource_bundles, platforms, **_kwargs):
```
jerrymarino, line 178: Remove _kwargs?

rules/library.bzl
```diff
@@ -194,7 +192,7 @@ def _generate_resource_bundles(name, library_tools, module_name, resource_bundle
         bundle_target_names.append(target_name)
     return bundle_target_names
 
-def _error_on_default_xcconfig(name, library_tools, default_xcconfig_name, **kwargs):
+def _error_on_default_xcconfig(name, default_xcconfig_name, **_kwargs):
```
jerrymarino, line 195: Remove _kwargs?

jerrymarino (approved):
Thanks for the cleanup 🙇 - I wonder why the CI isn't picking up on this.. Left a few inline to drop unused args

jerrymarino:
Also the `DEBUG` stuff in rules_ios is another one, I'm getting around this by filtering out this level of diags BTW:
```
build --ui_event_filters=-debug
```

rules/framework.bzl
```diff
@@ -1021,7 +1016,7 @@ that this target depends on.
         ),
         "_headermap_builder": attr.label(
             executable = True,
-            cfg = "host",
+            cfg = "exec",
```
thiagohmcruz, line 1019: Was curious to know the implications of this change but apparently `host` was deprecated and effectively is pretty much the same as `exec` now. For other reviewers / future reference: https://bazel.build/extending/rules#configurations

thiagohmcruz (approved):
Thx for doing this!

rules/framework.bzl
dierksen: Yes, this is one that got handled by the autofix; most of the rest were semi-manual.

jerrymarino:
@dierksen any chance you're using a newer buildifier fix than CI to catch this? If so we should probably upgrade to the more improved one to catch dead code or update settings

dierksen:
Doing another pass with the unused `kwargs` cleaned up. Also bumped the buildifier version; looks like they may have tweaked a few things since 5.0.1.

luispadron (approved):
🧼

PR #578, opened 2022-10-18
Remove duplication in xcode gen json file when using VFS

larrytaowang:
## Details
I noticed that with this flag `build --features apple.virtualize_frameworks` our Xcode generation becomes way slower, and `xcodegen.json` is much bigger, i.e., from ~20MB -> ~250MB.

I noticed that there are resources files duplication in the new xcodegen.json file. Let's say we have a pod A depends on pod B. Without VFS, in the `xcodegen.json` file, the source files of pod A is only included in the generated target A. With VFS, however, the source files of pod A are included in both target A and target B. That's why the new `xcodegen.json` file is so big.

With or without VFS, podB alwasys inlcude podA's source files. But the difference is that
- with VFS, a file from podA is like: `<source file Pods/A/A/foo.h>`
- without VFS, a file from podB is like `<generated file Pods/A/A/A.framework/Headers/foo.h>`

We have a filter like this so this is not an issue before VFS, since the souce files from other Pods are generated files. 
```
def _is_current_project_file(f):
    return f.is_source and _is_current_project_path(f.path)
```

With VFS since all files are source files, I think one possible solution is to filter the files that are not belong to the current package.

## Testing
Run the test scripts, there are two files removed: `common.pch` and `resource_bundle.plist`. I think this is a bug fix, since these two files are not part of the source code.

Compare the `xcodegen.json` file for our App. No change before/after this PR.


rules/legacy_xcodeproj.bzl
```diff
@@ -332,7 +335,7 @@ def _xcodeproj_aspect_impl(target, ctx):
                 asset_srcs += getattr(ctx.rule.files, attr, [])
         srcs = [f for f in srcs if _is_current_project_file(f)]
         non_arc_srcs = [f for f in non_arc_srcs if _is_current_project_file(f)]
-        asset_srcs = [f for f in asset_srcs if _is_current_project_file(f)]
+        asset_srcs = [f for f in asset_srcs if _is_current_package_file(f, target.label.package)]
```
jerrymarino, line 338: @larrytaowang I haven't fully looked at this but remembered there was 1-2 changes we needed. This said, at this point in the program - you have already a looped assets. You might dig into _why the upstream_ [code](https://github.com/bazel-ios/rules_ios/pull/578/files#diff-a15f63f3a76f05dbdfb6a9b817d2d5b1dfd28aa94bafc362eca9bd53f0dd77e4L332) processes it like this: looping `ctx.rule.files` in this way: e.g. why should it loop irrelevant sources and filter them down - better to fix at the root if possible IMO.

> Without VFS, in the xcodegen.json file, the source files of pod A is only included in the generated target A. With VFS, however, the source files of pod A are included in both target A and target B.

jerrymarino (approved):

rules/legacy_xcodeproj.bzl
larrytaowang: good question.  I think there is nothing wrong with the upstream code?

In the old way, the file `<generated file Pods/A/A/A.framework/Headers/foo.h>` is referencing an intermediate file, so it is generated. For VFS, the file `<source file Pods/A/A/foo.h>` is referencing a source file in the codebase, so it is a source file.

larrytaowang:
@jerrymarino I wonder is the CI still healthy? My `.xcodeproj Tests` child job failed consistently because 
```
diff -r ./tests/ios/xcodeproj/Test-LLDB-Logs-Project.xcodeproj/project.pbxproj ./tests/ios/xcodeproj/custom_output_path/Test-LLDB-Logs-Project.xcodeproj/project.pbxproj
45c45
< 			path = ../../..;
---
> 			path = ../../../..;
211c211
< 				BAZEL_WORKSPACE_ROOT = $SRCROOT/../../..;
---
> 				BAZEL_WORKSPACE_ROOT = $SRCROOT/../../../..;
268c268
< 				BAZEL_WORKSPACE_ROOT = $SRCROOT/../../..;
---
> 				BAZEL_WORKSPACE_ROOT = $SRCROOT/../../../..;
```

If I run the `./tests/xcodeproj-tests.sh --clean` command locally with the latest master, I also get this error, seems this is an existing issue not related to this PR?

larrytaowang:
i might forget to do the cleaning so I saw this error in my master branch. fixed the diff and now all the tests can pass

PR #580, opened 2022-10-20
Bump macos image, bazel, and carthage versions

dierksen:


rules/test/lldb/lldb_sim_runner.py
```diff
@@ -8,23 +8,21 @@
...
 import traceback
-import tempfile
-import shutil
-import json
 import contextlib
 
 logging.basicConfig(
     format="%(asctime)s.%(msecs)03d %(levelname)s %(message)s",
     datefmt="%Y-%m-%d %H:%M:%S",
-    level=logging.INFO)
```
jerrymarino, line 19: The formatting changes are adding a lot of extra stuff onto the Xcode bump PR - I'd suggest to break something like this out into a smaller one in general

.vscode/settings.json
```diff
@@ -3,4 +3,7 @@
         "bazel-*": true
     },
     "editor.formatOnSave": true,
+    "python.analysis.extraPaths": [
```
jerrymarino, line 6: Minor: would suggest moving this outside of the change

jerrymarino (commented):
LGTM in general - can you remove the `vscode` change and it's preferred python format changes? It has added a lot of extra "whitespace" noise to the PR and doesn't align with the pep8 style used in this repo

rules/test/lldb/lldb_sim_runner.py
dierksen: Yeah, I can revert this for now.

We have `buildifier` to lint/format all the starlark, which is the bulk of the repo, but the situation with bash, python, and whatever else we have floating around is pretty inconsistent. I was removing unused imports here, and `black` aggressively tried to take care of inconsistent quotes, dangling whitespace, etc.

luispadron:
I had started this in #511 but got blocked by Carthage issues

dierksen:
🤦 I had seen that PR and forgot about it until I got blocked on other stuff and then redid it. Let me look through there and see if maybe you caught something I missed 😅 

The good news is that I've got the carthage issues sorted! Now I'm just down to (hopefully) one failing test with the xcodeproj generators that I need to resolve.

repository_rules/framework_builder.bzl
```diff
@@ -105,20 +105,22 @@ def build_carthage_frameworks(
...
+CURRENT_XCODE_VERSION="$(xcodebuild -version | grep "Xcode" | cut -d' ' -f2 | cut -d'.' -f1)00"
+CURRENT_XCODE_BUILD=$(xcodebuild -version | grep "Build version" | cut -d' ' -f3)
+
+echo "EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_${CURRENT_XCODE_VERSION}__BUILD_${CURRENT_XCODE_BUILD} = arm64 arm64e armv7 armv7s armv6 armv8" >> $xcconfig
+ 
+echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_'${CURRENT_XCODE_VERSION}' = $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_$(XCODE_VERSION_MAJOR)__BUILD_$(XCODE_PRODUCT_BUILD_VERSION))' >> $xcconfig
+echo 'EXCLUDED_ARCHS = $(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT)__XCODE_$(XCODE_VERSION_MAJOR))' >> $xcconfig
+
+export XCODE_XCCONFIG_FILE="$xcconfig"
+"""
```
thiagohmcruz, line 123: What is the TL;DR for why the og carthage fix was not working? Is this name `carthage_xc12_workaround` still accurate?

thiagohmcruz (approved):

repository_rules/framework_builder.bzl
dierksen: It looks like the only difference is in the version parsing. Not sure what changed tbh, but it works now.

As for naming, it's broken in xcode 12+, and the workaround on carthage's docs still uses that name, so it's probably good for consistency/searchability.

luispadron (approved):
Sweet, excited to move us to 13! Thanks for finishing this up

luispadron:
Think we'll need to update github status checks to remove 12.5 and use the new Xcode check instead

data_generators/BUILD.bazel
```diff
@@ -12,16 +12,14 @@ genrule(
 cat <<EOS > $(OUTS)
 set -euo pipefail
 # xcode-locator prints json with keys as version names and values as paths to the developer dir for that xcode version
-readonly output=\\$$($(location @bazel_tools//tools/osx:xcode-locator) 2>/dev/null | grep : | sed -Ee s/.*'"(.+)": "(.+)".*'/'\\1.0|\\2'/g | sort --version-sort -k 1 -t '|' | tail -n 1)
-readonly xcode_version=\\$${output%%.0|*}
-readonly developer_dir=\\$${output#*|}
+readonly xcode_version=\\$$(xcodebuild -version | head -n1 | sed "s/Xcode //")
```
jerrymarino, line 15: Ideally this these lines / script would run automatically post analysis phase for a given version than be hardcoded to one in rules_ios.. This seems an OK given the current state of it.

jerrymarino (approved):
Thanks for the bump 🚢 👍 !

data_generators/BUILD.bazel
dierksen: Yeah, it could be more robust, but it's also just for generating this file to be committed here. If it were running as a repo rule and/or generating for multiple versions, it would be worth building some more infrastructure around it.

jerrymarino:
cc @mattrobmattrob @ob @ShawnMa16 ^ JFYI

dierksen:
> Think we'll need to update github status checks to remove 12.5 and use the new Xcode check instead

Took me a minute to figure out where to do that, but this is updated to the new check name.

PR #581, opened 2022-10-21
static vendored lib missing include paths

thiagohmcruz:
The `includes` attr of `objc_import` is supposed to propagate to targets that depend on it (see [docs](https://bazel.build/reference/be/objective-c#objc_import)) but the layers we have in `rules_ios` (`import_middleman` and `apple_framework_packaging`) were not doing that properly. 

This prevents depending on vendored static libs that might have `#include` statements that don't specify the module name, e.g., `#include <Foo.h>` instead of `#include <FooName/Foo.h>`. In many cases it's not an option to ask the vendor to change their code.

Regarding it's just incorrect to not propagate these includes and effectively break the `objc_import` intended behaviour. Additionally added one test case for this scenario. 

ps: Changes to `fetch_default_xcconfig` and `modulemap_generator` are meant to continue to allow consumers to specify custom attrs and also put back the public/private headers default values in `modulemap_generator`. I believe this is a regression introduced here: https://github.com/bazel-ios/rules_ios/pull/572 (cc @dierksen)

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -851,6 +851,10 @@ def _apple_framework_packaging_impl(ctx):
         direct = [],
         transitive = [getattr(dep[CcInfo].compilation_context, "defines") for dep in deps if CcInfo in dep],
     ))
+    objc_provider_utils.add_to_dict_if_present(compilation_context_fields, "includes", depset(
+        direct = [],
+        transitive = [getattr(dep[CcInfo].compilation_context, "includes") for dep in deps if CcInfo in dep],
+    ))
 
     # Compute cc_info and swift_info
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
--- a/rules/internal/objc_provider_utils.bzl
+++ b/rules/internal/objc_provider_utils.bzl
@@ -57,6 +57,8 @@ def _merge_dynamic_framework_providers(dynamic_framework_providers):
 
 def _merge_cc_info_providers(cc_info_providers, merge_keys = [
     "headers",
+    "includes",
+    "defines",
 ]):
     fields = {}
     for key in merge_keys:
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -97,7 +97,7 @@ extend_modulemap = rule(
     doc = "Extends a modulemap with a Swift submodule",
 )
 
-def _write_modulemap(name, umbrella_header = None, module_name = None, framework = False):
+def _write_modulemap(name, library_tools, umbrella_header = None, public_headers = [], private_headers = [], module_name = None, framework = False, **kwargs):
     basename = "{}.modulemap".format(name)
     destination = paths.join(name + "-modulemap", basename)
     if not module_name:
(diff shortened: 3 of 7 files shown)


rules/internal/objc_provider_utils.bzl
```diff
@@ -57,6 +57,8 @@ def _merge_dynamic_framework_providers(dynamic_framework_providers):
 
 def _merge_cc_info_providers(cc_info_providers, merge_keys = [
     "headers",
+    "includes",
+    "defines",
```
thiagohmcruz, line 61: Just adding `defines` here to be consistent with `apple_framework_packaging`, not 100% sure if we absolutely need it

rules/library.bzl
```diff
@@ -304,11 +304,12 @@ def _xcframework_slice(*, xcframework_name, identifier, platform, platform_varia
         # We need the path to the top-level slice directory so we can find headers, modulemaps and swiftmodules.
         slice_dir = paths.dirname(path)
         import_headers, import_module_map, import_swiftmodules = _xcframework_slice_imports(slice_dir)
+        includes = [paths.dirname(f) for f in import_headers]
         native.objc_import(
             name = resolved_target_name,
             archives = [path],
             hdrs = import_headers,
-            includes = ["%s/Headers" % slice_dir],
+            includes = includes,
```
thiagohmcruz, line 312: Assuming that the pattern here is always going to be `%s/Headers` is error prone. The test case that was already checked into `rules_ios` (`ObjcLib` under `xcframeworks` tests) is one example. The header is located under `Headers/ObjcLib`.

rules/library.bzl
```diff
@@ -797,9 +798,13 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
                 objc_hdrs.append(umbrella_header)
             module_map = library_tools["modulemap_generator"](
                 name = name,
+                library_tools = library_tools,
                 umbrella_header = paths.basename(umbrella_header),
+                public_headers = objc_hdrs,
+                private_headers = objc_private_hdrs,
```
thiagohmcruz, line 804: This was breaking our internal build due to missing default values, cc @dierksen

dierksen (approved):

luispadron (approved):
Thanks for fixing this up and supporting more than just `Headers/`, that seemed to be the standard but don't think there is any requirement around it.

thiagohmcruz:
Oh boy this CI is flaky

congt (approved):

rules/internal/objc_provider_utils.bzl
```diff
@@ -57,6 +57,8 @@ def _merge_dynamic_framework_providers(dynamic_framework_providers):
 
 def _merge_cc_info_providers(cc_info_providers, merge_keys = [
     "headers",
+    "includes",
```
jerrymarino, line 60: @thiagohmcruz do you happen to know if there are any implications of this change or other similar additions on build or IDE responsiveness?

rules/framework.bzl
```diff
@@ -851,6 +851,10 @@ def _apple_framework_packaging_impl(ctx):
         direct = [],
         transitive = [getattr(dep[CcInfo].compilation_context, "defines") for dep in deps if CcInfo in dep],
     ))
+    objc_provider_utils.add_to_dict_if_present(compilation_context_fields, "includes", depset(
+        direct = [],
```
jerrymarino, line 855: See below

jerrymarino (commented):
QQ if there is any additional impact of this change outside of fixing it

jerrymarino:
Sorry for the late review, just getting back into the swing of things here!

rules/internal/objc_provider_utils.bzl
thiagohmcruz: I didn't hit any issues while testing this change, if you have past context on this and are able to link issues / repro steps for something related to this I'm happy to investigate!

PR #589, opened 2022-11-04
Updating rules_apple to point to the latest on the bazel/5.x branch

mccorkill1:
When building an `apple_framework` as dynamic, if that framework depends on an `apple_resource_group` or an `apple_resource_bundle` the build will fail because `validate_all_resources_owned` is set to true in rules_apple. This https://github.com/bazelbuild/rules_apple/pull/1747 adds a `default_owner` so the check will pass and the `apple_framework` will build properly. This PR updates updates the rules_apple dependency to the latest on the `bazel/5.x` branch that includes this fix.

### DIFF
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -62,10 +62,10 @@ def rules_ios_dependencies():
     _maybe(
         github_repo,
         name = "build_bazel_rules_apple",
-        ref = "c17af58765d4903a8878d3d6c8004679de4fe8ab",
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):
I'm good with this. Let's get a review from @congt or @dierksen as well. 👍

dierksen (approved):

luispadron (approved):

mattrobmattrob:
Thanks for your first contribution, @mccorkill1!

PR #590, opened 2022-11-04
Fix issue merging xcconfig values that are lists

luispadron:
## Summary

`rules_ios` has support for providing `xcconfig` and `xcconfig_by_build_setting`, these two attributes allow `rules_ios` to translate them into the appropriate `copt`s for a given language. 

However, there is currently a bug when using `xcconfig_by_build_setting`. This manifests in issues with building due to missing `copt`s. 

For example, given this `BUILD` file:

```python
apple_framework(
    name = "Foo",
    xcconfig = {
        "GCC_PREPROCESSOR_DEFINITIONS": [
            "MACRO_A=1",
        ],
        "OTHER_SWIFT_FLAGS": [
            "-DMACRO_A",
        ],
    },
    xcconfig_by_build_setting = {
        "//my/flag:debug": {
            "GCC_PREPROCESSOR_DEFINITIONS": [
                "MACRO_B=1",
            ],
            "OTHER_SWIFT_FLAGS": [
                "-DMACRO_B",
            ],
        },
    },
)
```

The expected final `xcconfig` when using `//my/flag=debug` would look like:

```python
{
    "GCC_PREPROCESSOR_DEFINITIONS": [
        "MACRO_A=1",
        "MACRO_B=1",
    ],
    "OTHER_SWIFT_FLAGS": [
        "-DMACRO_A",
        "-DMACRO_B",
    ],
},
```

However, what we actually currently get is:

```python
{
    "GCC_PREPROCESSOR_DEFINITIONS": [
        "MACRO_B=1",
    ],
    "OTHER_SWIFT_FLAGS": [
        "-DMACRO_B",
    ],
},
```

This in turn causes the wrong `copt`s to be taken from the `xcconfig`s and can lead to other compile time issues as a result.

The main issue is how we merge `xcconfig` and `xcconfig_by_build_setting`, we are currently merging such that `xcconfig` is the default and any keys found in both get overriden by values in `xcconfig_by_build_setting`. This works as expected for `string -> string` xcconfig but has unexpected issues with `string -> [string]` xcconfigs. 

Instead, we should apply an approach where:

- If the key exists in `xcconfig` and `xcconfig_by_build_setting`
  - If the key is a string, use the latest value
  - If the key is a list, concatenate the values
- Merge non-matches normally

I've added a fixture in `//tests/ios/xcconfig` which fails to build on `master` but builds as expected with this PR.


rules/library/xcconfig.bzl
```diff
@@ -338,3 +325,30 @@ def copts_by_build_setting_with_defaults(xcconfig = {}, fetch_default_xcconfig =
...
+        A dictionary of Xcode build settings
+    """
+    merged_xcconfig = {}
+    for xcconfig in xcconfigs:
+        for (key, value) in dict(xcconfig).items():
+            if key in merged_xcconfig:
+                if types.is_list(value):
+                    merged_xcconfig[key] = merged_xcconfig[key] + value
+                else:
+                    merged_xcconfig[key] = value
```
dierksen, line 350: What about something like:
```suggestion
            if types.is_list(merged_xcconfig.get(key, None)) and types.is_list(value):
                merged_xcconfig[key] = merged_xcconfig[key] + value
```
I _think_ we should check that both sides are lists (not exactly sure what happens if you misconfigure something with mixed types), and this should also eliminate the need for the second `else`.

luispadron: That's better I'll update. I don't believe we fail if types are misconfigured, but the macros assume lists for some of these xcconfigs. 

We could add failing as a follow up, CocoaPods source has a list of all the list type xcconfig keys.

dierksen (approved):

tests/ios/xcconfig/WithDefinesObjc/WithDefines.m
```diff
@@ -0,0 +1,30 @@
+@import Foundation;
+
+@interface WithDefinesObjc : NSObject
+@end
+
+@implementation WithDefinesObjc
+
+#if !MACRO_A
+ thisShouldNotBeCompiled
```
jszumski, line 9: Would `#error "MACRO_A should be set via merged xcconfig"` more clearly indicate compilation is expected to fail when evaluating to true?

tests/ios/xcconfig/WithDefinesSwift/withDefines.swift
```diff
@@ -0,0 +1,22 @@
+enum WithDefinesSwift {
+  func foo() {
+#if !MACRO_H
+ thisShouldNotBeCompiled
```
jszumski, line 4: Same as above but `#error("MACRO_H should be set via merged xcconfig")`

luispadron: I'll update!

luispadron:
Added some unit tests for the `merge_xcconfigs` helper.

rules/library/xcconfig.bzl
```diff
@@ -338,3 +325,27 @@ def copts_by_build_setting_with_defaults(xcconfig = {}, fetch_default_xcconfig =
...
+    Args:
+        *xcconfigs: A list of dictionaries of Xcode build settings
+    Returns:
+        A dictionary of Xcode build settings
+    """
+    merged_xcconfig = {}
+    for xcconfig in xcconfigs:
+        for (key, value) in dict(xcconfig).items():
+            if types.is_list(merged_xcconfig.get(key, None)) and types.is_list(value):
+                merged_xcconfig[key] = merged_xcconfig[key] + value
```
thiagohmcruz, line 347: Can this contain duplicates? Should we remove if so?

thiagohmcruz (approved):
Just one small question! Thx for improving this and adding tests!

rules/library/xcconfig.bzl
luispadron: Id be inclined to not remove duplicates, that feels like user error we shouldn't correct for but happy to reconsider if folks feel different!

thiagohmcruz: Up to you! Agree that this would most likely apply to user error, so makes sense to me 👍

justinseanmartin (approved):

jerrymarino (approved):

luispadron:
Struggling with some really flaky CI again :(

Looks like some issues:
- timing out waiting for Bazel server to start, not sure if this a macOS-12 GitHub issue but looks new. 
- Some flaky tests in the Build and Test action, rerunning a few times resolved the issue

jerrymarino:
@luispadron after looking at this more it looks like there might be a gotcha on some of the semantics here.


Generally it wouldn't `concat` an array when merging 2 xcconfigs together given `import` / procedural semantics:

Given 
```
// a.xcconfig
FOO=(1,2)
```

```
// bar.xcconfig
import a.xcconfig
FOO=(3,4)
```

I'd expect it to take the last value of the xcconfig
```
FOO=(3,4)
```

### Import / procedural semantics

I assumed that is what we intend to have with `xcconfig` and `xcconfig_by_build_setting` / had prior to the change

### Xcode project -> target relationships

Potentially valid to have `project_xcconfig`.

### Podspec relationships

There's a number of merging going on here - this shouldn't belong here IMO. However, there might be some concat like relations or things people expected

PR #591, opened 2022-11-07
Add substitution support for Info.plist and XCConfigs

luispadron:
## Summary

This PR adds support for substituting values in `xcconfig` into `infoplists` and `xcconfig_by_build_setting` in to `infoplists_by_build_setting`. 

In the current rules we have most of what is required to allow substitution of xcconfigs into the plist files but users must create their own macros to do this. Since this is a very common use-case for xcconfigs, and given that `rules_ios` directly supports `xcconfig` and `xcconfig_by_build_setting` it would be a better user experience to automatically performing this variable substitution for users of `rules_ios`

### Problem

Given an `Info.plist`:

```xml
<plist version="1.0">
<dict>
    <key>CFBundleIdentifier</key>
    <string>${PRODUCT_BUNDLE_IDENTIFIER}</string>
    ...
    <key>CustonXCConfig1</key>
    <string>${CUSTOM_XCCONFIG_1}</string>
    <key>CustonXCConfig2</key>
    <string>$(CUSTOM_XCCONFIG_2)</string>
</dict>
</plist>
```

Given a `BUILD.bazel`:

```python
ios_application(
    name = "App",
    xcconfig = {
      "CUSTOM_XCCONFIG_1": "VALUE_1",
      "CUSTOM_XCCONFIG_2": "VALUE_2",
    },
    ...
)
```

It would be expected that we end up with a final `Info.plist`:

```xml
<plist version="1.0">
<dict>
    <key>CFBundleIdentifier</key>
    <string>${PRODUCT_BUNDLE_IDENTIFIER}</string>
    ...
    <key>CustonXCConfig1</key>
    <string>VALUE_1</string>
    <key>CustonXCConfig2</key>
    <string>VALUE_2</string>
</dict>
</plist>
```

However, we currently receive an Info.plist merge error:

```shell
ERROR: In target "//tests/ios/xcconfig:App"; unknown variable reference "${CUSTOM_XCCONFIG_1}" while merging plists (key: "CustonXCConfig1", value: "${CUSTOM_XCCONFIG_1}").
ERROR: /Users/lpadron/Development/rules_ios/tests/ios/xcconfig/BUILD.bazel:81:16 Bundling App failed: (Exit 1): plisttool failed: error executing command 
  (cd /private/var/tmp/_bazel_lpadron/7c5e30dd733c097d91026d4bca2f7997/sandbox/darwin-sandbox/1137/execroot/build_bazel_rules_ios && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=15.5 \
    XCODE_VERSION_OVERRIDE=13.4.1.13F100 \
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-dedf359489b7/bin/external/build_bazel_rules_apple/tools/plisttool/plisttool bazel-out/applebin_ios-ios_sim_arm64-dbg-ST-2dd91a95bcfe/bin/tests/ios/xcconfig/App-intermediates/Info.plist-root-control)
# Configuration: ee2650817cf108ae590de90e920dc3ebc71f739bfdb185cc497e135b023391fc
```

### Solution

The solution to the above is to use the existing tools in `rules` to perform variable substitution into the `infoplists` and `infoplists_by_build_setting`. 

We use `write_info_plist_if_needed` and `substitute_build_setting` along with the `xcconfig` and `xcconfig_by_build_setting` attributes provided to the rules to construct a dict of `infoplists_by_build_setting`. This dict is then `select`ed and passed as the `infoplists` attribute the relevant `rules_apple` rules which handle merging the plists.

### Depends on

- https://github.com/bazel-ios/rules_ios/pull/590

### Related

- https://github.com/bazelbuild/rules_apple/issues/226

### DIFF
--- a/rules/plists.bzl
+++ b/rules/plists.bzl
@@ -1,6 +1,92 @@
+"""
+Defines macros for working with plist files.
+"""
+
 load("@bazel_skylib//lib:types.bzl", "types")
+load("@bazel_skylib//lib:sets.bzl", "sets")
+load("@build_bazel_rules_ios//rules:substitute_build_settings.bzl", "substitute_build_settings")
 load("//rules:library.bzl", "write_file")
-load("//rules/library:xcconfig.bzl", "build_setting_name")
+load("//rules/library:xcconfig.bzl", "build_setting_name", "merge_xcconfigs")
+
+def process_infoplists(name, infoplists, infoplists_by_build_setting, xcconfig, xcconfig_by_build_setting):
+    """
+    Constructs substituted_plists by substituting build settings from an xcconfig dict into the variables of a plist.
+
+    Args:
+        name: The name of the target that the plist is being generated for.
+        infoplists: The plist files to be manipulated
+        infoplists_by_build_setting: A dictionary of infoplists keyed by config_setting, merged with infoplists created here.
+        xcconfig: the default build settings to expand the infoplists with
+        xcconfig_by_build_setting: the build settings grouped by config_setting to expand the infoplists with
+
+    Returns:
+        A selectable dict of the substituted_plists grouped by config_setting
+    """
+    infoplists_by_build_setting = dict(infoplists_by_build_setting)
+    xcconfig_by_build_setting = dict(xcconfig_by_build_setting)
+
+    # Final substituted infoplists_by_build_setting dict
+    substituted_infoplists_by_build_setting = {}
+
+    # Default infoplists are any defined in `infoplists` OR in `infoplists_by_build_setting` with the key `//conditions:default`
+    default_infoplists = infoplists_by_build_setting.pop("//conditions:default", infoplists)
+
+    # Default xcconfig is the one defined in `xcconfig` OR in `xcconfig_by_build_setting` with the key `//conditions:default`
+    default_xcconfig = xcconfig_by_build_setting.pop("//conditions:default", xcconfig)
+
+    # Substitute the default infoplists
+    substituted_infoplists_by_build_setting["//conditions:default"] = [
+        substituted_plist(
+            name = "%s.%s.infoplist_substituted" % (name, idx),
+            plist = plist,
+            xcconfig = default_xcconfig,
+        )
+        for idx, plist in enumerate(write_info_plists_if_needed(name = name, plists = default_infoplists))
+    ]
+
+    # Collect a set of config settings to iterate over
+    config_setting_names = sets.make(infoplists_by_build_setting.keys() + xcconfig_by_build_setting.keys())
+
+    # Substitute the infoplists_by_build_setting
+    for config_setting_name in sets.to_list(config_setting_names):
+        name_suffix = build_setting_name(config_setting_name)
+        infoplists_for_build_setting = infoplists_by_build_setting.get(config_setting_name, default_infoplists)
+        xccconfig_for_build_setting = xcconfig_by_build_setting.get(config_setting_name, {})
+        xcconfig_for_plist = merge_xcconfigs(default_xcconfig, xccconfig_for_build_setting)
+
+        # Substitute the build settings into the plists for this config_setting
+        substituted_infoplists_by_build_setting[config_setting_name] = [
+            substituted_plist(
+                name = "%s.%s.%s.infoplist_substituted" % (name, name_suffix, idx),
+                plist = plist,
+                xcconfig = xcconfig_for_plist,
+            )
+            for idx, plist in enumerate(write_info_plists_if_needed(name = "%s.%s" % (name, name_suffix), plists = infoplists_for_build_setting))
+        ]
+
+    return substituted_infoplists_by_build_setting
+
+def substituted_plist(name, plist, xcconfig):
+    """
+    Substitutes build settings from an xcconfig dict into the variables of a plist.
+
+    Args:
+      name: The name of the plist.
+      plist: The plist file to substitute into.
+      xcconfig: The xcconfig variables for substitution.
+
+    Returns:
+        The plist target with the substituted variables.
+    """
+    if not plist:
+        return None
+
+    sub_plist = _substitute_plist_vars(name, plist, xcconfig)
+
+    if sub_plist:
+        return sub_plist
+    else:
+        return plist
 
 def write_info_plists_if_needed(name, plists):
     """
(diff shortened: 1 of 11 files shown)


jszumski (approved):

jerrymarino (commented):
I think it's more expected that the API would do what you've got done here - No major feedback on this really.

What should people do here that already have the macros: it looks like this code is backwards compatible?

Also - I filed this one out of some of the issue I noticed with this over the years: https://github.com/bazel-ios/rules_ios/issues/594

jerrymarino (approved):
Thanks for the PR - giving this a tennative 🚢 and will review again after you get it out of the draft

luispadron:
> What should people do here that already have the macros: it looks like this code is backwards compatible?

@jerrymarino yeah, should continue to work you'll just have extra info plists to merge (since they'd be substituting and so would rules_ios) but this shouldn't be an issue. 

Thanks for filing the issue! Definitely agree that doing this during loading phase makes it difficult to use since macros have to handle the select via dict APIs.

jerrymarino:
@luispadron sure, I'll give this another run through this morning - thanks again for the improvement!

luispadron:
Spoke too soon, looks like PR is still modifying some fixtures which is unexpected I think. Looking into it!

luispadron:
I've resolved the issues where it was creating an extra plist, and thus failing the Xcode project tests. This should be ready now

rules/plists.bzl
```diff
@@ -1,6 +1,92 @@
...
+
+    # Collect a set of config settings to iterate over
+    config_setting_names = sets.make(infoplists_by_build_setting.keys() + xcconfig_by_build_setting.keys())
+
+    # Substitute the infoplists_by_build_setting
+    for config_setting_name in sets.to_list(config_setting_names):
+        name_suffix = build_setting_name(config_setting_name)
+        infoplists_for_build_setting = infoplists_by_build_setting.get(config_setting_name, default_infoplists)
+        xccconfig_for_build_setting = xcconfig_by_build_setting.get(config_setting_name, {})
+        xcconfig_for_plist = merge_xcconfigs(default_xcconfig, xccconfig_for_build_setting)
```
jerrymarino, line 55: @luispadron The issue I hit with this lines is ambiguous config results.. 

Is there a way to move plist computation / merging till a later point, say analysis phase? That is more robust overall and fixes the issues I have encountered

Imagine you arrive at:
`infoplists_by_build_setting =  { "foo" : [X] }`
`xcconfig_by_build_setting = { "bar" : [Y] }`


But `foo` is actually not mutually exclusive with `bar` from a configuration perspective.. 


I wonder if you can tweak this logic to lift some of the loading phase stuff like `merge_xcconfigs` to the analysis phase.  


For instance:
```
substituted_plist(
   name=name,
   plist=plist,
   base_xcconfig=xcconfig, 
   xcconfig=xcconfig_by_build_setting, 
)
```

rules/plists_test.bzl
```diff
@@ -0,0 +1,176 @@
...
+    xcconfig = {
+        "CUSTOM_PRODUCT_VERSION": "9.9.9",
+    }
+
+    # Given xcconfigs by build setting
+    xcconfig_by_build_setting = {
+        "//:debug": {
+            "CUSTOM_PRODUCT_VERSION": "7.7.7",
+        },
+        "//:release": {
```
jerrymarino, line 46: People can feed this rule code where `xcconfig_by_build_setting`  have different exclusivity than the selects in `infoplists_by_build_setting`.

The related code only works because it only keys on the xcconfig which isn't great from an API perspective

```
 58     for (build_setting, build_setting_xcconfig) in xcconfig_by_build_setting.items():
 59         name_suffix = build_setting_name(build_setting)
 60
```

jerrymarino (approved):
I think this is a good but will need some changes to use this for a few cases I have. LMK what you think - we fix "push this to the analysis phase" bit, here or later.

rules/plists.bzl
jerrymarino: It also removes merging in the loading phase and making assumptions based on the names 😄

luispadron:
Will follow up adding analysis time support for this

rules/plists.bzl
```diff
@@ -1,6 +1,92 @@
...
+        A selectable dict of the substituted_plists grouped by config_setting
+    """
+    infoplists_by_build_setting = dict(infoplists_by_build_setting)
+    xcconfig_by_build_setting = dict(xcconfig_by_build_setting)
+
+    # Final substituted infoplists_by_build_setting dict
+    substituted_infoplists_by_build_setting = {}
+
+    # Default infoplists are any defined in `infoplists` OR in `infoplists_by_build_setting` with the key `//conditions:default`
+    default_infoplists = infoplists_by_build_setting.pop("//conditions:default", infoplists)
```
jerrymarino, line 32: @luispadron is it fair to say that this logic just discards `infoplists` if you provide it with an `info_plists_by_build_setting["//conditions:default"] != None`?

It looks like it might be a breaking change / is divergent from this API is working in other contexts:
https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl#L469

rules/plists.bzl
```diff
@@ -1,6 +1,92 @@
...
+
+    # Default xcconfig is the one defined in `xcconfig` OR in `xcconfig_by_build_setting` with the key `//conditions:default`
+    default_xcconfig = xcconfig_by_build_setting.pop("//conditions:default", xcconfig)
```
jerrymarino, line 35: Similar as above

jerrymarino (commented):
One other thing I noticed, I was able to rework some code that used this but this one is probably worth addressing

rules/plists.bzl
luispadron: Yes thats what it'll do but thats the same way it worked before my changes: https://github.com/bazel-ios/rules_ios/blob/9c621595602e2cd6bdda711a20fba1c5e0bc2fd2/rules/plists.bzl#L41

The `infoplists_by_build_setting` and `xcconfig_by_build_setting` are local copies so `pop` shouldn't affect downstream

jerrymarino: Ok thanks for clarifying 🤯

PR #593, opened 2022-11-07
[Xcode 14.x.x] Remove INFOPLIST_FILE setting from the legacy xcodeproj rule

justinseanmartin:
In Xcode 14, we see that the invocation of `builtin-infoPlistUtility` sometimes will come after the 'Build with bazel' script phase. When this happens, the plist value will get overwritten by the one we've specified in our rules. As we always provide a base Info.plist (as required by rules_apple) if none is specified, this stub seems unnecessary.

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -860,7 +860,6 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["BAZEL_LLDB_INIT_FILE"] = lldbinit_file
 
         if product_type == "application" or product_type == "app-extension":
-            target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
             target_settings["PRODUCT_BUNDLE_IDENTIFIER"] = target_info.bundle_id
 
         if product_type == "bundle.unit-test":
@@ -1177,7 +1176,6 @@ def _xcodeproj_impl(ctx):
             "$(print_json_leaf_nodes_runfiles)": " ".join(print_json_leaf_nodes_runfiles),
             "$(build_wrapper_path)": ctx.executable.build_wrapper.short_path,
             "$(build_wrapper_runfile_short_paths)": " ".join(build_wrapper_runfile_paths),
-            "$(infoplist_stub)": ctx.file.infoplist_stub.short_path,
             "$(output_processor_path)": ctx.file.output_processor.short_path,
             "$(workspacesettings_xcsettings_short_path)": ctx.file._workspace_xcsettings.short_path,
             "$(ideworkspacechecks_plist_short_path)": ctx.file._workspace_checks.short_path,
@@ -1196,7 +1194,6 @@ def _xcodeproj_impl(ctx):
                          ctx.files.index_import +
                          ctx.files.ld_stub +
                          ctx.files.swiftc_stub +
-                         ctx.files.infoplist_stub +
                          ctx.files.print_json_leaf_nodes +
                          ctx.files._workspace_xcsettings +
                          ctx.files._workspace_checks +
@@ -1283,7 +1280,6 @@ For a full list, see under keys of `PRODUCT_TYPE_UTI` under
 https://www.rubydoc.info/github/CocoaPods/Xcodeproj/Xcodeproj/Constants
 """),
         "_xcodeproj_installer_template": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:xcodeproj-installer.sh"), allow_single_file = ["sh"]),
-        "infoplist_stub": attr.label(executable = False, default = Label("//rules/test_host_app:Info.plist"), allow_single_file = ["plist"]),
         "_workspace_xcsettings": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:WorkspaceSettings.xcsettings"), allow_single_file = ["xcsettings"]),
         "_workspace_checks": attr.label(executable = False, default = Label("//tools/xcodeproj_shims:IDEWorkspaceChecks.plist"), allow_single_file = ["plist"]),
         "output_processor": attr.label(executable = True, default = Label("//tools/xcodeproj_shims:output-processor.rb"), cfg = "exec", allow_single_file = True),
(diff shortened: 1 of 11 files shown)


thiagohmcruz (approved):

rules/legacy_xcodeproj.bzl
```diff
@@ -860,7 +860,6 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["BAZEL_LLDB_INIT_FILE"] = lldbinit_file
 
         if product_type == "application" or product_type == "app-extension":
-            target_settings["INFOPLIST_FILE"] = "$BAZEL_STUBS_DIR/Info-stub.plist"
```
jerrymarino, line 863: Does this cause any issues on Xcode 13 / legacy build system?

jerrymarino: Ok as long as you tested it out. Perhaps it was required to have this set in vintage releases but not on modern cuts

jerrymarino (approved):
Thanks for the fix - minor suggestion is to prefix the PR name with `[Xcode 14]` so folks know this is relate  to the bump

mattrobmattrob:
This seems to now be picking some `INFOPLIST_FILE` value via XcodeGen for our configuration:
```
$ bazel-bin/RedditBazel.xcodeproj/project.pbxproj
...
		507D44E145CED297EA2E7B11 /* Release */ = {
			isa = XCBuildConfiguration;
			buildSettings = {
				BAZEL_BIN_SUBDIR = /RedditApp;
				BAZEL_BUILD_TARGET_LABEL = "RedditApp:RedditApp";
				...
				INFOPLIST_FILE = ../../../Extensions/stickers/Info.plist;
...
		7C4ACEED3259E6BA3FE99D67 /* Debug */ = {
			isa = XCBuildConfiguration;
			buildSettings = {
				BAZEL_BIN_SUBDIR = /RedditApp;
				BAZEL_BUILD_TARGET_LABEL = "RedditApp:RedditApp";
				...
				INFOPLIST_FILE = ../../../Extensions/stickers/Info.plist;
...
```
Confirmed it's not in the XcodeGen input though:
```
$ grep -rn "INFOPLIST" bazel-bin/RedditBazel-xcodegen.json
# empty
```

What's the resulting value for you all now in Xcode, @justinseanmartin? This has broken app install in Xcode 14.0.1 since it's using the stickers `Info.plist` for us.

justinseanmartin:
Sorry for the delay, got buried today. That is super weird, our generated project has the value unset:
image

justinseanmartin:
Sorry for the breakage, feel free to DM me if I can help provide some info. From what I saw, this setting wasn't being used for anything, as the plist file in the INFOPLIST_FILE build setting gets (or is supposed to get) stomped by the one provided in the Bazel script phase.

The problem we ran into with Xcode 14.1 (not sure about 14.0) was that the "Processing Info-stub.plist" command would run after and overwrite the one generated by Bazel in the non-clean DerivedData case. Were you not seeing that on your end?

jerrymarino:
@justinseanmartin what do you think about better making the `Stub-Info.plist` a dependency of the `bazel build` pre phase.. That way it'd fix the "Processing Info-stub.plist" phase and work with the other constraint longer term of seeing the Plist wired up in the GUI

@mattrobmattrob basically we'd be doing prior to this commit:
- adding `$(BAZEL_STUBS_DIR)/Info-stub.plist` - as a dependency to `build with bazel` in the GUI

jerrymarino:
>This seems to now be picking some INFOPLIST_FILE value via XcodeGen for our configuration:

Because of this issue raised here - we 100% need to give it a stub. I don't think there is a test case of this in rules_ios or it's so obvious. Proofing out the original fix: hooking the users provided plist up so you can navigate to that into Xcode is the way to go

mattrobmattrob:
For the record, I plan to try to dig deeper into why a/the wrong plist is being grabbed for our targets in the next week or so. Haven't had time to dig deeper and we don't rely on the project gen functionality other than for prototypes's sake at this point.

justinseanmartin:
Jerry just pointed me at https://github.com/yonaskolb/XcodeGen/blob/e7f753785e6ee135e5ea77268d149e9cd10bb7ed/Sources/XcodeGenKit/PBXProjGenerator.swift#L1406 which I think explains it finding a file named `Info.plist` within the sources to use. In our case, the apps I was testing used a plist like `SPOS-Info.plist`, so it wasn't getting matched by that logic. I'm looking to try and repro locally as well.

mattrobmattrob:
That's exactly where I had ended up as well the other day. 👍 Was gonna run the XcodeGen fork locally on the generated XcodeGen JSON for our project to really answer it/hack around it for now. Most of ours are Bazel artifacts from some `genrule` at this point.

justinseanmartin:
I tried locally with an app target that contains one named `Info.plist`, but the `INFOPLIST_FILE` still wasn't set. Same thing for an app extension that we've got, both seem fine on our end AFAICT.

luispadron:
Catching up now but I think this is why these tests are failing on one of my PRs [here](https://github.com/bazel-ios/rules_ios/actions/runs/3432605641/jobs/5722057193)

jerrymarino:
Cool thanks for confirming! Given this, I'd still suggest the [proposed fix](https://github.com/bazel-ios/rules_ios/pull/593#issuecomment-1309303179) for this until we remove the stubbed tools.

@justinseanmartin it looks like you'll need to plumb a file called `Info.plist` in a very specific way to trigger the problem here. Based on the logic and his comment it looks like there are a few aspects at play that might be divergent. One being, In the comment the projects might not be at the root of the tree:
```
    INFOPLIST_FILE = ../../../Extensions/stickers/Info.plist;
```

If I had to guess w/o seeing it, it's probably triggered in one of these lines:
```
            // If we haven't yet looked for the default info plist, try doing so
            if searchForDefaultInfoPlist {
                searchForDefaultInfoPlist = false

                if let plistPath = getInfoPlist(target.sources) {
                    let basePath = projectDirectory ?? project.basePath.absolute()
                    let relative = (try? plistPath.relativePath(from: basePath)) ?? plistPath
                    defaultInfoPlist = relative.string
                }
            }
```

justinseanmartin:
Thanks for the repro case @luispadron! If I can't fix it today, I'll put up a PR to revert my change to get master healthy for other folks. Sound like a plan?

PR #597, opened 2022-11-10
Bump xchammer and xcbuildkit

thiagohmcruz:
Requires:
* https://github.com/jerrymarino/xcbuildkit/pull/51
* https://github.com/bazel-ios/xchammer/pull/34

Besides the bumps moves `bazel_build_service_config` to be a "top level" attr since it's not part of `project_config` anymore.

- [ ] Bump `xchammer` and `xcbuildkit` to `master` before landing

PR #600, opened 2022-11-14
Bump xchammer and xcbuildkit (2)

thiagohmcruz:
Same as https://github.com/bazel-ios/rules_ios/pull/597 that was merged by mistake and it's now ready to be merged for real.

Requires:
* https://github.com/jerrymarino/xcbuildkit/pull/51
* https://github.com/bazel-ios/xchammer/pull/34

Besides the bumps moves `bazel_build_service_config` to be a "top level" attr since it's not part of `project_config` anymore.

- [x] Bump `xchammer` and `xcbuildkit` to `master` before landing

### DIFF
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -46,8 +46,8 @@ def xcodeproj(name, **kwargs):
                 generate_xcode_schemes = generate_xcode_schemes,
                 paths = ["**"],
                 xcconfig_overrides = xcconfig_overrides,
-                bazel_build_service_config = bazel_build_service_config,
             ),
+            bazel_build_service_config = bazel_build_service_config,
             targets = kwargs.get("deps", []),
         )
     else:
--- a/rules/xcodeproj.bzl
+++ b/rules/xcodeproj.bzl
@@ -5,18 +5,20 @@ load("@xchammer//:BazelExtensions/xcodeproject.bzl", xchammer_xcodeproj = "xcode
 load("@xchammer//:BazelExtensions/xchammerconfig.bzl", "bazel_build_service_config", "project_config")
 load("//rules:library.bzl", "GLOBAL_INDEX_STORE_PATH")
 
-def _patch_bazel_build_service_config(kwargs_bazel_build_service_config):
-    """Overriding to set some default values for now.
+def _patch_bazel_build_service_config(name, kwargs_bazel_build_service_config):
+    """Adds sensible defaults
 
-    We need to better expose the option to
-    customize `GLOBAL_INDEX_STORE_PATH` and expecting to iterate a bit on `indexing_data_dir`
-    and run some tests, once it's battle tested will revert this and allow customers to fully configure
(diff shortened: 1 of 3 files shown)


congt (approved):

rules/xcodeproj.bzl
```diff
@@ -5,18 +5,20 @@ load("@xchammer//:BazelExtensions/xcodeproject.bzl", xchammer_xcodeproj = "xcode
...
+    It's really important to align the index stores with `xcbuildkit` so 'GLOBAL_INDEX_STORE_PATH' is always set
+    for now since this is how `rules_ios` is setup. Later on we can make this more configurable.
     """
     if kwargs_bazel_build_service_config:
         return bazel_build_service_config(
-            bep_path = kwargs_bazel_build_service_config.bepPath if kwargs_bazel_build_service_config.bepPath else "/tmp/bep.bep",
+            bep_path = kwargs_bazel_build_service_config.bepPath if kwargs_bazel_build_service_config.bepPath else "/tmp/{}.bep".format(name),
             index_store_path = "$SRCROOT/%s" % GLOBAL_INDEX_STORE_PATH,
-            indexing_data_dir = kwargs_bazel_build_service_config.indexingDataDir if kwargs_bazel_build_service_config.indexingDataDir else "/tmp/xcbuildkit-data/indexing",
+            indexing_data_dir = kwargs_bazel_build_service_config.indexingDataDir if kwargs_bazel_build_service_config.indexingDataDir else "/tmp/xcbuildkit-data/{}/indexing".format(name),
```
jerrymarino, line 21: Should probably in a persistent directory?
 `~/Library/Developer/Xcode/Indexing/Project-$(hash code name / dir)` 
where `$(hash code - factors into the project directory / dir)`.. Or maybe just levers `$DERIVED_DATA` build setting..

Otherwise it looks like it if you put units/records there:
- will blow out the index on reboot
- it can have issues when using multiple checkouts as it stands


Perhaps you could align them to an xcconfig that loads the path specific variable at build time?

jerrymarino (commented):
Minor suggestion inline around reboots/mutli-checkout

jerrymarino (approved):
@thiagohmcruz see inline - I think it's fine to land for now since you've tested it in this configuration - we can circle back on this at a later point

rules/xcodeproj.bzl
jerrymarino: > Atm indexing_data_dir is just for the aspect outputs that are stored to be fast-loaded 

Ah ok - makes sense. Such a schema could be ripe for conflicts in multi-checkout usages if the key space doesn't bake in the project path though. 

I think we can circle back on it / give it more consideration later. If it's going to be a project level config - then bake it in the project if you can. Otherwise, it's probably worth more consideration if it's _build_config.

thiagohmcruz: Yeah, this will allow us to start end-to-end testing so focusing mostly on that for now. Pretty sure there's lots of optimizations we can make from here but will focus on getting to a point where this is actually testable "as a user" so we can iterate. Thx for flagging!

PR #601, opened 2022-11-14
Explicitly unset the plist value, so XcodeGen doesn't go searching for one

justinseanmartin:
This is a followup to the fallout from landing #593, as there were some issues when running without a sandbox. The problem is that [XcodeGen will go looking for a plist value](https://github.com/yonaskolb/XcodeGen/blob/e7f753785e6ee135e5ea77268d149e9cd10bb7ed/Sources/XcodeGenKit/PBXProjGenerator.swift#L1406) if one isn't explicitly specified. This change will short circuit the logic in XcodeGen, but still ensures Xcode doesn't add its own step for processing the Plist (which on Xcode 14 would sometimes stomp the plist we're creating as part of the app bundle in Bazel).


mattrobmattrob (approved):
Confirmed it worked in our project with a variety extensions embedded in our app. Thanks, @justinseanmartin! 🎉

luispadron (approved):

rules/legacy_xcodeproj.bzl
```diff
@@ -860,6 +860,7 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
         target_settings["BAZEL_LLDB_INIT_FILE"] = lldbinit_file
 
         if product_type == "application" or product_type == "app-extension":
+            target_settings["INFOPLIST_FILE"] = ""
```
jerrymarino, line 863: @justinseanmartin would you adding a comment about _why_ this is there? I think we know the other suggested fixes so LGTM

jerrymarino (approved):
👏

PR #602, opened 2022-11-15
Revert "Fix issue merging xcconfig values that are lists (#590)"

jerrymarino:
This reverts commit f304c3721b23c426a37b33ceb8e739586c8219f1.

This mixes other kinds of semantics and would suggest to keep "import" semenatics of thses.

Ideally something like cocoapods "concat" is handled elsewhere: fixing the BUILD file generator / cocoapods-bazel to do `$(inherited)` or other merging. For instance if you need to concat or do inheritance of 2 settings - you'd have to concatenate out of this program:

```
xcconfig = { debug: [A] } <- optimization would be to remove this
xcconfig_by_build_setting = {
   debug: [A] + [B]
   release: [A] + [C]
}
```

### DIFF
--- a/rules/library/xcconfig.bzl
+++ b/rules/library/xcconfig.bzl
@@ -329,11 +329,7 @@ def copts_by_build_setting_with_defaults(xcconfig = {}, fetch_default_xcconfig =
 def merge_xcconfigs(*xcconfigs):
     """Merges a list of xcconfigs into a single dictionary
 
-    Uses some heuristics to merge xcconfigs in a way that is compatible with Xcode's behavior:
-
-    - If a key is present in multiple xcconfigs
-        - If the value is a list, the values are concatenated
-        - If the value is a string, the last xcconfig value is used
+    Overrides keys from the first xcconfig with the values from the latest one if they match.
 
     Args:
         *xcconfigs: A list of dictionaries of Xcode build settings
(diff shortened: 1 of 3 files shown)


luispadron (approved):

luispadron:
I was going to revert just the concat i.e. just this line: https://github.com/bazel-ios/rules_ios/blob/master/rules/library/xcconfig.bzl#L346 so that we could keep the tests I added but this may be safer

thiagohmcruz:
> I was going to revert just the concat i.e. just this line: https://github.com/bazel-ios/rules_ios/blob/master/rules/library/xcconfig.bzl#L346 so that we could keep the tests I added but this may be safer

+1 for keeping the tests, maybe it's easier if @luispadron pushes a commit here with the expected changes as the author?

luispadron:
@thiagohmcruz @jerrymarino Pushed the changes to keep the test and logic how it used to be

thiagohmcruz:
> @thiagohmcruz @jerrymarino Pushed the changes to keep the test and logic how it used to be

@luispadron I think Jerry is at BazelCon today so probably won't be able to land this. Mind taking a look at the CI failures?

luispadron:
Oops forgot to update the tests. I'm also at BazelCon so I'll get that fixed soon

thiagohmcruz:
> Oops forgot to update the tests. I'm also at BazelCon so I'll get that fixed soon

Oof sorry, I didn't know you were attending too! No rush, worst case we can punt this until you're back. Main reason was to bump our internal repo but there's nothing super urgent that we need landed today. Anytime this week would be great.

luispadron:
@thiagohmcruz updated! Tests should pass unless we run into the cc_config flakiness

thiagohmcruz:
Thank you @luispadron!

PR #605, opened 2022-11-17
Fix Xcode 14 missing `libXCTestBundleInject.dylib` when running tests

justinseanmartin:
Xcode 14 no longer supports the legacy build system, so we're silently
being upgraded to using the new build system. One issue that surfaces
from this is that the `libXCTestBundleInject.dylib` starts being
injected into the test host app as a step in building the *app host*
instead of in building the *test bundle* (as it used to). This is
problematic, because we `rsync --delete` the app bundle as part of the
Bazel install script after the dylib is copied, and thus it gets purged.

One interesting finding is that even without this fix, it works fine in
Xcode 14 with iOS 15/16 simulators. This appears to be due to some
difference in dyld behavior between sims, where it is able to successfully
fall back and load this library from the SDK dir instead of embedded in
the host app, even though the xctest run explicitly contains
DYLD_INSERT_LIBRARIES w/ `__TESTHOST__/Frameworks/libXCTestBundleInject.dylib`.

We could go about ensuring `libXCTestBundleInject.dylib` is present in a
couple ways:
* Let Xcode copy the dylib into the output directory from the 'Copy
  libXCTestBundleInject.dylib' task and assume Xcode's will always place
  it there when needed. This avoids duplicative copying and is a simple
  change.
* Add `libXCTestBundleInject.dylib` as a dependency of the app host in
  the build rules, but presumably only when we're building for testing.
  Maybe we'd need a test app host wrapper for applications rules that
  test targets would take a dependency on.
* See if there is some way to get dyld to behave differently on iOS 14
  to allow loading the library from the SDK path.

For this change, we've opted to go the first route. If this build rule
were going to be longer lived (instead of being replaced) and there was
more logic we'd want to run specifically for the case of test app hosts,
I'd consider going with the second approach.

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
@@ -92,6 +92,7 @@ for input in "${input_options[@]}"; do
     else
         rsync \
             --recursive --chmod=u+w --delete --copy-links \
+            --exclude 'Frameworks/libXCTestBundleInject.dylib' \
             "$input" "$output" >"$BAZEL_DIAGNOSTICS_DIR"/rsync-stdout-"$DATE_SUFFIX".log 2>"$BAZEL_DIAGNOSTICS_DIR"/rsync-stderr-"$DATE_SUFFIX".log
     fi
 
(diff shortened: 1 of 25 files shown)


thiagohmcruz (approved):
Thx for the detailed explanation. I think the first approach makes sense both thinking about mid-long term story for proj generation plus one more thing to unblock folks bumping Xcode sooner than later.

PR #606, opened 2022-11-17
Cancel old PR CI jobs when PR is updated

dierksen:
This should reduce load on CI when people like me keep pushing small changes to PRs and using up all the mac capacity.

Cribbed from:
https://docs.github.com/en/actions/using-jobs/using-concurrency


thiagohmcruz (approved):
Nice

justinseanmartin (approved):

PR #607, opened 2022-11-18
Adding plist to the bundled outputs on static frameworks

mccorkill1:
For static `apple_framework` targets, if the `infoplists` attribute is included, the plist isn't included in the .framework output by the rule. This PR makes it so that if the `infoplists` attribute is included, the computed plist will be included in the output. If the attribute is not included, the plist will not be included in the output which is the same as the behavior now.

This PR also adds `plist` to the acceptable parameters of the `skip_packaging` attribute in `apple_framework` targets


rules/framework.bzl
```diff
@@ -155,12 +155,13 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
...
 
+    if action in ctx.attr.skip_packaging:
+        return []
+
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
     if virtualize_frameworks:
         return inputs
 
-    if action in ctx.attr.skip_packaging:
-        return []
```
mattrobmattrob, line 163: What's the explanation for this change?

rules/framework/framework_packaging.py
```diff
@@ -108,6 +108,8 @@ def main():
             lambda args: _merge_binaries(args.framework_root, args.framework_name, args.inputs),
         "modulemap":
             lambda args: _copy_modulemap(args.framework_root, args.input()),
+        "plist":
+            lambda args: _cp(args.input(), args.output()),
```
mattrobmattrob, line 112: Since a framework bundle could contain other `*.plist` resources, can we name this explicitly `infoplist` action?

rules/framework.bzl
```diff
@@ -788,6 +789,13 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
 def _bundle_static_framework(ctx, is_extension_safe, outputs):
     """Returns bundle info for a static framework commonly used intra-build"""
     infoplist = _merge_root_infoplists(ctx)
+    plist_out = infoplist
+    framework_dir = "%s/%s.%s" % (ctx.attr.name, ctx.attr.framework_name, ctx.attr.bundle_extension)
+    plist_out = [paths.join(
+        framework_dir,
+        "Info.plist",
+    )]
+    plist_out = _framework_packaging(ctx, "plist", [infoplist], plist_out, None)
```
mattrobmattrob, line 798: Is it possible (or does it make sense) to move this within `_get_framework_files` like the rest of the "packaging" related things utilizing `_framework_packaging`?

rules/framework.bzl
```diff
@@ -896,12 +904,14 @@ def _apple_framework_packaging_impl(ctx):
...
 
     # If we link dynamic - then package it as dynamic
+    plist_out = None
     if ctx.attr.link_dynamic:
         bundle_outs = _bundle_dynamic_framework(ctx, is_extension_safe = is_extension_safe, avoid_deps = avoid_deps)
         avoid_deps_info = AvoidDepsInfo(libraries = depset(avoid_deps + ctx.attr.deps).to_list(), link_dynamic = True)
     else:
         bundle_outs = _bundle_static_framework(ctx, is_extension_safe = is_extension_safe, outputs = outputs)
         avoid_deps_info = AvoidDepsInfo(libraries = depset(avoid_deps).to_list(), link_dynamic = False)
+        plist_out = bundle_outs.providers[0].infoplist
```
mattrobmattrob, line 914: I don't think we should commit this since it's assuming the order of the returned `providers` stays the same. That seems bad to me. Move to `_get_framework_files` might mitigate this assumption here by relying on more general infrastructure.

mattrobmattrob (changes requested):
Blocking because of the `bundle_outs.providers[0].infoplist` assumption. Couple of comments as well.

rules/framework.bzl
jerrymarino: ➕ yeah agree - that was my main question about the PR too - Is it possible to align with the others? I was really hoping to keep 2 distinct branches in this if statement. If it's not possible maybe we can try to improve the code and re-work it as a followup.

tests/ios/frameworks/strict-deps/a/BUILD.bazel
```diff
@@ -1,10 +1,29 @@
...
     name = "a",
     srcs = glob(["*.swift"]),
     infoplists = ["//rules/test_host_app:Info.plist"],
     platforms = {"ios": "10.0"},
+    skip_packaging = ["plist"],
     visibility = ["//visibility:public"],
     deps = ["//tests/ios/frameworks/strict-deps/b"],
 )
+
+apple_framework(
```
jerrymarino, line 14: Thanks for adding a test on this - this was totally off peoples radar I think ❤️

rules/framework.bzl
jerrymarino: Agree - the change @mattrobmattrob mentioned might mitigate this change too:
https://github.com/bazel-ios/rules_ios/pull/607/files#diff-b5ab983bf150dfa86b7e361224e1c9b0b9856b0b57b5893dbc088c6ca6d8ba52L162

Concretely when `apple.virtualize_frameworks` is set - the rules shouldn't generate any actions for the frameworks like `modulemap` or `plist`

jerrymarino (commented):
Thanks for the PR @mccorkill1 ! Minor inline suggestion and we should land it otherwise. 

I was curious if you were distributing the frameworks outside of Bazel with this?

rules/framework.bzl
mccorkill1: I moved this to the `_get_framework_files ` but I also had to move the plist generation out of `_bundle_static_framework` so that it could be packaged inside this method. The generated plist is then passed to the `_bundle_static_framework` through `outputs` so that it can be included in the `AppleBundleInfo` provider.

mccorkill1:
@jerrymarino Because we don't have a build file for the full application target yet and are still working on the dependencies, we are using bazel to build the frameworks we depend on and pointing xcode to use those.

rules/framework.bzl
```diff
@@ -363,6 +366,13 @@ def _get_framework_files(ctx, deps):
     virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
     if not virtualize_frameworks:
         framework_manifest = ctx.actions.declare_file(framework_dir + ".manifest")
+        if ctx.attr.link_dynamic != True:
+            infoplist_in = _merge_root_infoplists(ctx)
+            infoplist_out = infoplist_in
+            infoplist_out = [paths.join(
+                framework_dir,
+                "Info.plist",
+            )]
```
mattrobmattrob, line 375: I wish we didn't have to bring `ctx.attr.link_dynamic` in here but it seems necessary to maintain the functionality from before. Little simplification of this code though before merge, please:
```suggestion
        if not ctx.attr.link_dynamic:
            infoplist_in = _merge_root_infoplists(ctx)
            infoplist_out = [paths.join(
                framework_dir,
                "Info.plist",
            )]
```

rules/framework.bzl
```diff
@@ -161,6 +161,7 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
 
     if action in ctx.attr.skip_packaging:
         return []
+
```
mattrobmattrob, line 164: [nit]
```suggestion
        return []
```

mattrobmattrob (approved):
:shipit: Thanks for the addition and refactor, @mccorkill1! I'd like @jerrymarino to also give an approval, if you don't mind.

mccorkill1:
@mattrobmattrob any idea what's going on with the LLDB tests? They pass for me locally.

mattrobmattrob:
> @mattrobmattrob any idea what's going on with the LLDB tests? They pass for me locally.

Looks like it passed when I re-ran it. 👍

jerrymarino:
Thanks for the PR @mccorkill1 !

PR #609, opened 2022-11-18
Fix duplicate testonly kwargs in test.bzl

luispadron:
## Summary

If a user were to pass `testonly` to a `ios_unit_test` target then it fails with duplicate kwargs error due to `testonly = True` being hardcoded in the rules.

This updates the rules to default to `true` for `testonly` file removing the `testonly` arg from kwargs to resolve the duplicate kwargs error.

Related change: https://github.com/bazel-ios/cocoapods-bazel/pull/82


jerrymarino (approved):
@luispadron LGTM overall and something we'd want even if we work towards other improvements / back port injecting the missing `XCTest` paths automatically

PR #611, opened 2022-11-23
Have LIBTOOL use the ld stub as well

justinseanmartin:
It looks like in Xcode 14, it is now passing a `-dependency_info` arg to libtool. Without this change, Xcode 14 generates an error when trying to find the `*_libtool_dependency_info.dat ` output file for creating static library targets. I've confirmed that this change works for both Xcode 13 & Xcode 14.

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -1061,7 +1061,7 @@ def _xcodeproj_impl(ctx):
         "CXX": "$CC",
         "CLANG_ANALYZER_EXEC": "$CC",
         "LD": "$BAZEL_STUBS_DIR/ld-stub",
-        "LIBTOOL": "/usr/bin/true",
+        "LIBTOOL": "$BAZEL_STUBS_DIR/ld-stub",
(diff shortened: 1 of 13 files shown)


gyfelton (approved):
so ld_stub is really a stub .dat file writer... :TIL:

thiagohmcruz (approved):

justinseanmartin:
> For Xcode 14 - we should back-port the logic from XCHammer into rules_ios for now which we syncd abt offline 🚢

We already have [similar logic](https://github.com/bazel-ios/rules_ios/blob/master/tools/xcodeproj_shims/ld-stub.sh#L9) to [xchammer](https://github.com/pinterest/xchammer/blob/c588329904e05072c8d674191318e64d8dabc685/XCHammerAssets/xcode_ld_stub.sh) for `ld`, we just weren't applying it to `libtool` as XCHammer was. It seems they weren't passing the `-dependency_info` flag previously.

jerrymarino:
Cool - if you've got it working that is a net benefit overall. We will remove the legacy one - but don't want to pull the rug by doing so.

PR #613, opened 2022-11-24
Some minor fixes around private_deps

dnkoutso:
Expose `private_deps` attribute for aspects to be able to use
Do not use `append()` for `private_deps` in case `select` is being used.


jerrymarino (approved):
LGTM - Nice PR 👏

PR #617, opened 2022-11-29
Cleanup plist output formulation

jerrymarino:
Today it formulates too long of files names for me on macOS. The total path size today is:
`(BuildFilePathLength + 2 * RuleNameLength + SuffixLength)`

Given Bazel supports a single rule with a given name per build file, we can safely eliminate this extra layer of namespacing. After the fix, the total path size is:
`(BuildFilePathLength + RuleNameLength + SuffixLength)`

Additionally:
- further reduce generated label / output sizes
- underscore generated rule names with _
- remove issue with ".." name suffix. This also gets us better looking plists names: For instance, a Foo dogfood plist gets you _Foo.dogfood.0.info.plist

### DIFF
--- a/rules/plists.bzl
+++ b/rules/plists.bzl
@@ -37,7 +37,7 @@ def process_infoplists(name, infoplists, infoplists_by_build_setting, xcconfig, 
     # Substitute the default infoplists
     substituted_infoplists_by_build_setting["//conditions:default"] = [
         substituted_plist(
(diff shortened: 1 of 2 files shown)


thiagohmcruz (approved):

luispadron (approved):
Thanks!

jerrymarino:
You got it 🚀

PR #618, opened 2022-11-30
Fix infoplist exclusivity / config inhertiance

jerrymarino:
The original version required that infoplists_by_build_setting and xcconfig_by_build_setting are mutually exclusive which was problematic for people who had different condtions. To remidate this problem, let build configuration pick _one_: move xcconfig_by_build_setting into a select() statement as-is so it is selected by configuration

Finally remove the "select" statement inheritance. From an API perspectie a select isn't inherited so we will only inherit from `xcconfig`. This is a breaking change if you relied on the previous behavior ( which was only a couple days ago )

### DIFF
--- a/rules/plists.bzl
+++ b/rules/plists.bzl
@@ -28,38 +28,33 @@ def process_infoplists(name, infoplists, infoplists_by_build_setting, xcconfig, 
     # Final substituted infoplists_by_build_setting dict
     substituted_infoplists_by_build_setting = {}
 
-    # Default infoplists are any defined in `infoplists` OR in `infoplists_by_build_setting` with the key `//conditions:default`
-    default_infoplists = infoplists_by_build_setting.pop("//conditions:default", infoplists)
+    # Default infoplists are any defined in `infoplists`
+    default_infoplists = infoplists if infoplists else []
 
-    # Default xcconfig is the one defined in `xcconfig` OR in `xcconfig_by_build_setting` with the key `//conditions:default`
-    default_xcconfig = xcconfig_by_build_setting.pop("//conditions:default", xcconfig)
-
-    # Substitute the default infoplists
-    substituted_infoplists_by_build_setting["//conditions:default"] = [
-        substituted_plist(
-            name = "_%s.%s.info" % (name, idx),
-            plist = plist,
-            xcconfig = default_xcconfig,
-        )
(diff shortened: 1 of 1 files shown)


jerrymarino:
cc @luispadron this addresses a few followups from the recent PR. The one change we might need to fix for your use case is the `//conditions:default` inheritance

rules/plists.bzl
```diff
@@ -132,17 +127,10 @@ def _substitute_plist_vars(name, plist, xcconfig):
     Returns:
         The name of the substituted plist
     """
-
-    variables = {k: v for (k, v) in xcconfig.items() if types.is_string(v)}
-
-    if len(variables) == 0:
```
luispadron, line 138: I added this check because otherwise we generated substituted plists when its not required.

rules/plists.bzl
```diff
@@ -28,38 +28,33 @@ def process_infoplists(name, infoplists, infoplists_by_build_setting, xcconfig,
...
     # Collect a set of config settings to iterate over
-    config_setting_names = sets.make(infoplists_by_build_setting.keys() + xcconfig_by_build_setting.keys())
+    merged_xcconfigs = {}
+    for config_setting_name in xcconfig_by_build_setting.keys():
+        xccconfig_for_build_setting = xcconfig_by_build_setting.get(config_setting_name, {})
+        merged_xcconfigs[config_setting_name]  = merge_xcconfigs(default_xcconfig, xccconfig_for_build_setting)
+
+    # If someone hasn't specified "//conditions:default" - inject it there as upstream callers will `select` on this
+    if len(infoplists_by_build_setting.keys()) == 0:
+        infoplists_by_build_setting["//conditions:default"] = default_infoplists
```
luispadron, line 45: I guess the confusing part is this differs from how `xcconfig` and `xcconfig_by_build_setting` work. 

`xcconfig` gets merged into all the different ones but this seems to ignore `infoplists` entirely if `infoplists_by_build_setting` is set

luispadron: If we dont want users to use them both at the same time (cash doesn't use `infoplists_by_build_setting` at all) should we just fail instead?

rules/plists.bzl
luispadron: Oh i guess its no longer possible because of `select`

PR #620, opened 2022-12-03
[Bazel6] Handle removal of `ObjcProvider.direct_headers`

mattrobmattrob:
Handle https://github.com/bazelbuild/bazel/issues/14559 (https://github.com/bazelbuild/bazel/commit/8a2b711a2700740575904682066dbe1e5c9f6d02) to prepare for Bazel 6 but still work in Bazel 5.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -410,7 +410,7 @@ def _get_direct_public_headers(provider, dep):
             return []
         return dep[provider].compilation_context.direct_public_headers
     elif provider == apple_common.Objc:
-        return dep[provider].direct_headers
+        return getattr(dep[provider], "direct_headers", [])
     else:
         fail("Unknown provider " + provider + " only CcInfo and Objc supported")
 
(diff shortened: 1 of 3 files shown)


rules/framework.bzl
```diff
@@ -410,7 +410,7 @@ def _get_direct_public_headers(provider, dep):
             return []
         return dep[provider].compilation_context.direct_public_headers
     elif provider == apple_common.Objc:
-        return dep[provider].direct_headers
+        return getattr(dep[provider], "direct_headers", [])
```
jerrymarino, line 413: @mattrobmattrob so the gist is that the previous condition is hit on Bazel 6. LGTM overall

jerrymarino (approved):
Thanks for tackling this in a way that works with the LTS release ❤️

rules/framework.bzl
mattrobmattrob: I didn't confirm that but as far as I can tell, yes. At least for all the cases tested by my code and the CI suite.

luispadron (approved):

PR #624, opened 2022-12-05
[Bazel6] Avoid copying in `.h` directories into `Foo.framework/Headers`

mattrobmattrob:
Fix the following mismatch between the declared output and the `symlink`’d input:
```
$ bazelisk test --local_test_jobs=1 --apple_platform_type=ios --deleted_packages='' -- //tests/ios/...
…
ERROR: /Users/matt.robinson/Developer/rules_ios/rules_ios/tests/ios/frameworks/intentdefinition/BUILD.bazel:3:16: in apple_framework_packaging rule //tests/ios/frameworks/intentdefinition:ObjCIntentConsumer:
	File "/Users/matt.robinson/Developer/rules_ios/rules_ios/rules/framework.bzl", line 854, column 43, in _apple_framework_packaging_impl
		framework_files = _get_framework_files(ctx, deps)
	File "/Users/matt.robinson/Developer/rules_ios/rules_ios/rules/framework.bzl", line 383, column 38, in _get_framework_files
		header_out = _framework_packaging(ctx, "header", header_in, header_out, framework_manifest)
	File "/Users/matt.robinson/Developer/rules_ios/rules_ios/rules/framework.bzl", line 185, column 45, in _framework_packaging
		_framework_packaging_symlink_headers(ctx, inputs, outputs)
	File "/Users/matt.robinson/Developer/rules_ios/rules_ios/rules/framework.bzl", line 150, column 28, in _framework_packaging_symlink_headers
		ctx.actions.symlink(output = output, target_file = input)
Error in symlink: symlink() with "target_file" directory param requires that "output" be declared as a directory (did you mean to use declare_directory() instead of declare_file()?)
ERROR: /Users/matt.robinson/Developer/rules_ios/rules_ios/tests/ios/frameworks/intentdefinition/BUILD.bazel:3:16: Analysis of target '//tests/ios/frameworks/intentdefinition:ObjCIntentConsumer' failed
```
If we `declare_directory`/`declare_file` correctly for the symlink'd outputs then we still run into an issue where there's a conflict on what's seemingly an incremental build:
```
ERROR: /Users/matt.robinson/Developer/rules_ios/rules_ios/tests/ios/frameworks/intentdefinition/BUILD.bazel:3:16: Creating symlink bazel-out/applebin_ios-ios_sim_arm64-dbg-ST-8e9d4392bc2b/bin/tests/ios/frameworks/intentdefinition/ObjCIntentConsumer/ObjCIntentConsumer.framework/Headers/ObjCIntentConsumer_Intents.intentdefinition_gen.hdrs.h failed: failed to create symbolic link 'tests/ios/frameworks/intentdefinition/ObjCIntentConsumer/ObjCIntentConsumer.framework/Headers/ObjCIntentConsumer_Intents.intentdefinition_gen.hdrs.h' to 'tests/ios/frameworks/intentdefinition/ObjCIntentConsumer_Intents.intentdefinition_gen.hdrs.h' due to I/O error: /private/var/tmp/_bazel_matt.robinson/0e64031a9efbb95e4728094ee9457143/execroot/build_bazel_rules_ios/bazel-out/applebin_ios-ios_sim_arm64-dbg-ST-8e9d4392bc2b/bin/tests/ios/frameworks/intentdefinition/ObjCIntentConsumer/ObjCIntentConsumer.framework/Headers/ObjCIntentConsumer_Intents.intentdefinition_gen.hdrs.h (File exists)
```
It makes little sense to add `.h` “files” to `Foo.framework/Headers` that are actually directories so stop adding them to the resulting frameworks `Headers` directory unless they're files.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -162,15 +162,7 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
     if action in ctx.attr.skip_packaging:
         return []
     action_inputs = [manifest] + inputs if manifest else inputs
-
-    # Use the input to determine whether to use `declare_directory` or `declare_file`
-    # since it'll need to match in the downstream symlink action.
-    outputs_by_index = {index: outputs[index] for index in range(len(outputs))}
-    outputs = [
-        ctx.actions.declare_directory(output) if inputs[index].is_directory else ctx.actions.declare_file(output)
-        for (index, output) in outputs_by_index.items()
-    ]
-
+    outputs = [ctx.actions.declare_file(f) for f in outputs]
     framework_name = ctx.attr.framework_name
     framework_dir = _find_framework_dir(outputs)
     args = ctx.actions.args().use_param_file("@%s").set_param_file_format("multiline")
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -331,7 +331,7 @@ def _get_framework_files(ctx, deps):
         for provider in [CcInfo, apple_common.Objc]:
             if provider in dep:
                 for hdr in _get_direct_public_headers(provider, dep):
-                    if hdr.path.endswith((".h", ".hh", ".hpp")):
+                    if not hdr.is_directory and hdr.path.endswith((".h", ".hh", ".hpp")):
                         has_header = True
                         header_in.append(hdr)
                         destination = paths.join(framework_dir, "Headers", hdr.basename)
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -162,7 +162,15 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
     if action in ctx.attr.skip_packaging:
         return []
     action_inputs = [manifest] + inputs if manifest else inputs
-    outputs = [ctx.actions.declare_file(f) for f in outputs]
+
+    # Use the input to determine whether to use `declare_directory` or `declare_file`
+    # since it'll need to match in the downstream symlink action.
+    outputs_by_index = {index: outputs[index] for index in range(len(outputs))}
+    outputs = [
+        ctx.actions.declare_directory(output) if inputs[index].is_directory else ctx.actions.declare_file(output)
+        for (index, output) in outputs_by_index.items()
+    ]
+
     framework_name = ctx.attr.framework_name
     framework_dir = _find_framework_dir(outputs)
     args = ctx.actions.args().use_param_file("@%s").set_param_file_format("multiline")


rules/framework.bzl
```diff
@@ -162,7 +162,15 @@ def _framework_packaging(ctx, action, inputs, outputs, manifest = None):
     if action in ctx.attr.skip_packaging:
         return []
     action_inputs = [manifest] + inputs if manifest else inputs
-    outputs = [ctx.actions.declare_file(f) for f in outputs]
```
jerrymarino, line 165: Minor: To avoid 1 iteration - can you do something like:
```
    outputs = [ctx.actions.declare_directory(f) if f.is_directory else ctx.actions.declare_file(f) for f in outputs]
```

jerrymarino (approved):

rules/framework.bzl
mattrobmattrob: Happy to collapse into a single loop. Was having some issues with this differing between Bazel 5 and Bazel 6 so seeing if it passes on CI then will update. 👍

mattrobmattrob: And in this particular case, the `symlink` action is complaining about the input `File` from `inputs` not matching the declared so it'll need to be a loop over indexes anyway. Something like:
```Python
    outputs = [
        ctx.actions.declare_directory(outputs[index]) if inputs[index].is_directory else ctx.actions.declare_file(outputs[index])
        for index in range(len(outputs))
    ]
```
Let me know if this isn't okay with you, @jerrymarino.

jerrymarino: Cool, makes sense! Yeah perhaps either Bazel 6 or rule bumps ended up getting us into a situation that converted something that was a file into dir?

mattrobmattrob: The case the `rules_ios` test suite catches is the intent definition's declaration of `.h`/`.m` directories:

https://github.com/bazelbuild/rules_apple/blob/c13b1072832737c7ca7d24ec598a53c590e4e010/apple/internal/resource_rules/apple_intent_library.bzl#L55-L56

mattrobmattrob:
Re-requesting your review, @jerrymarino, since the change is now quite different from the initial implementation.

rules/framework.bzl
```diff
@@ -323,7 +323,7 @@ def _get_framework_files(ctx, deps):
         for provider in [CcInfo, apple_common.Objc]:
             if provider in dep:
                 for hdr in _get_direct_public_headers(provider, dep):
-                    if hdr.path.endswith((".h", ".hh", ".hpp")):
+                    if not hdr.is_directory and hdr.path.endswith((".h", ".hh", ".hpp")):
```
jerrymarino, line 326: This LGTM - I think the upstream code would break if it was a dir anyhow

jerrymarino (approved):

PR #626, opened 2022-12-08
Align swiftc flags between Bazel and Xcode in XCHammer

thiagohmcruz:
Requires: 

* https://github.com/bazel-ios/xchammer/pull/37
* https://github.com/jerrymarino/xcbuildkit/pull/54

Also, changes the global index store path for ObjC to match Swift's. Open to suggestions here if anyone sees this as a potential breaking change, alternative is to conditionally set this when using XCHammer and let the current value be the default.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -21,7 +21,7 @@ PrivateHeadersInfo = provider(
     },
 )
 
-GLOBAL_INDEX_STORE_PATH = "bazel-out/rules_ios_global_index_store.indexstore"
+GLOBAL_INDEX_STORE_PATH = "bazel-out/_global_index_store"
 
 _MANUAL = ["manual"]
 
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -136,15 +136,15 @@ swift_binary(
             name = "xchammer",
             remote = "https://github.com/bazel-ios/xchammer.git",
             # XCHammer dev branch: bazel-ios/rules-ios-xchammer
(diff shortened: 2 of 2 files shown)


rules/library.bzl
```diff
@@ -21,7 +21,7 @@ PrivateHeadersInfo = provider(
     },
 )
 
-GLOBAL_INDEX_STORE_PATH = "bazel-out/rules_ios_global_index_store.indexstore"
+GLOBAL_INDEX_STORE_PATH = "bazel-out/_global_index_store"
```
thiagohmcruz, line 24: Curious to know if anyone sees an issue with changing this?

jerrymarino (approved):

rules/library.bzl
thiagohmcruz: cc @jerrymarino do you see any issues with this?

I think I'll come up with a good way to test it and verify if it impacts existing users

jerrymarino: @thiagohmcruz no major issues AFAIK - I originally suggested this change the other day in our meeting and think it is the right way.

congt (approved):

thiagohmcruz:
Our internal CI is 🟢 with this, also ran some tests simulating someone using the legacy generator and moving from a previous SHA to this one and didn't hit any issues.

For other contributors, if you hit any issue due to the fact that the path to the global index store has changed let me know and I can revert / help troubleshoot.

PR #627, opened 2022-12-09
Pass product module name into test_rule

dostrander:
This will need to wait until [this](https://github.com/bazelbuild/rules_apple/pull/1785) is merged

### DIFF
--- a/patches/rules_apple_module_name.patch
+++ b/patches/rules_apple_module_name.patch
@@ -1,0 +1,84 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Derek Ostrander <derko@squareup.com>
+Date: Fri, 9 Dec 2022 13:55:16 -0500
+Subject: Add product module name as an optional arg for the test rule
+
+If there are modules under the tests that are named differently then the target itself it will cause issues with the xctestrun file. Specifically using Swift the test_filters do not work unless the module names line up with the test target name. Guessing because Swift's full class names are `MODULE.CLASS` so they do the magic under the hood.
+
+diff --git a/apple/internal/rule_factory.bzl b/apple/internal/rule_factory.bzl
+index 33d4312f..b196a577 100644
+--- a/apple/internal/rule_factory.bzl
++++ b/apple/internal/rule_factory.bzl
+@@ -205,6 +205,12 @@ AppleTestRunnerInfo provider.
+     "test_filter": attr.string(
+         doc = """
+ Test filter string that will be passed into the test runner to select which tests will run.
++""",
++        default = "",
++    ),
++    "module_name": attr.string(
++        doc = """
++Module name of the underlying test to be run. 
+ """,
+         default = "",
+     ),
+diff --git a/apple/internal/testing/apple_test_rule_support.bzl b/apple/internal/testing/apple_test_rule_support.bzl
+index 8392862a..49ff1f65 100644
+--- a/apple/internal/testing/apple_test_rule_support.bzl
++++ b/apple/internal/testing/apple_test_rule_support.bzl
+@@ -102,7 +102,7 @@ This aspect propagates a `CoverageFilesInfo` provider.
+     implementation = _coverage_files_aspect_impl,
+ )
+ 
+-def _get_template_substitutions(test_type, test_bundle, test_environment, test_host = None, test_filter = None):
++def _get_template_substitutions(test_type, test_bundle, test_environment, test_host = None, test_filter = None, product_module_name = None):
+     """Dictionary with the substitutions to be applied to the template script."""
+     subs = {}
+ 
+@@ -114,6 +114,7 @@ def _get_template_substitutions(test_type, test_bundle, test_environment, test_h
+     subs["test_type"] = test_type.upper()
+     subs["test_env"] = ",".join([k + "=" + v for (k, v) in test_environment.items()])
+     subs["test_filter"] = test_filter or ""
++    subs["product_module_name"] = product_module_name or ""
+ 
+     return {"%(" + k + ")s": subs[k] for k in subs}
+ 
+@@ -176,6 +177,7 @@ def _apple_test_rule_impl(ctx, test_type):
+             test_environment,
+             test_host = test_host_archive,
+             test_filter = ctx.attr.test_filter,
++            product_module_name = ctx.attr.module_name,
+         ),
+         is_executable = True,
+     )
+diff --git a/apple/testing/default_runner/ios_test_runner.template.sh b/apple/testing/default_runner/ios_test_runner.template.sh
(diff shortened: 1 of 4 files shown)


rules/test.bzl
```diff
@@ -105,6 +105,7 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
             tests.append(test_name)
             split_rule(
                 name = test_name,
+                product_module_name = name.replace("-", "_"),
```
jerrymarino, line 108: Should this use `kwargs.module_name`?

jerrymarino (approved):

rules/test.bzl
dostrander: Yes! I thought this was a thing but I couldn't find it initially

mattrobmattrob:
Have we solved additions to `rules_apple` like this in a backwards compatible way in the future or just assume you'll need to bump to be on this change or newer?

rules/test.bzl
```diff
@@ -105,6 +105,7 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
             tests.append(test_name)
             split_rule(
                 name = test_name,
+                module_name = kwargs["module_name"],
```
jszumski, line 108: Would adding `"module_name"` to `_IOS_TEST_KWARGS` accomplish the same thing while keeping all the important keys in one place?

dostrander: Yeah it should. Tried it at first but it failed for a way i didn't understand and din't have to time to debug. I actually didn't mean to push this up but was int he wrong repo :). I need to debug why it wasn't working and hopefully get it to work

dostrander:
> Have we solved additions to `rules_apple` like this in a backwards compatible way in the future or just assume you'll need to bump to be on this change or newer?

@mattrobmattrob I was thinking of looking into some patching for this PR as well but i'm not sure how future-proof that is

rules/test.bzl
dostrander: So this wasn't able to be done because the kwargs["module_name"] was also used in `apple_library` our current implementation of `_IOS_TEST_KWARGS` pops them off so they aren't passed into apple_library. I added a comment explaining in code so it's not as confusing int he future

dostrander:
@mattrobmattrob @jerrymarino PTAL at how I plan on doing patching. LMK if there is something else we may want to do

patches/rules_apple_module_name.patch
```diff
@@ -0,0 +1,84 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Derek Ostrander <derko@squareup.com>
+Date: Fri, 9 Dec 2022 13:55:16 -0500
+Subject: Add product module name as an optional arg for the test rule
```
jerrymarino, line 4: What's the longer term solution for this one? It'll likely be painful to maintain this longer term. Someone should either:
1. propose upstream fixes to it
2. copy the rules_apple code we need for these rules into rules_ios

patches/rules_apple_module_name.patch
```diff
@@ -0,0 +1,84 @@
...
+diff --git a/apple/testing/default_runner/ios_test_runner.template.sh b/apple/testing/default_runner/ios_test_runner.template.sh
+index 0804007c..0fbf190f 100644
+--- a/apple/testing/default_runner/ios_test_runner.template.sh
++++ b/apple/testing/default_runner/ios_test_runner.template.sh
+@@ -52,6 +52,11 @@ TMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/test_runner_work_dir.XXXXXX")"
+ trap 'rm -rf "${TMP_DIR}"' ERR EXIT
+ runner_flags+=("--work_dir=${TMP_DIR}")
+ 
++PRODUCT_MODULE_NAME="%(product_module_name)s"
++if [[ -n "$PRODUCT_MODULE_NAME" ]]; then
```
jerrymarino, line 63: This is necessary for filtering anyhow right? I think it's reasonable to fix that on rules_apple

patches/xctestrunner_module_name.patch
```diff
@@ -0,0 +1,147 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
jerrymarino, line 1: For xctestrunner we can ping the core maintainer with the patch - did you make a PR for that? If they are MIA we can push to a rules_ios fork if it comes to that.

jerrymarino (approved):
Filtering looks fine - my main comment is on maintaining patch files for rules_apple bumps.. Do you have PRs linked for that? We can take the other approach for a subset of rules_apple if they aren't upstreamable.

jerrymarino:
Also are there concerns the rules_apple patch will break for people even with landing this? I gave it a stamp assuming we'll fix it - but it might cause an issue.

jerrymarino:
>  Have we solved additions to rules_apple like this in a backwards compatible way in the future or just assume you'll need to bump to be on this change or newer?

➕ to @mattrobmattrob comment on this

rules/test.bzl
```diff
@@ -124,6 +129,7 @@ def _ios_test(name, test_rule, test_suite_rule, apple_library, infoplists_by_bui
 
         rule(
             name = name,
+            module_name = module_name,
```
jerrymarino, line 132: I think this will break other people that have a different version of rules_apple ( similar of @mattrobmattrob's comment )

jerrymarino (changes requested):

jerrymarino:
We already keep a fork of`xctestrunner` for our patches from time to time. To better integrate with how we do things, and to integrate with github workflows I'd like to use that instead of patch files:  https://github.com/bazel-ios/xctestrunner - similar for rules_apple as well.

This said - we'll still need to make it backwards compatible.

mattrobmattrob:
FWIW, I don't want to block a merge because of backwards compatibility, necessarily. Just want it to be a conversation. 👍 We are essentially on trunk so it shouldn't cause any issues for us.

jerrymarino:
The "backwards compatibility" issue I encountered was the PR seemed to require me to use the rules_apple version generated by the patch: in other words, if it needs argument `module_name` which doesn't exist outside of the patches. So if you have a different version of rules_apple than rules_ios gives you it won't work.

jerrymarino:
@dostrander if you want to add an if statement in a macro - I came up with a mechansim to add this:
https://github.com/bazel-ios/rules_ios/pull/628

```
load("@rules_ios_cfg//:config.bzl", "using_internal_rules_apple")

if using_internal_rules_apple:
   # some
else:
   # other
```

We might want this anyhow - but let me know what you're thoughts are. I added some suggestions to simplify rules_apple in that review BTW - both which can be backwards compatible.

PR #628, opened 2022-12-14
Make it possible to detect trunk deps in macros

jerrymarino:
In the loading phase we want to switch on things you wouldn't detect

This provides a mechanism you can feed to your macros:

```
load(@rules_ios_cfg, optimized_rules_swift)
```
And then do various stuff in the loading phase

### DIFF
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -51,14 +51,15 @@ def github_repo(name, project, repo, ref, sha256 = None, **kwargs):
 
 
 def _rules_ios_cfg_impl(ctx):
-    root_build_file = """
+    root_cfg_file = """
        optimized_rules_apple={optimized_rules_apple}
        optimized_rules_swift={optimized_rules_swift}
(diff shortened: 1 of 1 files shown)


rules/repositories.bzl
```diff
@@ -68,6 +97,12 @@ def rules_ios_dependencies():
         sha256 = "5e82a98a591efda772a5ee96ed17bcad38338aafeba6055daab04a5d6c13ea50",
     )
 
+    rules_ios_cfg(
```
jerrymarino, line 100: If we want to pursue @dostrander 's loading phase `kwargs` injection we'll need something like this. The PR is in a WIP / draft state so we can move forward on that if needed.

jerrymarino: Open to a better suggestion to switch in the loading phase if someone has it. I don't know of a great way you can introspect a `StarlarkCallable` off the top of my head and this might be better anyways.

PR #629, opened 2022-12-15
Fix analysis failure for targets that define `features`

jszumski:
Fix issue where targets that define `features` would not get merged with the internal `swift_library`'s `features` causing an analysis failure:

```
File "/private/var/tmp/_bazel_jszumski/ed7e7efa198ffcfaf8285220a74bc6f4/external/build_bazel_rules_ios/rules/test.bzl", line 142, column 14, in ios_unit_test
	_ios_test(name, rules_apple_ios_unit_test, rules_apple_ios_unit_test_suite, apple_library, **kwargs)
File "/private/var/tmp/_bazel_jszumski/ed7e7efa198ffcfaf8285220a74bc6f4/external/build_bazel_rules_ios/rules/test.bzl", line 62, column 28, in _ios_test
	library = apple_library(name = name, namespace_is_module_name = False, platforms = {"ios": ios_test_kwargs.get("minimum_os_version")}, testonly = testonly, **kwargs)
File "/Users/jszumski/Development/cash-ios/Tools/Rules/library.bzl", line 46, column 25, in cash_library
	return apple_library(
File "/private/var/tmp/_bazel_jszumski/ed7e7efa198ffcfaf8285220a74bc6f4/external/build_bazel_rules_ios/rules/library.bzl", line 964, column 22, in apple_library
	swift_library(
Error in swift_library: rule(...) got multiple values for parameter 'features'
```

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -568,6 +568,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     enable_framework_vfs = kwargs.pop("enable_framework_vfs", False) or namespace_is_module_name
     defines = kwargs.pop("defines", [])
     testonly = kwargs.pop("testonly", False)
+    features = kwargs.pop("features", [])
 
     for (k, v) in {"momc_copts": momc_copts, "mapc_copts": mapc_copts, "ibtool_copts": ibtool_copts}.items():
         if v:
@@ -978,7 +979,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
                 "//conditions:default": [framework_vfs_overlay_name_swift] if enable_framework_vfs else [],
             }),
             swiftc_inputs = swiftc_inputs,
-            features = ["swift.no_generated_module_map", "swift.use_pch_output_dir"] + select({
+            features = features + ["swift.no_generated_module_map", "swift.use_pch_output_dir"] + select({
                 "@build_bazel_rules_ios//:virtualize_frameworks": ["swift.vfsoverlay"],
                 "//conditions:default": [],
             }),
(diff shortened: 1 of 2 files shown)


luispadron (approved):

rules/library.bzl
```diff
@@ -961,6 +961,8 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
     )
 
     if swift_sources:
+        features = kwargs.pop("features", [])
```
mattrobmattrob, line 964: Should we pull this out at a higher level and forward to "all" places (maybe just those getting `**kwargs` for now)? Since `objc_library` below was getting and could/should also get these `features`.

tests/ios/unit-test/BUILD.bazel
```diff
@@ -70,3 +70,13 @@ ios_unit_test(
...
     minimum_os_version = "11.0",
 )
+
+ios_unit_test(
+    name = "ExplicitFeatures",
+    srcs = [
+        "empty.m",
+        "empty.swift",
+    ],
+    features = ["apple.skip_codesign_simulator_bundles"],
```
mattrobmattrob, line 80: Is this a performance optimization or just a test?

rules/library.bzl
jszumski: I think the actual Obj-C one (L1067) is getting it via `**kwargs` but you make a good point about the C++ one (L1024).

tests/ios/unit-test/BUILD.bazel
jszumski: Just a test.  This particular value just happened to be the one that led me to find the issue.

rules/library.bzl
mattrobmattrob: Not if you pop it here, right?

jszumski: Fixed!

mattrobmattrob (approved):
🎉

PR #635, opened 2023-01-09
Pass the ios args to the test suite

dostrander:
Without this different attributes like "visibility" are not able to be set on the test suites. This inherits the ios args that are passed into the test rule to also be passed into the test suite on split ios tests


jerrymarino (approved):
Thanks for the fix @dostrander 🚢

Issue #637, opened 2023-01-09
[xcodeproj] App extensions on `test_host` not included as targets


mattrobmattrob:
An `xcodeproj` target with a dep to an `ios_unit_test` with a `test_host` that has extensions end up creating schemes for the extensions but no targets. This means the project generations fails because the schemes references nonexistent targets.

Likely "caused by" https://github.com/bazel-ios/rules_ios/pull/486 which started adding these schemes.

# Repro
Change to reproduce:
```diff
diff --git a/tests/ios/unit-test/BUILD.bazel b/tests/ios/unit-test/BUILD.bazel
index ae75672..dac5b10 100644
--- a/tests/ios/unit-test/BUILD.bazel
+++ b/tests/ios/unit-test/BUILD.bazel
@@ -29,7 +29,7 @@ ios_unit_test(
         "empty.swift",
     ],
     minimum_os_version = "12.0",
-    test_host = "//rules/test_host_app:iOS-9.3-AppHost",
+    test_host = "//tests/ios/app:App",
     # Adding visibility so the xcodeproject tests can depend on this target
     visibility = ["//visibility:public"],
 )
```

Testing:
```
$ bazelisk run //tests/ios/xcodeproj:Test-With-Host-App
...
ERROR: /Users/matt.robinson/Developer/rules_ios/rules_ios/tests/ios/xcodeproj/BUILD.bazel:103:10: Action tests/ios/xcodeproj/TestWithHostApp.xcodeproj failed: (Exit 1): xcodegen failed: error executing command 
  (cd /private/var/tmp/_bazel_matt.robinson/0e64031a9efbb95e4728094ee9457143/sandbox/darwin-sandbox/14/execroot/build_bazel_rules_ios && \
  exec env - \
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/com_github_yonaskolb_xcodegen/xcodegen --quiet --no-env --spec bazel-out/darwin_arm64-dbg/bin/tests/ios/xcodeproj/Test-With-Host-App-xcodegen.json --project bazel-out/darwin_arm64-dbg/bin/tests/ios/xcodeproj)
# Configuration: bcb4a48eff0aba08e8b828fb48d94a0d9d929518f05634ebb9dd6f75e31cef88
# Execution platform: @local_config_platform//:host

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
[...]

PR #638, opened 2023-01-10
[xcodeproj] Ensure extension targets are created when in schemes

mattrobmattrob:
Ensure that when extension schemes are created that they have corresponding targets as well.

Fixes https://github.com/bazel-ios/rules_ios/issues/637.

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -233,6 +233,8 @@ def _xcodeproj_aspect_impl(target, ctx):
     bazel_bin_subdir = "%s/%s" % (target.label.workspace_root, target.label.package)
 
     if AppleBundleInfo in target:
+        extensions = getattr(ctx.rule.attr, "extensions", [])
+
         swift_objc_header_path = None
         if SwiftInfo in target:
             for h in getattr(target[apple_common.Objc], "direct_headers", []):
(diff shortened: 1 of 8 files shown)


tests/ios/unit-test/BUILD.bazel
```diff
@@ -29,7 +29,8 @@ ios_unit_test(
         "empty.swift",
     ],
     minimum_os_version = "12.0",
-    test_host = "//rules/test_host_app:iOS-9.3-AppHost",
```
jerrymarino, line 32: Do we have another check with this versioned app host - otherwise it looks to drop coverage and would suggest to spin up another `ios_unit_test`

rules/legacy_xcodeproj.bzl
```diff
@@ -315,8 +311,16 @@ def _xcodeproj_aspect_impl(target, ctx):
...
-            # Add the test_host_target to the targets to be added to the xcode project
-            test_host_direct_targets = test_host_target[_TargetInfo].direct_targets
+            test_host_direct_targets = []
+
+            for target in test_host_target[_TargetInfo].direct_targets:
+                # Add the test_host_target to the targets to be added to the xcode project
+                test_host_direct_targets.extend([target])
+
+                # Add the test_host_target's extensions to the targets to be added to the xcode project
+                for extension in target.extensions.to_list():
```
jerrymarino, line 321: Does this aggregate a transitive extension for a test bundle which depends on a hosted app?

jerrymarino (approved):
This mostly seems fine, QQ inline around a few cases

jerrymarino:
Thanks for the PR @mattrobmattrob 🥳

tests/ios/unit-test/BUILD.bazel
mattrobmattrob: Will add another test case. 👍

rules/legacy_xcodeproj.bzl
mattrobmattrob: Yes. Previously, if I had `//tests/ios/unit-test:ExplicitHostedWithAppExtensions`, it wouldn't end up building in Xcode since the extensions for the test host would be missing.

PR #639, opened 2023-01-10
Correct `AppleBundleInfo.infoplist` value

mattrobmattrob:
Ensure the `AppleBundleInfo.infoplist` value is a single `File` instead of a list like `rules_apple` wants:

https://github.com/bazelbuild/rules_apple/blob/4f2b1f840db65ba567f86e1e4b51d19612db2f57/apple/providers.bzl#L72-L74

This was introduced in https://github.com/bazel-ios/rules_ios/pull/607.


jerrymarino (approved):
LGTM - thanks for fixing this 👍 !

PR #640, opened 2023-01-10
Clean up `_get_framework_files` naming

mattrobmattrob:
In an effort to make it easier to understand, make the naming clear when an output is a single item or `None` OR a list or `None`. This makes the downstream handling of these `struct` values easier to comprehend for changes to `framework.bzl`.


luispadron (approved):
Looks clearer 🚀

Issue #542, opened 2022-08-16
Generalized strategies to mitigate NDeps * MApps * KTests disk usage in bundling mechanism


jerrymarino:
In a big mono-repo I'd like to build the tree to vet refactoring. Say I've got an order of 1k deps, 1k apps, and 1k tests - I'll have significant disk growth. In rules_ios consider a mechanism in both the simulator application and simulator tests builds to normalize the data. Of course, please make the changes work with remote execution and buildfarm

For instance:
### Normalizing resource bundles in apps/tests to symlinked runfiles
Move this to a symlinked runfile to ensure it works with remote execution
```
Bogs/Some.bundle -> bazel-out/MyTest.xctest/some.bundle 
```

### Normalizing dynamic deps to in apps/tests to symlinked runfiles
```
Bogs/Some.framework -> bazel-out/MyTest.xctest/Frameworks/Bogs/Some.framework
```

On further thought - we might consider but could have issues with be able to use apfs copy on write - in the context of remote execution.

PR #648, opened 2023-01-25
[RFC] Basic aspect to  _just compile_ a subset

jerrymarino:
In a big repo I'd like to compile the tree to vet refactoring. However, in the face of #542 we need more changes to fix disk growth. That aside, I generally don't want need to link, codesign, or bundle, so this gets an elemental subset of compiling only with a quick and dirty aspect.

It's useful for CLI development only and wouldn't be part of the core rules.

@congt  I think we can use this to some sort of "type checking" the build as well: just asserting it will compile for something like a merge queue

### DIFF
--- a/tools/compile_only_aspect.bzl
+++ b/tools/compile_only_aspect.bzl
@@ -1,0 +1,35 @@
+"""
+Helper aspect to do objc and swift compilation only for fast iteration cycles on
+big mono-repo refactors.
+
+In a mono-repo a developer will change much objc and swift source code. We don't
+want to link or codesign the intermediates. This program collects the relevate
+compile actions 
+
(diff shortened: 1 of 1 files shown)


congt (approved):

thiagohmcruz (approved):

congt:
I got this error: 
```
 Label '//tools:compile_only_aspect.bzl' is invalid because 'tools' is not a package; perhaps you meant to put the colon here: '//:tools/compile_only_aspect.bzl'?
```
when running:
```
bazel build -s tests/ios/app:App --aspects rules/compile_only_aspect.bzl%compile_only_aspect --output_groups=compiles
```
We need a BUILD.bazel file in tools

PR #651, opened 2023-01-31
Feed custom `module_map` to compile actions

jerrymarino:
At some point in recent updates/bumps we hit a regression which caused custom module maps to no longer be propagated to compile actions. This broke remote execution and caching.

Ensures the `input` maps also are sent to `objc_library` - prior to overriding it in the proceeding line of code.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -945,6 +945,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
                 module_name = module_name,
                 tags = _MANUAL,
             )
+            objc_hdrs.append(module_map)
             module_map = "%s.extended.modulemap" % name
 
     # Note: this needs to go here, in order to virtualize the extended module


rules/library.bzl
```diff
@@ -945,6 +945,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
                 module_name = module_name,
                 tags = _MANUAL,
             )
+            objc_hdrs.append(module_map)
```
mattrobmattrob, line 948: You don't want the extended module map here intentionally, right?

mattrobmattrob (approved):

thiagohmcruz (approved):

rules/library.bzl
jerrymarino: I'll try to circle back on this soon, it looks like we're having cache invalidation issues on a lot of these 🤔

jerrymarino:
Going to close this one out for now, it seems to causes other issues If we propagate it like this

PR #652, opened 2023-02-07
[hmap][Bazel 6.0] Harden hmap writes

jerrymarino:
Harden hmap logic. After updating to Bazel 6.0 I hav observed several invalid hmap files in `bazel-out` and `--disk_cache`s under load only. The system in question was running Bazel 6, intel, and macOS 12.5.2. This hardens the hmap code to fix 2 obscure/intermittent categories of failure.

1. Fix hmapbuild return code

First if a write fails in rules_ios - in  `hmapbuilder.c` isn't propagated. This is a longstanding bug and maybe masking other problems.

2. Call fsync to guard against data corruption

Per the man page

> fsync() causes all modified data and attributes of fildes to be moved to
> a permanent storage device. This normally results in all in-core
> modified copies of buffers for the associated file to be written to a
> disk.

### DIFF
--- a/rules/hmap/hmap.c
+++ b/rules/hmap/hmap.c
@@ -191,22 +191,25 @@ int hmap_save(HeaderMap* hmap, char* path) {
         perror(path);
         return 1;
     }
+    int rc = 0;
     if (write(fd, hmap->data, hmap->stringsTableNextEntry) !=
         hmap->stringsTableNextEntry) {
         perror(path);
-        return 1;
+        rc = 1;
     }
 
-    if (fsync(fd)) {
+    // Don't fsync if it failed
+    if (rc != 1 && fsync(fd) < 0) {
         perror(path);
-        return  1;
+        rc = 1;
     }
 
+    // Close regardless of failure
     if (close(fd)) {
         perror(path);
-        return  1;
+        rc = 1;
     }
-    return 0;
+    return rc;
 }
 
 void hmap_dump(HeaderMap* hmap) {
--- a/rules/hmap/hmap.c
+++ b/rules/hmap/hmap.c
@@ -191,19 +191,22 @@ int hmap_save(HeaderMap* hmap, char* path) {
         perror(path);
         return 1;
     }
-    int rc = 0;
     if (write(fd, hmap->data, hmap->stringsTableNextEntry) !=
         hmap->stringsTableNextEntry) {
         perror(path);
-        rc = 1;
+        return 1;
     }
 
-    if (fsync(fd) < 0) {
-        rc = 1;
+    if (fsync(fd)) {
+        perror(path);
+        return  1;
     }
 
-    close(fd);
-    return rc;
+    if (close(fd)) {
+        perror(path);
+        return  1;
+    }
+    return 0;
 }
 
 void hmap_dump(HeaderMap* hmap) {
--- a/rules/hmap/hmap.c
+++ b/rules/hmap/hmap.c
@@ -197,6 +197,11 @@ int hmap_save(HeaderMap* hmap, char* path) {
         perror(path);
         rc = 1;
     }
+
+    if (fsync(fd) < 0) {
+        rc = 1;
+    }
+
     close(fd);
     return rc;
 }
--- a/rules/hmap/hmapbuild.c
+++ b/rules/hmap/hmapbuild.c
@@ -60,11 +60,12 @@ int main(int ac, char **av) {
             fprintf(stderr, "failed to add '%s' to hmap\n", m->key);
         }
     }
-    if (hmap_save(hmap, cli_args.output_file)) {
+    int rc = hmap_save(hmap, cli_args.output_file);
+    if (rc) {
         perror(cli_args.output_file);
     }
     hmap_free(hmap);
-    return 0;
+    return rc;
 }
 
 static void usage() {
--- a/rules/hmap/hmapbuild.c
+++ b/rules/hmap/hmapbuild.c
@@ -55,15 +55,14 @@ int main(int ac, char **av) {
     }
     debug("Writing hmap with %u entries", numEntries);
     HeaderMap *hmap = hmap_new(numEntries);
+    int rc = 0;
     for (m = entries; m != NULL; m = m->hh.next) {
         if (hmap_addEntry(hmap, m->key, m->value)) {
             fprintf(stderr, "failed to add '%s' to hmap\n", m->key);
+            rc |= 1;
         }
     }
-    int rc = hmap_save(hmap, cli_args.output_file);
-    if (rc) {
-        perror(cli_args.output_file);
-    }
+    rc |= hmap_save(hmap, cli_args.output_file);
     hmap_free(hmap);
     return rc;
 }
--- a/rules/hmap/hmapbuild.c
+++ b/rules/hmap/hmapbuild.c
@@ -55,13 +55,12 @@ int main(int ac, char **av) {
     }
     debug("Writing hmap with %u entries", numEntries);
     HeaderMap *hmap = hmap_new(numEntries);
-    int rc = 0;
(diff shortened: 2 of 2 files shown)


luispadron (approved):

justinseanmartin (approved):

thiagohmcruz (approved):

mattrobmattrob (approved):

rules/hmap/hmap.c
```diff
@@ -197,6 +197,11 @@ int hmap_save(HeaderMap* hmap, char* path) {
         perror(path);
         rc = 1;
     }
+
+    if (fsync(fd) < 0) {
```
gyfelton, line 201: TIL

ob:
Interesting... thanks for tightening the error checking, I added a bit more checks to see if we can catch what is happening. 

Do you have examples of the corrupted hmap files? Are they empty? Missing info? Full of NULLs?

Adding the call to `fsync()` should be unnecessary because `close()` and process exit semantics are supposed to guarantee flushing the buffers. I left it in but note it'll have about a 15% performance penalty (which might be fine since hmap generation is the fastest piece in the build).

```
rules_ios ❯ hyperfine "./test" "./test -s"
Benchmark 1: ./test
  Time (mean ± σ):      1.136 s ±  0.028 s    [User: 0.009 s, System: 1.058 s]
  Range (min … max):    1.100 s …  1.182 s    10 runs
 
Benchmark 2: ./test -s
  Time (mean ± σ):      1.288 s ±  0.051 s    [User: 0.011 s, System: 1.056 s]
  Range (min … max):    1.236 s …  1.389 s    10 runs
 
Summary
  './test' ran
    1.13 ± 0.05 times faster than './test -s'
```

That `test.c` is essentially:

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/errno.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/stat.h>


int main(int ac, char **av) {
    const char *dir = "/tmp";
    const char *file = "hmaptest.hmap";
    char path[1024];
    char buf[1024];
    int do_fsync = 0;

    if ( ac > 1 && strcmp(av[1], "-s") == 0) {
        do_fsync = 1;
    }

    snprintf(path, sizeof(path), "%s/%s", dir, file);
    int NTIMES = 10000;
    for (int i = 0; i < NTIMES; ++i) {
        int fd = open(path, O_WRONLY | O_CREAT, 0666);
        if (fd < 0) {
            printf("open failed: %s\n", strerror(errno));
            return 1;
        }
        int rc = write(fd, buf, sizeof(buf));
        if (rc < 0) {
            printf("write failed: %s\n", strerror(errno));
            return 1;
        }
        if (do_fsync) {
            if ( fsync(fd) < 0 ) {
                printf("fsync failed: %s\n", strerror(errno));
                return 1;
            }
        }
        close(fd);
        struct stat st;
        if ( stat(path, &st) < 0 ) {
            printf("stat failed: %s\n", strerror(errno));
            return 1;
        }
        if ( st.st_size != sizeof(buf) ) {
            printf("size mismatch: %lld != %ld\n", st.st_size, sizeof(buf));
            return 1;

        }
        if ( unlink(path) < 0 ) {
            printf("unlink failed: %s\n", strerror(errno));
            return 1;
        }
    }
    return 0;
}
```

thiagohmcruz:
> 

@ob We were able to repro the issue on CI workers, from one of the "bad" workspaces I have access to I get (redacted the fmw name):
```sh
$ file bazel-out//ios-x86_64-min14.0-applebin_ios-ios_x86_64-dbg-ST-2db1df0745c8/bin/Frameworks/SomeInternalFmw/SomeInternalFmw_private_hmap.hmap
bazel-out//ios-x86_64-min14.0-applebin_ios-ios_x86_64-dbg-ST-2db1df0745c8/bin/Frameworks/SomeInternalFmw/SomeInternalFmw_private_hmap.hmap: empty
```

During the internal incident we had because of this I remember finding non-empty `.hmap` files but with only `37B` in it and the content was:
```sh
$ cat bazel-out//ios-arm64-min12.0-applebin_ios-ios_arm64-dbg-ST-acb08275552e/bin/Pods/Foo/Foo_private_hmap.hmap
pamh$% 
```

jerrymarino:
Yeah the problem with the original code is it doesn’t check any error codes - so Bazel will persit a bad hmap when writes ( or Hmap creation ) fails.

This said, I’m hitting an un-root-caused issue with the disk cache in Bazel 6.0. These changes were added to guard against misc failures - not explicitly to fix this corruption. What’s odd is it’s only hmaps that pop as messed up.

A penalty on these files would be negligible because they are tiny but just might help stability. Further it’s less of an perf issue because on macOS APFS fsync is buffered and is not a true sync like the standard would have you believe anyhow ( see the man page for what I mean 🤕)

I’ll re post when we get to the bottom of this disk cache corruption. ATM we are making sure the LTS release version ( 5.0 ) is working so we can continue to update rules JFYI @chenxiao0228

jerrymarino:
PS I’ve both added details and flagged the disk cache issue if you are wanting to know about it:
https://github.com/bazelbuild/bazel/issues/17428#issuecomment-1421052652

The potential way reality is that _if_ this is the only broken rule with disk_cache we can stop disk caching it but I’d like to get to the bottom of it first: if the disk cache is corrupted for all rules it will be an insidious problem so I’m avoiding Bazel 6 until we find the issue 😅

rules/hmap/hmap.c
```diff
@@ -191,14 +191,22 @@ int hmap_save(HeaderMap* hmap, char* path) {
         perror(path);
         return 1;
     }
-    int rc = 0;
```
jerrymarino, line 194: @ob this change removed the `close` on failure which I previously had handled.  I'm not really sure it matters in the context of this program but it's probably more correct to close the open descriptors in general

rules/hmap/hmapbuild.c
```diff
@@ -55,16 +55,16 @@ int main(int ac, char **av) {
     }
     debug("Writing hmap with %u entries", numEntries);
     HeaderMap *hmap = hmap_new(numEntries);
+    int rc = 0;
     for (m = entries; m != NULL; m = m->hh.next) {
         if (hmap_addEntry(hmap, m->key, m->value)) {
             fprintf(stderr, "failed to add '%s' to hmap\n", m->key);
+            rc |= 1;
```
jerrymarino, line 62: AFAIK this will cause an issue for few rules I've seen internally and I think it's better to address these logic errors outside of the broken write code

rules/hmap/hmap.c
ob: it'll close on exit but sure, probably cleaner to close it....

rules/hmap/hmapbuild.c
ob: The only way `hmap_addEntry()` can return non-zero is if the table is full:

https://github.com/bazel-ios/rules_ios/blob/985d578d8d4c3e4b6fe23dc69436ecafa33d30b4/rules/hmap/hmap.c#L150-L154

I think this would be important to catch and abort no?

jerrymarino: Yeah it's probably fine to do would just rather batch to a smaller PR and I can resubmit this, with this impending disk cache problem I wanted to keep the changes minimal 

PS: I dug through logs and some funky build files I've seen rely on is this one:
```
static void add_entry(mapping **hashmap, char *key, char *value) {
    mapping *entry = NULL;
    HASH_FIND_STR(*hashmap, key, entry);
    if (entry) {
        // key already in the hash, verify the value
        if (strcmp(value, entry->value)) {
            fprintf(stderr,
                    "WARNING: header '%s' already in cache as '%s' ignoring "
                    "new value: '%s'\n",
                    key, entry->value, value);
        }
        return;
    }
```

PR #655, opened 2023-02-08
[LTS][Bazel 5] Provide minimum LTS support

jerrymarino:
From time to time we'll evaluate supporting this to allow us to all run on HEAD and keep pushing / working together with minimal dammage.

Detect if the user is on Bazel 5+ and fallback to our tag of `rules_apple` and corresponding API. Because we depend on some of the more stable internals, this isn't such a big deal for now. Add CI vetting and a README note on the status.

Outside of this, a minor fix make it work for me: disable strict version validation we added around deps.

Bazel 6.0 has a number of open questions that will be answered later

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,6 +70,28 @@ jobs:
           name: bazel-testlogs
           path: bazel-testlogs
 
+  lts_ios_integration_tests:
+    name: Build and Test ( Virtual Frameworks + LTS )
+    runs-on: macos-12
+    steps:
+      - uses: actions/checkout@v3
+      - name: Select Xcode
+        run: .github/workflows/xcode_select.sh
+      - name: Build and Test
+        run: |
+          export USE_BAZEL_VERSION=5.3.2
+          bazelisk build --features apple.virtualize_frameworks \
+              --local_test_jobs=1 \
+              --apple_platform_type=ios \
+              --deleted_packages='' \
+              -- //tests/ios/... \
+              -//tests/ios/frameworks/sources-with-prebuilt-binaries/...
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: bazel-testlogs
+          path: bazel-testlogs
+
   build_arm64_simulator:
     name: Build arm64 Simulator
     runs-on: macos-12
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,9 +81,7 @@ jobs:
         run: |
           export USE_BAZEL_VERSION=5.3.2
           bazelisk build --features apple.virtualize_frameworks \
-              --local_test_jobs=1 \
-              --apple_platform_type=ios \
-              --deleted_packages='' \
+              --config=ios \
               -- //tests/ios/... \
               -//tests/ios/frameworks/sources-with-prebuilt-binaries/...
       - uses: actions/upload-artifact@v2
--- a/README.md
+++ b/README.md
@@ -172,14 +172,34 @@ Bazel version required by current rules is [here](https://github.com/bazel-ios/r
 for the documentation.
 
 
-### 5.x.x LTS Support
+## 5.x.x LTS Support on HEAD
 
 LTS is a concept in [Bazel for versions](https://blog.bazel.build/2020/11/10/long-term-support-release.html)
+and we currently support 5.x.x only for a brief time. _The rough ETA for
+removing Bazel 5.x.x LTS support is end of Q2 2023 and if it's an issue we can
+correct course sooner._
 
-Because `rules_ios` should be loosely coupled to a given Bazel versions, we
-often handle several Bazel versions concurrently. Our users have a large
-variance of appetite for breakage, custom rules, application complexity, and
-infrastructure.  Dictating a single Bazel version is not possible amongst
-`rules_ios` users.  Because we want to keep running on HEAD ( and have everyone
-do so ) LTS support capabilities will often come into play and the burden to be
-an outlier should fall on the outliers here.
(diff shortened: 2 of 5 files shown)


jerrymarino:
Bazel 6.0 caused a few perf and stability issues in a few bigger use cases so we'll have to do more vetting of it there.

congt (approved):

justinseanmartin (approved):

README.md
```diff
@@ -171,3 +171,15 @@ Bazel version required by current rules is [here](https://github.com/bazel-ios/r
 [Click here](https://github.com/bazel-ios/rules_ios/tree/master/docs)
 for the documentation.
 
+
+### 5.x.x LTS Support
+
+LTS is a concept in [Bazel for versions](https://blog.bazel.build/2020/11/10/long-term-support-release.html)
+
+Because `rules_ios` should be loosely coupled to a given Bazel versions, we
```
thiagohmcruz, line 179:
```suggestion
Because `rules_ios` should be loosely coupled to a given Bazel version, we
```

rules/framework.bzl
```diff
@@ -128,6 +128,10 @@ def _validate_deps_minimum_os_version(framework_name, minimum_os_version, deps):
     deps: the deps of this framework
     """
 
+    # TODO @cshi we need a number of fixes to enable this
+    if True:
```
thiagohmcruz, line 132: Can we file a GH issue to address this?

rules/repositories.bzl
```diff
@@ -59,14 +69,29 @@ def rules_ios_dependencies():
...
-    _maybe(
-        github_repo,
-        name = "build_bazel_rules_apple",
-        ref = "b43c7f60b584d68d7d187c236d4328162ba2e806",
-        project = "bazelbuild",
-        repo = "rules_apple",
-        sha256 = "d3e8549c0c8966ba0e547a02f5601440be25a0b8143bd57ab5834b65b02c22be",
-    )
+    bazel_version = _get_bazel_version()
+    if bazel_version.major == "5":
```
thiagohmcruz, line 73: I think it would be helpful for others to add instructions in the `README` or as comments here on how to proceed to bump `rules_apple` in the future. It might be obvious what to do for some but IMO we shouldn't assume that that's the case for all contributors.

thiagohmcruz (approved):

.github/workflows/tests.yml
```diff
@@ -70,6 +70,28 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Select Xcode
+        run: .github/workflows/xcode_select.sh
+      - name: Build and Test
+        run: |
+          export USE_BAZEL_VERSION=5.3.2
+          bazelisk build --features apple.virtualize_frameworks \
+              --local_test_jobs=1 \
+              --apple_platform_type=ios \
+              --deleted_packages='' \
```
mattrobmattrob, line 86: Since this is all configured in `xcode_select.sh`, could simplify this a bit:
```suggestion
              --config=ios \
```

rules/framework.bzl
```diff
@@ -591,6 +595,14 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
         # processed infoplit its possible, but validate for the common case
         fail("Missing bundle_id: Info.plist actions require one")
 
+    # Determine the interface version of rules_apple. We don't want to force
+    # "hail mary" type rules and Bazel bumps on the community to run a given
+    # version of rules_ios and the API we depend on is relatively stable. If
+    # this ceases to be the case than consider mainlining the few components we
+    # use from it to remove this complexity.
+    rules_apple_api_version = getattr(bundling_support, "rule_api_version", None)
```
mattrobmattrob, line 603: Where does this attribute come from? I can't seem to find it in https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/bundling_support.bzl or its history.

mattrobmattrob (approved):

rules/repositories.bzl
jerrymarino: My current thinking about this is in the README ATM we can expand it
> the burden to be an outlier should fall on the outliers here.

In practice this is less of an issue now that the PR allows us to run many versions of Bazel with rules_ios HEAD and from a maintainer perspective caring about 6.0 as long as the `rules_apple_api` is upheld. If the API `rules_apple_api_version` is divergent, then you'll just need to add a simple if statement for the new API. You can think about this in a similar way as large scale migrations or similar to code that supports on 2 python versions. 

e.g. @mattrobmattrob  they can continue to bump rules_apple and generally not have to reason about how to back porting features for LTS support and our CI vets that it's green. 

I estimate burning this LTS 5.0 support pretty anyhow - say 6 months. We kept around 4.0 support for a similar time

rules/framework.bzl
mattrobmattrob: I had missed that is was `bazel-ios` instead of `bazelbuild`. 🤦

jerrymarino: You've got a good point - let me update the repo to point at `bazel-ios` to make it a bit more clear.

jerrymarino: I've updated the README to address any concerns on this and clarify that you won't have to worry about LTS support if you're not running it:
https://github.com/bazel-ios/rules_ios/pull/655/commits/e7a635b68a046846cc26370a87c7f134d51363c5#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R182

rules/framework.bzl
jerrymarino: The root of why it's broken in the mono-repo - so it might be tracked elsewhere - cc @congt

PR #656, opened 2023-02-08
Fixes the compile_only_aspect

congt:
There may be many ObjCompile actions in a objc_library target.

### DIFF
--- a/tools/compile_only_aspect.bzl
+++ b/tools/compile_only_aspect.bzl
@@ -20,10 +20,7 @@ def _compile_only_aspect_impl(target, ctx):
                 outs = action.outputs
                 break
     elif ctx.rule.kind == "objc_library":
-        for action in target.actions:
-            if action.mnemonic == "ObjcCompile":
-                outs = action.outputs
-                break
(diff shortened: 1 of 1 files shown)


tools/compile_only_aspect.bzl
```diff
@@ -15,15 +15,9 @@ def _compile_only_aspect_impl(target, ctx):
     # Consider in Swift compilation how to only generate the intefaces
     # through the feature of interface splitting
     if ctx.rule.kind == "swift_library":
-        for action in target.actions:
-            if action.mnemonic == "SwiftCompile":
```
jerrymarino, line 19: @congt If the action is a `swift_library` is it possible there are other action types in there or no? This seems fine none the less - my thought was it might add extra actions other than a `SwiftCompile` we dont want

jerrymarino (approved):
LGTM - minor inline - thanks for the fix!

tools/compile_only_aspect.bzl
congt: I only see one SwiftCompile action. I can revert this line.

congt: Reverted

mattrobmattrob (approved):

Issue #654, opened 2023-02-07
Fix xcspec parsing in Xcode 14.2+


luispadron:
Some of the parsing seems to have broken in Xcode 14.2 xcspec eval parser since it's causing some Starlark crashes while performing boolean operations on strings.

More info and temporary workaround: https://github.com/bazel-ios/rules_ios/pull/653

PR #653, opened 2023-02-07
Fix boolean operation Starlark xcconfig crash in Xcode 14.2 xcspec

luispadron:
# Summary

Regression introduced in #644 due to change in `xcspec_evals.bzl`

The issue is the `DEFINES_MODULE` xcconfig expects a string `YES` or `NO`, the code now does a check for `"<YES/NO>" and "<YES/NO>"` which causes a Starlark crash. 

Instead, revert back to using `var == "YES"` to get the correct boolean type out of the xcconfig value. 

We should follow up to understand _why_ this was caused.

I added a test to ensure we catch something similar in the future with that particular value.

## Error

```
bazel build //tests/ios/app:AppWithConfigurableAttrs

Analyzing: target //tests/ios/app:AppWithConfigurableAttrs (59 packages loaded, 1262 targets configured)
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'ConfiguredTargetKey{label=//tests/ios/app:AppWithConfigurableAttrs_objc, config=BuildConfigurationKey[9ea6e389a26228ecb8df2e1d4412bec7804688be29c838be3a56bc00782d436b]}' (requested by nodes 'ConfiguredTargetKey{label=//tests/ios/app:AppWithConfigurableAttrs.force_load_direct_deps, config=BuildConfigurationKey[9ea6e389a26228ecb8df2e1d4412bec7804688be29c838be3a56bc00782d436b]}', 'ConfiguredTargetKey{label=//tests/ios/app:AppWithConfigurableAttrs.dep_middleman, config=BuildConfigurationKey[9ea6e389a26228ecb8df2e1d4412bec7804688be29c838be3a56bc00782d436b]}', 'ConfiguredTargetKey{label=//tests/ios/app:AppWithConfigurableAttrs.framework_middleman, config=BuildConfigurationKey[5f63a04dfbb11fae3a7047038bee955d71f88f5203d9780a0b6fab50f5ad7344]}')
        at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:642)
        at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:382)
        at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(Unknown Source)
[...]

PR #658, opened 2023-02-08
Fix xcspec evals for Xcode 14.2

justinseanmartin:
This fixes the root cause underlying #654 (and overwrites the manual stopgap from #653). 

In Xcode 14, Apple changed their xcspecs to use implicit booleans as opposed to explicitly checking against YES/NO string values. For example, in the `Clang LLVM 1.0.xcspec`:
```
     {
        Name = "CLANG_ENABLE_MODULE_IMPLEMENTATION_OF";
        Type = Boolean;
        DefaultValue = YES;
        CommandLineArgs = {
          NO = ();
          YES = ("-fmodule-name=$(PRODUCT_MODULE_NAME)");
        };
        Condition = "$(CLANG_ENABLE_MODULES) == YES && $(DEFINES_MODULE) == YES";
      },
```

Became:
```
      {
        Name = "CLANG_ENABLE_MODULE_IMPLEMENTATION_OF";
        Type = Boolean;
        DefaultValue = YES;
        CommandLineArgs = {
          NO = ();
          YES = ("-fmodule-name=$(PRODUCT_MODULE_NAME)");
        };
        Condition = "$(CLANG_ENABLE_MODULES) && $(DEFINES_MODULE)";
      },
```

The changes in this PR should generate effectively equivalent logic to what was there previously. I haven't looked into the root cause of the Bazel crash called out in #653, but instead am restoring the prior behavior in our xcspec bazel codegen output.

For some unclear reason, the behavior of running the script on my local machine leads to some differences in whitespace and quotes being used, so @luispadron ran and pushed that for me.

### DIFF
--- a/data_generators/xcspec_extractor.py
+++ b/data_generators/xcspec_extractor.py
@@ -5,7 +5,7 @@ Extracts xcspec info from a given Xcode developer dir
 and prints out a bzl file that contains a single constant, `SETTINGS`,
 that can be used to access the xcspecs from starlark.
 
-Usage: xcspec_extractor.py /Application/Xcode.app/Contents/Developer 11.2.1
+Usage: xcspec_extractor.py /Application/Xcode.app/Contents/Developer 11.2.1 ../data/xcspecs.bzl  ../data/xcspec_evals.bzl
 """
 
 import plistlib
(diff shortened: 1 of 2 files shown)


thiagohmcruz (approved):

data_generators/xcspec_extractor.py
```diff
@@ -5,7 +5,7 @@
 and prints out a bzl file that contains a single constant, `SETTINGS`,
 that can be used to access the xcspecs from starlark.
 
-Usage: xcspec_extractor.py /Application/Xcode.app/Contents/Developer 11.2.1
+Usage: xcspec_extractor.py /Application/Xcode.app/Contents/Developer 11.2.1 ../data/xcspecs.bzl  ../data/xcspec_evals.bzl
```
luispadron, line 8: Intentional?

luispadron (approved):

luispadron:
Thanks for fixing root cause and write-up!

luispadron:
Closes https://github.com/bazel-ios/rules_ios/issues/654

data_generators/xcspec_extractor.py
justinseanmartin: Yea, those are actual required parameters when running it directly from python. Probably could also document using `bazel run //data_generators:extract_xcspecs` instead

mattrobmattrob (approved):
👏

PR #661, opened 2023-02-23
[DNR][DNM][HOLD] Fixes for indexing in Xcode 14.x

thiagohmcruz:


rules/legacy_xcodeproj.bzl
```diff
@@ -595,8 +595,14 @@ def _header_search_paths_for_target(target_name, all_transitive_targets):
     header_search_paths.append("\"$BAZEL_WORKSPACE_ROOT\"")
     return " ".join(header_search_paths)
 
-def _swift_copts_for_target(target_name, all_transitive_targets):
+def _swift_copts_for_target(target_name, all_transitive_targets, mixed = False):
     copts = []
+    if mixed:
```
jerrymarino, line 600: The fact that this never read `copts` is one of the larger issues with it but will go away when we remove this code. This seems fine for now

rules/legacy_xcodeproj.bzl
```diff
@@ -881,6 +897,18 @@ def _populate_xcodeproj_targets_and_schemes(ctx, targets, src_dot_dots, all_tran
             target_settings["TARGETED_DEVICE_FAMILY"] = target_info.targeted_device_family
 
         pre_build_scripts = []
+
+        if virtualize_frameworks:
+            pre_build_scripts.append({
```
jerrymarino, line 902: Is there a way that we could prevent syncing frameworks when `virtualize_frameworks` is set - perhaps pass a new argument to the installed script.

If a developer builds a framework AFAIK there is an issue with it

rules/legacy_xcodeproj.bzl
```diff
@@ -1058,6 +1087,7 @@ def _xcodeproj_impl(ctx):
         "BAZEL_CONFIGS": ctx.attr.configs.keys(),
         "BAZEL_ADDITIONAL_BAZEL_BUILD_OPTIONS": " ".join(["{} ".format(opt) for opt in ctx.attr.additional_bazel_build_options]),
         "BAZEL_ADDITIONAL_LLDB_SETTINGS": "\n".join(ctx.attr.additional_lldb_settings),
+        "INDEX_DATA_STORE_DIR": "$(INDEX_DATA_STORE_DIR)",
```
jerrymarino, line 1090: So this is set by xcconfig but no longer passed to the script? This seems fine

rules/library.bzl
```diff
@@ -991,31 +991,32 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-            name = swift_angle_bracket_hmap_name,
-            namespace = namespace,
-            hdrs = [],
-            direct_hdr_providers = [swift_libname],
-            tags = _MANUAL,
-            testonly = testonly,
-        )
-        private_dep_names.append(swift_angle_bracket_hmap_name)
-        _append_headermap_copts(swift_angle_bracket_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
+        if objc_sources or cpp_sources:
```
jerrymarino, line 994: Is this change necessary / related to the indexing bit? Does it just dump garbage targets?

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
```diff
@@ -24,10 +24,14 @@ com.apple.product-type.framework.static)
     # We only care about the entry ending with TARGET_NAME" so that we can get the path to its directory
     QUERY=$(grep -A 2 important_output "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep ${TARGET_NAME}\" | sed "s/uri: \"file:\/\///")
     if [[ -z $QUERY ]]; then
-        # For virtual frameworks, this is disabled
-        exit 0
+        # For virtual frameworks, this is expected to be empty. Nothing to do.
```
jerrymarino, line 27: See the other comment - it shouldn't really install the framework to the directory. Perhaps in this step we'd remove them?

https://github.com/bazel-ios/rules_ios/pull/661/files#diff-a15f63f3a76f05dbdfb6a9b817d2d5b1dfd28aa94bafc362eca9bd53f0dd77e4R902

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
```diff
@@ -58,55 +62,58 @@ com.apple.product-type.app-extension)
     exit 1
     ;;
 esac
-output="$TARGET_BUILD_DIR/$FULL_PRODUCT_NAME"
 
-mkdir -p $OBJECT_FILE_DIR_normal/$CURRENT_ARCH/
-chmod -R +w $OBJECT_FILE_DIR_normal/$CURRENT_ARCH/
+if [ -n "$input_options" ]; then
```
jerrymarino, line 66: See the comment for the virtual frameworks case, that should probably not be installing or creating anything

jerrymarino (commented):
Thanks for fixing this @thiagohmcruz. A few misc questions inline, I'm with you on not boiling the ocean to fix it and we should expedite the transition to XCHammer instead of dealing with the shortcomings here. A few minor tweaks might make it a bit nicer

jerrymarino (approved):

rules/legacy_xcodeproj.bzl
thiagohmcruz: This is set here because otherwise `INDEX_DATA_STORE_DIR` is empty during scripts phase. Tested this in both Xcode 13.x and 14.x and when this setting is set `INDEX_DATA_STORE_DIR` contains the respective path to the index store. Most likely Xcode needs this value explicitly requested in a build setting to make it non-empty during script phase.

rules/library.bzl
thiagohmcruz: I hit an issue in a pure swift app where the indexing `swifc` invocation would ask for this hmaps but they were never created causing the indexing invocation to fail. Will add a test case for this.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
thiagohmcruz: That's probably an improvement from what I have if it works, will try to remove the problematic fmw paths here and see how it goes. The issue is that `XCBBuildService` creates these framework paths under `DerivedData` so I don't think it's our code creating them but will double check. 

Even if there's an empty `.framework` directory under `DerivedData` the `clang/swiftc` indexing invocation in Xcode will fail due to missing headers/modulemaps and ignore the VFS overlay flag in the same invocation. Honestly this might be a `clang` bug but not sure.

Will try to clean things up at this step and report back. It should indeed work.

rules/legacy_xcodeproj.bzl
thiagohmcruz: Most issues I've found were when building a framework, building apps taking dep on fmws seems to be bypassing some edge cases here.

The malformed frameworks being created under `DerivedData` are created when Xcode launches by `XCBBuildService` and not our post-build scripts (elaborated more [here](https://github.com/bazel-ios/rules_ios/pull/661#discussion_r1122080506)). Since the current generator is missing a lot of logic to get this right I believe a pre-build cleanup is a more effective workaround. Def not ideal but again, we're moving to `XCHammer` soon.

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
```diff
@@ -24,10 +24,14 @@ com.apple.product-type.framework.static)
     # We only care about the entry ending with TARGET_NAME" so that we can get the path to its directory
     QUERY=$(grep -A 2 important_output "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep ${TARGET_NAME}\" | sed "s/uri: \"file:\/\///")
     if [[ -z $QUERY ]]; then
-        # For virtual frameworks, this is disabled
-        exit 0
+        # For virtual frameworks, this is expected to be empty. Nothing to do.
```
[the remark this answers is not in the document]
thiagohmcruz: 1. At least as far as I was able to debug this I don't think we can easily (without `xcbuildkit`) prevent `XCBBuildService` from creating it but I could be wrong and missed something here. Still I'd rather make the legacy generator work at some capacity instead of continuing to fix something that is going away soon (see [comment](https://github.com/bazel-ios/rules_ios/pull/661#discussion_r1122168280))

2. My concern with post build is cleaning up more things than necessary and end up removing post-build outputs that we actually need, I figured a pre-build cleanup step would be more reliable to clean what `XCBuildService` did on Xcode launch. No strong opinions here though.

rules/library.bzl
thiagohmcruz: I can't seem to repro this 🤔, I was in the middle of debugging something when I saw so maybe it was bad local state or something else that I missed.

Will revert for now, if we hit other issues in the future I can follow up and propose a proper fix.

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/_swiftmodule.sh
```diff
@@ -22,11 +26,11 @@ for index_dir in "${dirs[@]}"; do
         sourceinfo_dir="$module_bundle/Project"
         mkdir -p "$sourceinfo_dir"
 
-        cp "$module" "$module_bundle/$CURRENT_ARCH.swiftmodule" || true
-        cp "$doc" "$module_bundle/$CURRENT_ARCH.swiftdoc" || true
-        cp "$sourceinfo" "$sourceinfo_dir/$CURRENT_ARCH.swiftsourceinfo" || true
+        cp "$module" "$module_bundle/${ARCHS[0]}.swiftmodule" || true
+        cp "$doc" "$module_bundle/${ARCHS[0]}.swiftdoc" || true
+        cp "$sourceinfo" "$sourceinfo_dir/${ARCHS[0]}.swiftsourceinfo" || true
```
thiagohmcruz, line 31: FYI `CURRENT_ARCH` was always `unsupported` for me, had to use `ARCHS` and pick the first element to get the expected value (e.g. `arm64` on M1)

tools/xcodeproj_shims/installers/_swiftmodule.sh
```diff
@@ -22,11 +26,11 @@ for index_dir in "${dirs[@]}"; do
         sourceinfo_dir="$module_bundle/Project"
         mkdir -p "$sourceinfo_dir"
 
-        cp "$module" "$module_bundle/$CURRENT_ARCH.swiftmodule" || true
-        cp "$doc" "$module_bundle/$CURRENT_ARCH.swiftdoc" || true
-        cp "$sourceinfo" "$sourceinfo_dir/$CURRENT_ARCH.swiftsourceinfo" || true
+        cp "$module" "$module_bundle/${ARCHS[0]}.swiftmodule" || true
+        cp "$doc" "$module_bundle/${ARCHS[0]}.swiftdoc" || true
+        cp "$sourceinfo" "$sourceinfo_dir/${ARCHS[0]}.swiftsourceinfo" || true
```
thiagohmcruz, line 31: FYI `CURRENT_ARCH` was always `unsupported` for me, had to use `ARCHS` and pick the first element to get the expected value (e.g. `arm64` on M1)

tools/xcodeproj_shims/installers/indexstores.sh
```diff
@@ -6,7 +6,7 @@ set -euo pipefail
 
 echo "Start remapping index files at `date`"
 
-FOUND_INDEXSTORES=`grep -o 'command_line: "\(.*\.indexstore\)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | sed 's/.*command_line: "//' | sort | uniq` || true
+FOUND_INDEXSTORES=`grep -o 'command_line: "\(.*\.indexstore\)\|command_line: "\(.*_global_index_store\)' $BAZEL_BUILD_EVENT_TEXT_FILENAME | sed 's/.*command_line: "//' | sed 's/-Xwrapped-swift=-global-index-store-import-path=//' | sort | uniq` || true
```
thiagohmcruz, line 9: This regex was not picking up all stores we need, this should now pick the swift index stores and also the correct path to the global one where ObjC units/records are.

tools/xcodeproj_shims/installers/swiftmodules.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 # Copy Bazel build `.swiftmodule` files to `DerivedData`. This is used by Xcode
 # and its indexing.
 echo "Start copying swiftmodules at `date`"
-FOUND_SWIFTMODULES=`grep -A 2 primary_output "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep "\.swiftmodule" | sed -e "s/uri: \"file:\/\///" -e "s/\"\$//" || true`
+FOUND_SWIFTMODULES=`grep -A 2 -E "important_output|primary_output" "$BAZEL_BUILD_EVENT_TEXT_FILENAME" | grep -w uri | grep -Eo "bazel-out.*\.swiftmodule" | sed -e "s/uri: \"file:\/\///" -e "s/\"\$//" | sort | uniq || true`
```
thiagohmcruz, line 8: This was missing outputs in many cases, added `important_output` to pick the missing ones and `sort | uniq` to avoid duplicates

justinseanmartin (approved):

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
thiagohmcruz: Btw I tried to clean things in this step but at least in the repro that I had that wasn't enough to fix the indexing invocations

Other state changes in Xcode apparently trigger adding the malformed fmws under DD. Will stick with the pre-build workaround for now.

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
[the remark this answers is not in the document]
jerrymarino: Ah so basically you're hitting breakages from the build service creating them? I think it's still possible for it to break then right?

rules/library.bzl
thiagohmcruz: I think CI's repro-ed what I was talking about: https://github.com/bazel-ios/rules_ios/actions/runs/4307287081/jobs/7512162662
```sh
Make sure hmap files exist after build
File at ../../../bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-d31cf6d5fbab/bin/tests/ios/frameworks/dynamic/b/b_swift_doublequote_hmap.hmap not found!
```

That dynamic fmw is not going to produce `_swift_doublequote_hmap` since the `objc_library` is never built due to this target not having ObjC sources. 

I might have to bring back the conditional I had added here. Will double check if the test is searching for a valid list of hmaps first.

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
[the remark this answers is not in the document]
thiagohmcruz: If the build service creates these under DD at different times it's possible to hit the indexer failing issues, yes. Then another build should fix it and clean things up again. Based on my testing though after the cleanup happened I navigated to different sources, edited files, changed schemes back and forth and DD continued to be in a good state. If there's something else that triggers the build service re-creating those it's not something that happen often. That's why I'm calling things a workaround, I don't have evidence that this fixes it for good (and it was not the intent).

rules/library.bzl
thiagohmcruz: That said, I'm bringing back the conditional.

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
jerrymarino: Ok seeing you answered the other Q here, perhaps then it was an Xcode 14 issue which started creating them on demand. I just wonder if we can simply turn that feature off - they used to have some xcconfigs related to it.

Overall what you've got seems resonable, but if we don't address this issue the issue will live on: but make moving over to our new generator that much better 🥳

thiagohmcruz: Yeah tbh I don't know where exactly that happens in Xcode, the behaviour I noticed was that on launch even without triggering a build the fmw files would be created under DD. I mainly want to keep the generator usable while we might to `XCHammer`, also learning how this works is going to help me to better integrate things in `XCHammer`+`xcbuildkit`.

tests/macos/xcodeproj/Test-Target-With-Test-Host-Project.xcodeproj/bazelinstallers/install.sh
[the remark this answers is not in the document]
thiagohmcruz: Yeah, I don't know exactly what part of the stack is creating this. Noticed only that on launch `XCBBuildService` is the process responsible for creating the fmw files on disk under DD (without triggering a build in Xcode). It does that most likely based on the project configuration and declared targets/deps.

>When the product type was a .framework we wouldn't install - this used to be good enough in Xcode 13

FWIW the old behaviour should continue to exist, the `QUERY` we had for static fmws should continue to return whatever it was returning before, I just added a conditional to not exit early and instead continue and run `index-import` at the end, we still need that to run even if `input_options` is empty. As long as DD is cleaned up (i.e. not malformed fmw directories in there) the indexing invocations should respect the VFS and compile succesfully.

PR #668, opened 2023-03-15
Adding 15.0 as a version of the test host app to generate

mccorkill1:
Adding 15.0 to the list of DEFAULT_MINIMUM_OS_VERSION_LIST in test_host_app.bzl. This will allow `ios_unit_test` targets with `minimum_os_version = 15.0` using `test_host = True` to work.


mattrobmattrob (approved):

jerrymarino (approved):
Thanks for the PR 👍

PR #673, opened 2023-04-03
Update macro arguments to specify rules_ios modified arguments

luispadron:
A follow-up to #670 where we talked about moving the rules_ios modified kwargs to explicit arguments in the macros. Additionally, generates missing docs & updates outdated documentation.

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -18,7 +18,6 @@ _IOS_APPLICATION_KWARGS = [
     "executable_name",
     "minimum_os_version",
     "test_host",
-    "families",
     "entitlements",
     "entitlements_validation",
     "extensions",
@@ -39,13 +38,23 @@ _IOS_APPLICATION_KWARGS = [
     "app_clips",
 ]
 
-def ios_application(name, apple_library = apple_library, infoplists_by_build_setting = {}, **kwargs):
+def ios_application(
+        name,
+        families = ["iphone", "ipad"],
+        apple_library = apple_library,
+        infoplists = [],
+        infoplists_by_build_setting = {},
+        xcconfig = {},
+        xcconfig_by_build_setting = {},
+        **kwargs):
     """
     Builds and packages an iOS application.
 
     Args:
         name: The name of the iOS application.
+        families: A list of iOS device families the target supports.
         apple_library: The macro used to package sources into a library.
+        infoplists: A list of Info.plist files to be merged into the iOS app.
         infoplists_by_build_setting: A dictionary of infoplists grouped by bazel build setting.
 
                                      Each value is applied if the respective bazel build setting
@@ -53,6 +62,14 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
 
                                      If '//conditions:default' is not set the value in 'infoplists'
                                      is set as default.
+        xcconfig: A dictionary of xcconfigs to be applied to the iOS app by default.
+        xcconfig_by_build_setting: A dictionary of xcconfigs grouped by bazel build setting.
+
+                                   Each value is applied if the respective bazel build setting
+                                   is resolved during the analysis phase.
+
+                                   If '//conditions:default' is not set the value in 'xcconfig'
+                                   is set as default.
         **kwargs: Arguments passed to the apple_library and ios_application rules as appropriate.
     """
 
@@ -67,7 +84,6 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
     )
 
     application_kwargs["launch_storyboard"] = application_kwargs.pop("launch_storyboard", library.launch_screen_storyboard_name)
-    application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
 
     # Setup force loading here - need to process deps and libs
     force_load_name = name + ".force_load_direct_deps"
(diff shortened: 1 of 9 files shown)


rules/app.bzl
```diff
@@ -67,7 +83,6 @@ def ios_application(name, apple_library = apple_library, infoplists_by_build_set
     )
 
     application_kwargs["launch_storyboard"] = application_kwargs.pop("launch_storyboard", library.launch_screen_storyboard_name)
-    application_kwargs["families"] = application_kwargs.pop("families", ["iphone", "ipad"])
```
luispadron, line 70: Not sure why we decided to default this, but agree it's probably best to not default here and instead make sure the user provides it.

This will be a breaking change though so not sure if its worth it? Let me know

thiagohmcruz: I also don't know why we default-ed here, since this is required on the `rules_apple` side I think it makes sense to default to empty and allow consumers to configure this properly

I'll tag @justinseanmartin and @congt here since they might have context on older versions of this file and raise any concerns with this breaking change in favour of a more reasonable default

That said, I think it's a small change on users BUILD files, so I'd only consider not doing it if there's a major concern here

rules/framework.bzl
```diff
@@ -40,27 +40,48 @@ _APPLE_FRAMEWORK_PACKAGING_KWARGS = [
...
         **kwargs: Arguments passed to the apple_library and apple_framework_packaging rules as appropriate.
     """
     framework_packaging_kwargs = {arg: kwargs.pop(arg) for arg in _APPLE_FRAMEWORK_PACKAGING_KWARGS if arg in kwargs}
     kwargs["enable_framework_vfs"] = kwargs.pop("enable_framework_vfs", True)
 
-    infoplists_by_build_setting = kwargs.pop("infoplists_by_build_setting", {})
-    default_infoplists = kwargs.pop("infoplists", [])
-    infoplists = None
-    if len(infoplists_by_build_setting.values()) > 0 or len(default_infoplists) > 0:
+    merged_infoplists = None
```
thiagohmcruz, line 77: `merged_infoplists` is not being set anywhere IIUC?

thiagohmcruz (approved):
I think it's a good idea to standardize these, also thx for adding the new docs! I'd get feedback from others too re the minor breaking change here but AFAICT it's not major.

rules/framework.bzl
luispadron:
```
    apple_framework_packaging(
        name = name,
        framework_name = library.namespace,
        infoplists = merged_infoplists,
```

Gets set on the framework

thiagohmcruz: yep but if I followed the code paths correctly it's set to `None` above and then passed as `None` to the fmw right? I don't see it receive any non-`None` value anywhere

rules/app.bzl
```diff
@@ -39,20 +39,36 @@ _IOS_APPLICATION_KWARGS = [
...
+        xcconfig = {},
+        xcconfig_by_build_setting = {},
+        **kwargs):
     """
     Builds and packages an iOS application.
 
     Args:
         name: The name of the iOS application.
         apple_library: The macro used to package sources into a library.
+        infoplists: A list of Info.plist files to be merged into the app clip.
```
mattrobmattrob, line 56:
```suggestion
        infoplists: A list of Info.plist files to be merged into the iOS app.
```

rules/framework.bzl
luispadron: 🤦🏼 ah right, i've updated it to get set as the result of the `process_infoplist`!

rules/extension.bzl
```diff
@@ -14,22 +20,29 @@ def ios_extension(name, infoplists_by_build_setting = {}, **kwargs):
 
     Args:
         name: The name of the iOS extension.
+        infoplists: A list of Info.plist files to be merged into the app clip.
```
mattrobmattrob, line 23:
```suggestion
        infoplists: A list of Info.plist files to be merged into the extension.
```

rules/extension.bzl
```diff
@@ -14,22 +20,29 @@ def ios_extension(name, infoplists_by_build_setting = {}, **kwargs):
...
         infoplists_by_build_setting: A dictionary of infoplists grouped by bazel build setting.
 
                                      Each value is applied if the respective bazel build setting
                                      is resolved during the analysis phase.
 
                                      If '//conditions:default' is not set the value in 'infoplists'
                                      is set as default.
+        xcconfig: A dictionary of xcconfigs to be applied to the app clip by default.
```
mattrobmattrob, line 31:
```suggestion
        xcconfig: A dictionary of xcconfigs to be applied to the extension by default.
```

rules/framework.bzl
```diff
@@ -40,27 +40,48 @@ _APPLE_FRAMEWORK_PACKAGING_KWARGS = [
...
+        infoplists_by_build_setting = {},
+        xcconfig = {},
+        xcconfig_by_build_setting = {},
+        **kwargs):
     """Builds and packages an Apple framework.
 
     Args:
         name: The name of the framework.
         apple_library: The macro used to package sources into a library.
+        infoplists: A list of Info.plist files to be merged into the app clip.
```
mattrobmattrob, line 56:
```suggestion
        infoplists: A list of Info.plist files to be merged into the framework.
```

rules/framework.bzl
```diff
@@ -40,27 +40,48 @@ _APPLE_FRAMEWORK_PACKAGING_KWARGS = [
...
+        infoplists_by_build_setting: A dictionary of infoplists grouped by bazel build setting.
+
+                                     Each value is applied if the respective bazel build setting
+                                     is resolved during the analysis phase.
+
+                                     If '//conditions:default' is not set the value in 'infoplists'
+                                     is set as default.
+        xcconfig: A dictionary of xcconfigs to be applied to the app clip by default.
```
mattrobmattrob, line 64:
```suggestion
        xcconfig: A dictionary of xcconfigs to be applied to the framework by default.
```

mattrobmattrob (approved):
I'd prefer we leave the default `families` value but make it show up in the docs.

rules/app.bzl
thiagohmcruz: I'm ok with not reverting it then 👍, +1 to making it visible in the docs

thiagohmcruz: @mattrobmattrob so just to confirm the preference would be to keep the two families as default everywhere or only for apps?

mattrobmattrob: I suppose I'd prefer the default (assuming we keep it) is the same anywhere it's applicable as a param to the underlying rules.

luispadron: I'll update to move that to the macro args so its documented in all places it was previously used 👍🏼

luispadron:
@mattrobmattrob @thiagohmcruz thanks for the reviews! I've updated this to default `families` to `iphone, ipad` and added them to the macros args so they're documented.

mattrobmattrob (approved):

PR #376, opened 2021-12-10
Add a remote cache

thii:
Test running is taking almost an hour each to run. This adds a remote
cache option to speed it up, using the unauthenticate remote cache from
BuildBuddy.

PR #676, opened 2023-04-03
Add remote cache for CI jobs

mattrobmattrob:
New version of https://github.com/bazel-ios/rules_ios/pull/376 to use BuildBuddy remote cache for CI jobs. Only enabled for CI jobs at this point vs. in the `.bazelrc` to avoid being prescriptive for dev workflows.

### DIFF
--- a/.github/workflows/xcode_select.sh
+++ b/.github/workflows/xcode_select.sh
@@ -19,3 +19,7 @@ echo "build --xcode_version_config=//:host_xcodes" >> user.bazelrc
 
 # `deleted_packages` is needed below in order to override the value of the .bazelrc file
 echo "build:ios --apple_platform_type=ios --deleted_packages=''" >> user.bazelrc
+
+# Remote cache
+echo "build --remote_cache=grpcs://remote.buildbuddy.io" >> user.bazelrc
+echo "build --remote_timeout=3600" >> user.bazelrc
--- a/.github/workflows/xcode_select.sh
+++ b/.github/workflows/xcode_select.sh
@@ -23,3 +23,4 @@ echo "build:ios --apple_platform_type=ios --deleted_packages=''" >> user.bazelrc
 # Remote cache
 echo "build --remote_cache=grpcs://remote.buildbuddy.io" >> user.bazelrc
 echo "build --remote_timeout=3600" >> user.bazelrc
+echo "build --modify_execution_info=^(BitcodeSymbolsCopy|BundleApp|BundleTreeApp|DsymDwarf|DsymLipo|GenerateAppleSymbolsFile|ObjcBinarySymbolStrip|CppLink|ObjcLink|ProcessAndSign|SignBinary|SwiftArchive|SwiftStdlibCopy)$=+no-remote,^(BundleResources|ImportedDynamicFrameworkProcessor)$=+no-remote-exec" >> user.bazelrc


mattrobmattrob:
We'd probably ideally also add a mnemonic exclusion for the larger packaging operations but will let that change stand on its own for now.

jerrymarino:
I'd suggest to keep rules_ios as simple CI and remain neutral toward vendors. This said, for a reliable batteries included vendor cache - Flare's CDN is likely my first choice due it's real world perf and edge support.

Have you tried to setup the github action disk cache as a quick way to speed up?

mattrobmattrob:
> I'd suggest to keep rules_ios as simple CI and remain neutral toward vendors. This said, for a reliable batteries included vendor cache - Flare's CDN is likely my first choice due it's real world perf and edge support.
> 
> 
> 
> Have you tried to setup the github action disk cache as a quick way to speed up?

BuildBuddy has a similar CDN approach at this point from what I know. I'd bet performance is comparable.

And no, haven't tried a disk cache.

mattrobmattrob:
Mostly trying to replicate the rules_xcodeproj CI which is so incredibly pleasant to use. rules_ios continues to slow down so would be great to make an effort to improve it for contributors.

thiagohmcruz (approved):
The changes make sense to me as I don't see any major downside here. Even if there's issues or if we start having any kind of issues with the vendor (unlikely) we can easily revert or pivot to a different provider.

That said, If we can make the DX here better today I don't see why we wouldn't start with something that works and iterate from there.

Thx for raising this and taking the PR! 

Nit: Should we add a blurb in the `README` telling folks where this is configured and maybe simple steps to disable/debug if they need more visibility into non-hermetic stuff? Maybe there isn't anything useful to add to a `README` and I'm over complicating this, so take it as a nit.

zachgrayio:
Thanks for the honorable mention @jerrymarino .

>BuildBuddy has a similar CDN approach at this point from what I know. I'd bet performance is comparable.

depending on geo we have seen a ~5x build speed increase over this public endpoint in the PR here (as measured in end to end build time of real world ios apps), and this space is an area of active investment for us so this number will continue to evolve.

also one other lesser-known fact, we are offering free access to the Bitrise remote caching and CDN products for OSS projects, this isn't well documented yet but details are forthcoming as well as a more frictionless onboarding to the platform. happy to drop a PR here on that also for comparison if there's interest

mattrobmattrob:
Couldn't find any reference on https://flare.build/ for the setup to use with OSS projects. Happy to use whatever.

That being said, I don't think performance is going to matter all that much in this case given the sizes of the examples. We just need _some_ remote cache.

mattrobmattrob:
What would you prefer, @jerrymarino?

zachgrayio:
> Couldn't find any reference on https://flare.build/ for the setup to use with OSS projects. Happy to use whatever.
> 
> That being said, I don't think performance is going to matter all that much in this case given the sizes of the examples. We just need _some_ remote cache.

Any new details will be over at [bitrise](https://bitrise.io/why/features/mobile-build-caching-for-better-build-test-performance) marketing site, at the blog, and also the [tech docs](https://devcenter.bitrise.io/en/builds/caching.html) at some point of course. Probably a blog post first.

that said lets not hold this up either

jerrymarino:
@zachgrayio thanks for the note - yeah I'd be curious to know what the delta difference is for this solution. 

My largest concern for something like this repo is we'd muddy waters and prospects by demonstrating an affinity towards a particular vendor - little known is many of the vendors bring us contributors, contributions, and users to rules_ios. Perhaps we can have an evidence based approach to the CI / really evaluate them - but defining an evidence based to selecting one is not trivial. 

I also the idea to showcase the community based solutions here, or if there is an alternative e.g. doing something simple like using the github action cache. Is the CI time really that bad in general? Perhaps we can start gutting some of the shards if concurrency is the issue and backing up PRs.

mattrobmattrob:
I'd be remiss if I didn't mention that it's pretty odd to not go with a provider that offers a free solution for OSS projects, is a fantastic open source contributor as a company, has a large group of users in the community, and has a vested interest in rules_ios/Bazel being successful. Whatever that vendor is, BuildBuddy or otherwise.

Seems like any concern about pointing to a specific vendor is cutting off one's nose to spite one's face here. The speed ups are essentially free until someone can do the work you mention, which we should also do!

luispadron (approved):

jszumski:
> @mattrobmattrob: Seems like any concern about pointing to a specific vendor is cutting off one's nose to spite one's face here. The speed ups are essentially free until someone can do the work you mention, which we should also do!

+1 to this and it's very easy to pivot away if our needs change in the future

> @jerrymarino: if there is an alternative e.g. doing something simple like using the github action cache

I'm not sure it can get much simpler than this 3 line diff.  Why is GitHub as the vendor any different than BuildBuddy or Flare?

thiagohmcruz:
IIUC from the discussion above we have a few alternatives here to address @jerrymarino's concerns:

1. Pick a vendor via a more methodical approach
2. Consider non-remote cache alternatives (e.g. GH actions cache)
3. Consider community based remote cache solutions (which probably means building something from scratch?)

I'm personally ok if anyone wants to invest on any of the above and have a more "evidence based" way to pick the best solution here. 

My point still holds though: I don't see why we wouldn't land this as is in this PR to improve DX for maintainers today instead of "one day"

We're not signing any contracts here, we can revert this anytime or even better *iterate* on it using this initial change as a baseline to do any of the items listed above if someone really wants to go that route.

mattrobmattrob:
I don't want to circumvent you, @jerrymarino, but the other maintainers seem supportive of this change.

Are you comfortable merging this and seeing how it goes?

jerrymarino:
> Are you comfortable merging this and seeing how it goes?

I'd like to know how this compares against Flare, EngFlow, or other vendors before selecting one so we can understand how to make the best contributor experience.

In the current state of it, I also don't think we've resolved the issue of tieing our remote cache workflows to an opaque remote cache and giving a preferential treatment towards one of them without evidence or comparing them.

It also isn't a huge fire to land this - so I'd like to see these issues resolved but won't block you

mattrobmattrob:
> @jerrymarino: I'd like to know how this compares against Flare, EngFlow, or other vendors before selecting one so we can understand how to make the best contributor experience.

I don't think this should block so I'll land this for now. If it doesn't improve things or causes issues, we can refactor our approach/remove it all together. Happy for someone to contribute a comparison if they're so inclined. I also can't find any docs for a free-to-use-for-OSS solution on any of the other providers' websites but I could be missing that.

> @jerrymarino: In the current state of it, I also don't think we've resolved the issue of tieing our remote cache workflows to an opaque remote cache and giving a preferential treatment towards one of them without evidence or comparing them.

Happy to add a comment stating that we didn't explicitly do performance testing if that'd make you feel better about it. I don't think being preferential (even if incidentally so) is a bad thing. If it helps our use case, we should use it.

PR #678, opened 2023-04-05
Minor vfs perf improvements

thiagohmcruz:
Running some tests using `--starlark_cpu_profile` in one big app we have and noticed some hot spots we can easily improve here. Nothing major but still a good optimization to make IMO.

The overall improvement at the `_framework_vfs_overlay_impl` level is from `109.82s` to down to `77.82s`:

before:
Screen Shot 2023-04-06 at 10 17 43 AM

after:
Screen Shot 2023-04-06 at 10 17 52 AM

I was surprised with how much time is spent on `append/extend/len` operations but I guess there's not much we can do here without revisiting the algo as a whole and see if there's a better way to go about generating the same outputs.

It's definitely weird that removing the macro `_get_external_contents` that was just concatenating strings has improved things and I ran the same "before vs after" profiles without making the change in this commit https://github.com/bazel-ios/rules_ios/pull/678/commits/908d03befb963b572849c48a8707eba63f5330d1 to see if it was indeed causing any impact. When I did that the improvement was not as a good, `_framework_vfs_overlay_impl` went from `109.82s` down to `92.12s`. I don't know if there's additional overhead in Starlark when macros are invoked but unless my interpretation is incorrect I think we should keep this commit.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -1,3 +1,14 @@
+"""framework_vfs_overlay impl
+
+Note on `external-contents` key set many times in this file:
+
+Internal to swift and clang - LLVM `VirtualFileSystem` object can
+serialize paths relative to the absolute path of the overlay. This
+requires the paths are relative to the overlay. While deriving the
+in-memory tree roots, it pre-pends the prefix of the `vfsoverlay` path
+to each of the entries.
+"""
+
 load("//rules:providers.bzl", "FrameworkInfo")
 load("//rules:features.bzl", "feature_names")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -91,10 +91,11 @@ def _get_public_framework_header(path):
 # and incrementality. For imported frameworks, there is additional search paths
 # enabled
 def _make_root(vfs_parent, target_triple, swiftmodules, root_dir, extra_search_paths, module_map, hdrs, private_hdrs):
-    vfs_prefix = _make_relative_prefix(len(vfs_parent.split("/")) - 1)
+    vfs_parent_len = len(vfs_parent.split("/")) - 1
+    vfs_prefix = _make_relative_prefix(vfs_parent_len)
     private_headers_contents = []
     headers_contents = []
-    vfs_prefix = _make_relative_prefix(len(vfs_parent.split("/")) - 1)
+    vfs_prefix = _make_relative_prefix(vfs_parent_len)
 
     if extra_search_paths:
         paths = []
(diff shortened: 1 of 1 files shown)


rules/framework/vfs_overlay.bzl
```diff
@@ -62,15 +65,16 @@ def _build_subtrees(paths, vfs_prefix):
         # Loop the _framework_ path and add each dir to the current tree.
         # Assume the last bit is a file then add it as a file
         idx = 0
+
+        parts_len = len(parts)
         for part in parts:
-            if idx == len(parts) - 1:
```
jerrymarino, line 66: 🥳

rules/framework/vfs_overlay.bzl
```diff
@@ -62,15 +65,16 @@ def _build_subtrees(paths, vfs_prefix):
...
-                ext_c = _get_external_contents(vfs_prefix, path_info.path)
+            if idx == parts_len - 1:
                 curr_subdirs.dict[part] = -1
-                curr_subdirs.json["contents"].append({"name": part, "type": "file", "external-contents": ext_c})
+                curr_subdirs.json["contents"].append({"name": part, "type": "file", "external-contents": vfs_prefix + path_info.path})
                 break
 
             # Lookup a value for the current subdirs, otherwise append
-            next_subdirs = curr_subdirs.dict.get(part, None)
+            next_subdirs = curr_subdirs.dict[part] if part in curr_subdirs.dict else None
```
luispadron, line 77: nit: think the `get` before was more clear

luispadron (approved):
Looks good, thanks for profiling. Not sure why `len` would be so slow (unless these aren't actual list types?). But regardless, think the the cleanup is good here

rules/framework/vfs_overlay.bzl
thiagohmcruz: this is also more performant 🤯, I don't remember by how much but `.get` seems to also cause more overhead than this other approach

thiagohmcruz: I think we only notice this in really big apps that call this many many times during the analysis phase

mattrobmattrob, line 77: My guess is that this operates the same as before but is just a bit harder to read so we can just leave it as is:

```suggestion
            next_subdirs = curr_subdirs.dict.get(part, None)
```

Starlark Go implementation:

https://github.com/google/starlark-go/blob/4b1e35fe22541876eb7aa2d666416d865d905028/starlark/library.go#L1193-L1207

luispadron: Intersting

> Python has to do more work for dict.get():
>
> get is an attribute, so Python has to look this up, and then bind the descriptor found to the dictionary instance.
() is a call, so the current frame has to be pushed on the stack, a call has to be made, then the frame has to be popped again from the stack to continue.
The [...] notation, used with a dict, doesn't require a separate attribute step or frame push and pop.

https://stackoverflow.com/questions/36566331/why-does-dict-getkey-run-slower-than-dictkey

mattrobmattrob (approved):
👏 

In a somewhat awful `n=1` simulation of before/after, here's what it looks like for us for the `_framework_vfs_overlay_impl`: 

|Situation|Time|Image|
|---|---|---|
|Before|4.38s|Screenshot 2023-04-06 at 8 25 50 AM
|After|2.38s|Screenshot 2023-04-06 at 8 37 20 AM|

Tested via `--starlark_cpu_profile=/tmp/starlark.pprof.gz` + `go run github.com/google/pprof@latest -http=: /tmp/starlark.pprof.gz`.

rules/framework/vfs_overlay.bzl
thiagohmcruz: yep, in Luis's comment, but I'm also with keeping it since there's evidence that it performs better

PR #680, opened 2023-04-06
Calculate `vfs_prefix` once per `make_vfsoverlay`

mattrobmattrob:


thiagohmcruz (approved):
I see a drop here using our app to test: `_framework_vfs_overlay_impl` goes from `77.82s` down to `73.02s`

So I think it's a good thing to land for sure. Thx!

mattrobmattrob:
Thanks for testing it, @thiagohmcruz!

luispadron (approved):
Loving all these perf PRs

jerrymarino:
@mattrobmattrob thanks for the PR 👍 . How much do ya'll save on the analysis phase with this one? There is definitely a lot of room for improvement still WRT analysis

mattrobmattrob:
> @mattrobmattrob thanks for the PR 👍 . How much do ya'll save on the analysis phase with this one? There is definitely a lot of room for improvement still WRT analysis

Much smaller than https://github.com/bazel-ios/rules_ios/pull/678 for me. I didn't take super exact measurements though just a run before and after to see if it went down.

PR #681, opened 2023-04-06
_min_os_version_or_none perf improvements

thiagohmcruz:
Similar to https://github.com/bazel-ios/rules_ios/pull/678, using `--starlark_cpu_profile` to find hot spots. Looking at `_apple_rule_transition_impl` now.

Before:
image

After:
image

### DIFF
--- a/rules/transition_support.bzl
+++ b/rules/transition_support.bzl
@@ -18,13 +18,13 @@ def _cpu_string(platform_type, settings):
     # If the cpu value has already been transformed to the correct value, we must not change it anymore.
     # Otherwise, we may build for the wrong arch.
     cpu_value = settings["//command_line_option:cpu"]
-    if (platform_type == "macos" and cpu_value.startswith("{}_".format(platform_type))) or cpu_value.startswith("{}_".format(platform_type)):
+    if (platform_type == "macos" and cpu_value.startswith("%s_" % platform_type)) or cpu_value.startswith("%s_" % platform_type):
         return cpu_value
 
     if platform_type == "ios":
         ios_cpus = settings["//command_line_option:ios_multi_cpus"]
         if ios_cpus:
-            return "ios_{}".format(ios_cpus[0])
+            return "ios_%s" % ios_cpus[0]
         if settings["//command_line_option:cpu"] == "darwin_arm64":
             return "ios_sim_arm64"
         else:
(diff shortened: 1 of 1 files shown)


luispadron (approved):

rules/transition_support.bzl
```diff
@@ -18,47 +18,46 @@ def _cpu_string(platform_type, settings):
...
+def _min_os_version_or_none(attr, attr_platforms, platform, attr_platform_type):
     if attr_platform_type != platform:
         return None
 
-    if hasattr(attr, "platforms"):
-        platforms = attr.platforms
-        value = platforms.get(platform)
-        return value
-    elif hasattr(attr, "minimum_os_version"):
+    if attr_platforms:
```
thiagohmcruz, line 54: This is already calculated earlier on the call site so passing it here since it's a private macro should be fine

mattrobmattrob (approved):

thiagohmcruz:
oops, unintended behaviour change here since CI is complaining, will push a fix soon

rules/transition_support.bzl
```diff
@@ -18,47 +18,46 @@ def _cpu_string(platform_type, settings):
...
     if attr_platform_type != platform:
         return None
 
-    if hasattr(attr, "platforms"):
-        platforms = attr.platforms
-        value = platforms.get(platform)
-        return value
-    elif hasattr(attr, "minimum_os_version"):
+    if attr_platforms != None:
+        return attr_platforms[platform] if platform in attr_platforms else None
```
mattrobmattrob, line 55: Would this previous crash/error out since it used `.get(platform)` without a default? Is this a change in functionality?

thiagohmcruz: So if `attr.platforms` was `{}` it would just try to `.get(platform)` and return `None`, `attr.platforms` being `{}` does happen during the tests: `bazelisk test --local_test_jobs=1 -- //... -//tests/ios/...`

With this new code the same possibilities continue to run the same way but without needing to `hasattr(attr, "platforms")` which was one hot spots

- `attr_platforms` is `None` nothing to do
- `attr_platforms` is `{}` it tries to `attr_platforms[platform]` and returns `None` if it's not there
- `attr_platforms` has info it uses it

I think this is the same behaviour and not a change in behaviour but without the extra `hasattr`. Does that make sense or am I missing something?

mattrobmattrob: 🤦 You're right. Didn't realize that `unknown dict.get(key, default=None)` has a `default=None`.

thiagohmcruz: Cool, thx for calling it out though, it's a good way for me to also triple check if I'm not changing behaviour here

Issue #563, opened 2022-09-16
Swift/Objc helper / demonstrate how to do hermetic snapshot loading


jerrymarino:
Provide some minimal code that demonstrates structuring and execute iOS snapshot tests with Bazel and rules_ios. The snapshot tests can be structured something like this - where the `GoldenImages.bundle` - allows easily loading of the images - or perhaps it's a symlink to the real bundle to avoid duping images.

```
My.app/Plugins/MySnapshot.xctest/GoldenImages.bundle`
```

Finally add an example to handle with "recording" snapshots in the cloud
```
bazel run //tests --record 
```
If people are checking the snaps into an external file store - illustrate how to handle this end to end: e.g. downloading the snapshot images into `bazel-out` - and assimilating them back into the "external" repo. 

Getting a basic example / test case should separate out opinions about storing images.

jerrymarino:
cc @dostrander I was thinking we could add something similar to the built in C++ snapshot lib: https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.h#L26 but for `NSBundle`. We might need extra consideration for how the remote test runner works with RBE but from an API perspective the C++ code might be a good starting point

PR #682, opened 2023-04-11
Basic skeleton for snapshot, quirks with bazel_tools

jerrymarino:
Plumb in the basics of snapshot library support. Please vet if we can reuse Bazel's lib here per #563

@dostrander JFYI I've thrown together this quick jib to get this plumbed in as a library ya'll can import in the project


rules/test/snapshot/BundleBridge.mm
```diff
@@ -0,0 +1,14 @@
+#import "BundleBridge.h"
+
+@implementation NSBundle (BazelBundleBridge)
+
+- (NSString *)bazelRunfilePathForResource:(NSString *)resource
+{
+    // TODO Determine:
+    // Does the existing C++ code solving our problem, if so that is awesome and
+    // potentially  more robust than a simple implementation that we do for
```
jerrymarino, line 9: On this line of code, we could confirm if this code is setup to work E2E 👍 
https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.h#L28

tests/ios/app/snapshot_test.m
```diff
@@ -0,0 +1,17 @@
...
+
+@end
+
+@implementation EmptyTests
+
+- (void)testFailureEndToEnd
+{
+    // TODO: update for snapshots trees
+    // For now, we load the actual test bundle
+    XCTAssertNonNil([[NSBundle mainBundle] bazelRunfilePathForResource:@"build_bazel_rules_ios/tests/ios/app/TestSnapshots"]);
```
jerrymarino, line 14: iOS developers may not find it idiomatic to pass in the path of their BUILD file in like this. Perhaps we can hack this in somehow so they can call:
```
    XCTAssertNonNil([[NSBundle mainBundle] bazelRunfilePathForResource:@"ReferenceImages"]);
```

Otherwise we can add an API like this:
```
-[NSBundle testReferenceImages]
```

rules/test/snapshot/BazelRunfileBundleBridge.mm
```diff
@@ -0,0 +1,28 @@
+#import "BazelRunfileBundleBridge.h"
+#include "tools/cpp/runfiles/runfiles.h"
+#include <string>
+
+@implementation NSBundle (BazelRunfileBundleBridge)
```
dostrander, line 5: This probably shouldn't be an `NSBundle` extension as this has nothing to do with `NSBundle`. It probably should just be it's own class method on an object like `BZLRunfile` or just a regular function like `static inline NSString* BZLRunfileForPath(NSString *path)`

jerrymarino: ATM it's not really tied to it, but otherwise - how would we'd feed a value to the `NSBundle` to get the prefix w/o the bundle layer?
> iOS developers may not find it idiomatic to pass in the path of their BUILD file in like this. Perhaps we can hack this in somehow so they can call:
>  ```
>   XCTAssertNonNil([[NSBundle mainBundle] bazelRunfilePathForResource:@"ReferenceImages"]);
> ```

https://github.com/bazel-ios/rules_ios/pull/682#discussion_r1164807887

IME the code needs that relative path, but perhaps we can iterate on it. In general - do you think we should reasonably provide the relative path to all `BUILD` files?

jerrymarino: It looks like this code:
```
        std::string path = runfiles->Rlocation(resource.UTF8String);
```
needs to be prefixed with the build file path. 

In practice, how this would apply for the iOS code:

### Handled automatically by rules_ios  - loading the reference images
```
    NSString *dir = [[NSBundle mainBundle] bazelRunfilePathForResource:@"ReferenceImages"];
```

### Handled by custom snapshot code - loading the reference images
```
    NSString *dir = [[NSBundle mainBundle] bazelRunfilePathForResource:@"squareup/Veticals/Foo/Tests/ReferenceImages"];
```

jerrymarino: Checkout in the PR of how it is able to load a runfile:
https://github.com/bazel-ios/rules_ios/pull/682/files#diff-67c69506235c47f27e2cb75c8931b0389995d9211a38ee530b8053b3f6f80ef7R14

jerrymarino: Probably the latter is too specialized - perhaps better to have `relativeRunfilePath`

jerrymarino: I'm going to leave this as basic as possible for now, we can determine how to make it find relative paths for the above instance in @steve-sun-me 's change set

steve-sun-me: Yeah I am fine with that. We can alter as we go.

rules/test.bzl
```diff
@@ -160,3 +160,14 @@ def ios_ui_test(name, apple_library = apple_library, **kwargs):
...
+def ios_unit_snapshot_test(name, apple_library = apple_library, **kwargs):
+    """
+    Builds and packages iOS Unit Snapshot Tests.
+
+    Args:
+        name: The name of the UI test.
+        apple_library: The macro used to package sources into a library.
+        **kwargs: Arguments passed to the apple_library and ios_ui_test rules as appropriate.
+    """
+    _ios_test(name, rules_apple_ios_ui_test, rules_apple_ios_ui_test_suite, apple_library, internal_test_deps = ["@bazel_tools//tools/cpp/runfiles"], **kwargs)
```
steve-sun-me, line 173: Should this be `rules_apple_ios_unit_test`?

jerrymarino: Good catch 👍

rules/test.bzl
```diff
@@ -160,3 +160,14 @@ def ios_ui_test(name, apple_library = apple_library, **kwargs):
...
     _ios_test(name, rules_apple_ios_ui_test, rules_apple_ios_ui_test_suite, apple_library, **kwargs)
+
+def ios_unit_snapshot_test(name, apple_library = apple_library, **kwargs):
+    """
+    Builds and packages iOS Unit Snapshot Tests.
+
+    Args:
+        name: The name of the UI test.
+        apple_library: The macro used to package sources into a library.
+        **kwargs: Arguments passed to the apple_library and ios_ui_test rules as appropriate.
```
steve-sun-me, line 171: `ios_ui_test` -> `ios_unit_test`

rules/test.bzl
```diff
@@ -160,3 +160,14 @@ def ios_ui_test(name, apple_library = apple_library, **kwargs):
     if not kwargs.get("test_host", None):
         fail("test_host is required for ios_ui_test.")
     _ios_test(name, rules_apple_ios_ui_test, rules_apple_ios_ui_test_suite, apple_library, **kwargs)
+
+def ios_unit_snapshot_test(name, apple_library = apple_library, **kwargs):
```
steve-sun-me, line 164: I was thinking if we can just reuse `ios_unit_test` function instead of creating a new one

steve-sun-me (approved):

rules/test.bzl
jerrymarino: I've got some features conventions for snapshots that I'm hoping to inject here longer term; today we tag a `TestRunner` as as a `snapshot` which can implicate various side effects; ATM it also has to be like this for the lines added in this PR due to other issues but this sets us in the right direction

rules/test/snapshot/BazelRunfileBridge.mm
```diff
@@ -0,0 +1,23 @@
+#import "BazelRunfileBridge.h"
+#include "tools/cpp/runfiles/runfiles.h"
+#include <string>
+
+using namespace bazel::tools::cpp::runfiles;
+
+NSString *BazelRunfilePathForResource(NSString *resource)
+{
+    // TODO: confirm if Bazel runfile support will work E2E 👍
```
dostrander, line 9: There's still todos in here?

jerrymarino: Yeah this PR was listed as scaffolding to get a function call to the referenced library wired up and ensure it's buildable and callable at a basic level. Weather or not the lib is a complete answer for snapshots loading needs more testing and consideration and you can atleast more easily try it out. These are left as TODOs here that we can remove once https://github.com/bazel-ios/rules_ios/issues/563 is resolved.

PR #683, opened 2023-04-12
Loading XCHammer is opt-in

thiagohmcruz:
Prevents unnecessarily downloading XCHammer dependencies if there's no intent of using it. If they do want to use it they just need to invoke this new macro in their `WORKSPACE` file.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,9 +3,12 @@ workspace(name = "build_bazel_rules_ios")
 load(
     "//rules:repositories.bzl",
     "rules_ios_dependencies",
+    "rules_ios_xchammer_dependencies",
 )
 
 rules_ios_dependencies()
+
+rules_ios_xchammer_dependencies()
 
 load(
     "@build_bazel_rules_apple//apple:repositories.bzl",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,12 +3,9 @@ workspace(name = "build_bazel_rules_ios")
 load(
     "//rules:repositories.bzl",
     "rules_ios_dependencies",
-    "rules_ios_xchammer_dependencies",
 )
 
-rules_ios_dependencies()
-
-rules_ios_xchammer_dependencies()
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -3,10 +3,13 @@ workspace(name = "build_bazel_rules_ios")
 load(
     "//rules:repositories.bzl",
     "rules_ios_dependencies",
+    "rules_ios_xchammer_dependencies",
 )
 
 rules_ios_dependencies()
```
mattrobmattrob, line 9: What do you think about a parameter to this so the discoverability is a bit better? Something like `include_xchammer = True` or similar?

thiagohmcruz: I considered something like that but I've never seen this pattern being used in other rules sets? I could be wrong and it's a common thing to do 

Alternative is to update the `README` and mention this?

mattrobmattrob: `rules_xcodeproj` has a pattern similar (but for a admittedly different purpose):

https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/3a8f83960c11dd9ad5e1cd5e45c53510c32cf3a0/xcodeproj/repositories.bzl#L86-L89

Same with `rules_js`, sorta:

https://github.com/aspect-build/rules_js/blob/1fcd70fc6dee2f5b4050f151846010cce4e013c3/npm/npm_import.bzl#L65-L73

Nice for it to be a single API, IMO. Probably still worth mentioning in the README, the error if it's not present, and/or the xcodeproj macro's attribute to enable xchammer support.

justinseanmartin (approved):
I like the idea of adding it as an optional param that defaults to false.

thiagohmcruz:
@mattrobmattrob @justinseanmartin added the flag and updated the `README` to be more precise and mention all options users have to generate Xcode projects. Let me know what you think.

mattrobmattrob (approved):

PR #688, opened 2023-04-14
Fix the comments about angle bracket / double quote imports

qyang-nj:
I'm pretty sure the comments were reversed.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -992,7 +992,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         lib_names.append(swift_libname)
 
         if objc_sources or cpp_sources:
-            # Add generated swift header to header maps for angle bracket imports
+            # Add generated swift header to header maps for double quote imports
             swift_doublequote_hmap_name = name + "_swift_doublequote_hmap"
             headermap(
                 name = swift_doublequote_hmap_name,
@@ -1005,7 +1005,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             private_dep_names.append(swift_doublequote_hmap_name)
             _append_headermap_copts(swift_doublequote_hmap_name, "-iquote", additional_objc_copts, additional_swift_copts, additional_cc_copts)
 
-            # Add generated swift header to header maps for double quote imports
+            # Add generated swift header to header maps for angle bracket imports
             swift_angle_bracket_hmap_name = name + "_swift_angle_bracket_hmap"
             headermap(
                 name = swift_angle_bracket_hmap_name,


ob (approved):
thanks for the fix

thiagohmcruz (approved):

rules/library.bzl
```diff
@@ -992,7 +992,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
         lib_names.append(swift_libname)
 
         if objc_sources or cpp_sources:
-            # Add generated swift header to header maps for angle bracket imports
+            # Add generated swift header to header maps for double quote imports
             swift_doublequote_hmap_name = name + "_swift_doublequote_hmap"
```
jerrymarino, line 996: IME we might be able to drop this; given if `-I` includes the same headers. In other-words the comment `double quote` might not be on point if a `-I` include would allow you to quote import a header? Having "doublequote hmap"'s never added up to me

rules/library.bzl
```diff
@@ -1005,7 +1005,7 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
             private_dep_names.append(swift_doublequote_hmap_name)
             _append_headermap_copts(swift_doublequote_hmap_name, "-iquote", additional_objc_copts, additional_swift_copts, additional_cc_copts)
 
-            # Add generated swift header to header maps for double quote imports
+            # Add generated swift header to header maps for angle bracket imports
```
jerrymarino, line 1008: AFAIK this is `double quote` *and* `angle bracket` - not sure if there is some quirk that required them both!

jerrymarino:
```               ^
tests/ios/unit-test/test-imports-app/test.swift:2:18: error: failed to import bridging header 'tests/ios/unit-test/test-imports-app/TestImports-App-Bridging-Header.h'
@testable import TestImports_App
                 ^
Target //tests/ios/unit-test/test-imports-app:TestImports-Unit-Tests failed to build
INFO: Elapsed time: 9.364s, Critical Path: 6.92s
INFO: 27 processes: 10 remote cache hit, 17 internal.
```
Can't say I've seen this CI error before 🤔

rules/library.bzl
qyang-nj: Totally agree. I don't think we need those separate "double quote" imports as well. I can delete them in a separate PR.

qyang-nj:
Not sure if there is anything I need to do to get this merged. I don't think my change can cause that CI error. Also it seems I don't have permission to rerun the job.

thiagohmcruz:
Whatever is going on on CI shouldn't be related to just updated comments to will merge this

rules/library.bzl
jerrymarino: Cool - thanks again for the PR, I ended up testing this out a bit this afternoon after I saw it and pushed it here: https://github.com/bazel-ios/rules_ios/pull/689!

PR #694, opened 2023-04-18
Update generate_release_notes.sh

luispadron:
Was using `bazelbuild` instead of `bazel-ios`


luispadron:
Closing for https://github.com/bazel-ios/rules_ios/pull/695

PR #696, opened 2023-04-19
Enable Runfiles deps for all test rules

steve-sun-me:
We introduced `internal_test_deps` in this [PR](https://github.com/bazel-ios/rules_ios/pull/682) for snapshot tests. This enable us to have more freedom to control whether we want `internal_test_deps` in our test rules

### DIFF
--- a/docs/test_doc.md
+++ b/docs/test_doc.md
@@ -39,6 +39,7 @@ ios_unit_snapshot_test(<a href="#ios_unit_snapshot_test-name">name</a>, <a href=
 | :------------- | :------------- | :------------- |
 | <a id="ios_unit_snapshot_test-name"></a>name |  The name of the UI test.   |  none |
 | <a id="ios_unit_snapshot_test-apple_library"></a>apple_library |  The macro used to package sources into a library.   |  <code>&lt;function apple_library&gt;</code> |
+| <a id="ios_unit_snapshot_test-include_internal_test_deps"></a>include_internal_test_deps |  Arguments passed to the rules_ios to include internal_test_deps.   |  none |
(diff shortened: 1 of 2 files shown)


mattrobmattrob (approved):
Curious if there's any reason to not make this ~a separate macro or~ some internal implementation for you all vs. in `rules_ios`?

This approach seems reasonable if it's going to be in `rules_ios` but just wanted to ask. 👍

steve-sun-me:
> Curious if there's any reason to not make this some internal implementation for you all vs. in rules_ios?

Yeah agreed here as well. Personally feel like this can just be done in the internal implementation.

@jerrymarino @thiagohmcruz curious about your thoughts?

thiagohmcruz:
> > Curious if there's any reason to not make this some internal implementation for you all vs. in rules_ios?
> 
> Yeah agreed here as well. Personally feel like this can just be done in the internal implementation. However, I think we want show how we can get runfiles etc. @jerrymarino @thiagohmcruz curious about your thoughts?

IIUC this is related to https://github.com/bazel-ios/rules_ios/issues/563, but looking at the diff here in this PR specifically I agree with @mattrobmattrob point, why not just conditionally add to the internal `deps` of tests instead of here in `rules_ios`?

If we want to exercise this wouldn't it be better to keep `test.bzl` untouched and instead add a test and `tests/` with whatever setup we want to exercise with this?

steve-sun-me:
Sounds good! Going to close the PR and keep it as internal implementation. Thanks @mattrobmattrob @thiagohmcruz

mattrobmattrob:
>If we want to exercise this wouldn't it be better to keep `test.bzl` untouched and instead add a test under tests/ with whatever setup we want to exercise with this?

And I agree with this. Helpful to have coverage of this use case for you all WRT `rules_ios` changes. Just funny to have this branching for a conditional dep (or the dep at all) in the public interface/macros to me. 👍

PR #699, opened 2023-04-20
Xcode 14.3 support on Bazel 5

thiagohmcruz:


thiagohmcruz:
Bumping to Bazel `5.4.1` is enough for this, not toolchain override needed as the fix is already in there: https://github.com/bazelbuild/bazel/releases/tag/5.4.1

PR #703, opened 2023-05-03
Add support for bzlmod

luispadron:
Adds initial support for bzlmod to rules_ios.

Major changes:

- Update versions in `repositories.bzl` to use minimum supported bzlmod version
- Add `MODULE.bazel` and required files to enable bzlmod by default in Bazel 6+
- Enable bzlmod by default in the repository

Required workarounds:

- Disable bzlmod when running legacy .xcodeproj tests. This rule collects workspace name in an aspect which now resolves to simply `_main` in bzlmod. This doesnt seem to cause any actual issues with the generated project besides the repository being named `_main`. Proper fix is to have this use runfiles but I do not have enough context to make that change. 

Things to follow-up on:

- Add bzlmod support for stardoc
  - This requires some more changes to bzl_library usage and update to Stardoc version, will do this in a separate PR before publishing.
- Publish to Bazel Central Registry (and introduce files to required to do so)
- Finalize documentation and update README (after publication)
- Consider better support for legacy xcode project generator, the tests will continue to run without bzlmod as the aspect requires workspace name collection udpates.

Depends on: #716, #719

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -90,11 +90,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Preflight Env
-        run: .github/workflows/preflight_env.sh
+        # Dont test Bazel 5 with bzlmod
+        run: .github/workflows/preflight_env.sh --no-bzlmod
       - name: Build and Test
         run: |
-          # iOS tests
+          # iOS tests without bzlmod on Bazel 5
           bazelisk build \
+            --noexperimental_enable_bzlmod \
             --config=ci_with_caches \
             --config=ios \
             --config=vfs \
@@ -147,7 +149,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Preflight Env
-        run: .github/workflows/preflight_env.sh
+        # TODO: stardoc doesn't currently support bzlmod: https://github.com/bazelbuild/stardoc/issues/117
+        run: .github/workflows/preflight_env.sh --no-bzlmod
         # Note: we need to pass the absolute to the Bazel run
       - name: buildifier
         run: |
@@ -166,7 +169,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Preflight Env
-        run: .github/workflows/preflight_env.sh
+        run: .github/workflows/preflight_env.sh --no-bzlmod
       - name: Run tests
         run: ./tests/xcodeproj-tests.sh --run && ./tests/test-tests.sh
       - uses: actions/upload-artifact@v2
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -3,13 +3,15 @@
 load("@build_bazel_rules_swift//swift/internal:feature_names.bzl", "SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE")
 load("//rules:features.bzl", "feature_names")
 load("@bazel_skylib//lib:selects.bzl", "selects")
+load("@buildifier_prebuilt//:rules.bzl", "buildifier")
 
-sh_binary(
+buildifier(
     name = "buildifier",
-    srcs = select({
-        "@bazel_tools//src/conditions:darwin_x86_64": ["@buildifier.mac.amd64//file"],
-        "@bazel_tools//src/conditions:darwin_arm64": ["@buildifier.mac.arm64//file"],
-    }),
+    exclude_patterns = [
+        "./.git/*",
+        "./bazel-*/*",
+    ],
+    mode = "fix",
 )
 
 config_setting(
--- a/rules/module_extensions.bzl
+++ b/rules/module_extensions.bzl
@@ -1,0 +1,43 @@
+"""Definitions for bzlmod module extensions."""
+
+load(
+    "//rules:repositories.bzl",
+    "rules_ios_dependencies",
+    "rules_ios_dev_dependencies",
+)
+load(
+    "//tools/toolchains/xcode_configure:xcode_configure.bzl",
+    _xcode_configure = "xcode_configure",
+)
+
+def _non_module_deps_impl(_):
+    rules_ios_dependencies(
+        load_bzlmod_dependencies = False,
+    )
+
+non_module_deps = module_extension(implementation = _non_module_deps_impl)
+
+def _non_module_dev_deps_impl(_):
+    rules_ios_dev_dependencies(load_bzlmod_dependencies = False)
+
+non_module_dev_deps = module_extension(implementation = _non_module_dev_deps_impl)
+
+def _xcode_configure_impl(module_ctx):
+    for mod in module_ctx.modules:
+        for xcode_configure_attr in mod.tags.configure:
+            _xcode_configure(
+                remote_xcode_label = xcode_configure_attr.remote_xcode_label,
+                xcode_locator_label = xcode_configure_attr.xcode_locator_label,
+            )
+
+xcode_configure = module_extension(
+    implementation = _xcode_configure_impl,
+    tag_classes = {
+        "configure": tag_class(
+            attrs = {
+                "remote_xcode_label": attr.string(mandatory = True),
+                "xcode_locator_label": attr.string(mandatory = True),
+            },
+        ),
+    },
+)
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -5,7 +5,10 @@ load(
     "http_archive",
     "http_file",
 )
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
+load(
+    "@bazel_tools//tools/build_defs/repo:git.bzl",
+    "new_git_repository",
+)
 
 def _maybe(repo_rule, name, **kwargs):
     """Executes the given repository rule if it hasn't been executed already.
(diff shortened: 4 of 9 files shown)


luispadron:
@thiagohmcruz mostly done with this PR, however, Xcode tests are failing probably because of how repos are named under bzlmod (they don't have a standard name). 

Do you know where the logic is to determine the workspace name for the Xcode generator?

thiagohmcruz:
IIUC your question, this:

https://github.com/bazel-ios/rules_ios/blob/master/rules/legacy_xcodeproj.bzl#L829

gets its value from the aspect collecting info here:

https://github.com/bazel-ios/rules_ios/blob/master/rules/legacy_xcodeproj.bzl#L265

thiagohmcruz:
Should this add a `Bzlmod Snippet` to releases so folks know how to consume it using Bzlmod? Also we've introduced releases recently but I assume folks might continue to consume `HEAD` for some time, should we have instructions in the root `README` on how to load it if you're using Bzlmod?

luispadron:
> IIUC your question, this:
> 
> https://github.com/bazel-ios/rules_ios/blob/master/rules/legacy_xcodeproj.bzl#L829
> 
> gets its value from the aspect collecting info here:
> 
> https://github.com/bazel-ios/rules_ios/blob/master/rules/legacy_xcodeproj.bzl#L265

Yep thats the attr we need to fixup, it's always `_main` when using `bzlmod`, this is because with bzlmod enabled `workspace_name()` and equivalent native functions in rules/aspects always return `_main`. 

From some digging the proper way to do this in bzlmod is to collect the repository name from the runfiles somehow, either via their path or some other mechanism. I don't really understand the generator enough to do that update. This continues to work in both modes, but with the bzlmod quirk of having `_main` as the Xcode project name.

I'm thinking of adding the legacy xcode tests to just the LTS CI jobs which don't use bzlmod.


> Should this add a `Bzlmod Snippet` to releases so folks know how to consume it using Bzlmod? Also we've introduced releases recently but I assume folks might continue to consume `HEAD` for some time, should we have instructions in the root `README` on how to load it if you're using Bzlmod?

Totally, I'm thinking of merging initial support which is barebones and following up with the required tools/docs to publish this into the BCR

.github/workflows/tests.yml
```diff
@@ -155,7 +156,9 @@ jobs:
           git diff --exit-code
       - name: Check docs
         run: |
+          # TODO: stardoc doesn't currently support bzlmod: https://github.com/bazelbuild/stardoc/issues/117
```
luispadron, line 159: I'll create a separate PR for this it will update the version to the required stardoc (requires some changes to our use of `bzl_library`)

BUILD.bazel
```diff
@@ -3,13 +3,15 @@
 load("@build_bazel_rules_swift//swift/internal:feature_names.bzl", "SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE")
 load("//rules:features.bzl", "feature_names")
 load("@bazel_skylib//lib:selects.bzl", "selects")
+load("@buildifier_prebuilt//:rules.bzl", "buildifier")
 
-sh_binary(
+buildifier(
```
luispadron, line 8: This replaces the custom `sh_binary` wrapped implementation, additionally this supports bzlmod.

rules/module_extensions.bzl
```diff
@@ -0,0 +1,44 @@
...
+    _xcode_configure = "xcode_configure",
+)
+
+def _non_module_deps_impl(_):
+    rules_ios_dependencies(
+        load_xchammer_dependencies = True,
+        load_bzlmod_dependencies = False,
+    )
+
+non_module_deps = module_extension(implementation = _non_module_deps_impl)
```
luispadron, line 19: This loads things not supported in the workspace (i.e. the repository rules) we had in WORKSPACE previously

rules/module_extensions.bzl
```diff
@@ -0,0 +1,44 @@
...
+
+def _non_module_dev_deps_impl(_):
+    rules_ios_dev_dependencies(load_bzlmod_dependencies = False)
+
+non_module_dev_deps = module_extension(implementation = _non_module_dev_deps_impl)
```
luispadron, line 24: This loads and runs the xcode configuration that was previously in the WORKSPACE

rules/repositories.bzl
```diff
@@ -58,20 +58,52 @@ def _get_bazel_version():
     return struct(major = 0, minor = 0, patch = 0)
 
 def rules_ios_dependencies(
```
luispadron, line 60: Changes here:

- Update versions to minimum required for bzlmod
- Move into one additional branch for bzlmod dependencies (i.e. things that support bzlmod)
  - With bzlmod these are loaded in `MODULE.bazel` so they need to be filtered here but with older WORKSPACE we default to include these

tests/xcodeproj-tests.sh
luispadron, line 1: Changes here are to disable bzlmod support for test purposes, this still works from what I can tell in bzlmod but has a `_main` for the workspace name (i.e. fails CI tests).  

I looked at updating and was blocked for a while, the change required seems to be to update the aspect that collects this the workspace name to not do that during aspect time and instead use the runfiles API to access the repository mappings file.

BUILD.bazel
```diff
@@ -3,13 +3,15 @@
...
     name = "buildifier",
-    srcs = select({
-        "@bazel_tools//src/conditions:darwin_x86_64": ["@buildifier.mac.amd64//file"],
-        "@bazel_tools//src/conditions:darwin_arm64": ["@buildifier.mac.arm64//file"],
-    }),
+    exclude_patterns = [
+        "./.git/*",
+        "./bazel-*/*",
```
jerrymarino, line 12: Does this build it from source though?

rules/repositories.bzl
```diff
@@ -91,33 +123,80 @@ def rules_ios_dependencies(
...
+        sha256 = "8f9ee2dc10c1ae514ee599a8b42ed99fa262b757058f65ad3c384289ff70c4b8",
     )
 
+    _maybe(
+        http_archive,
+        name = "buildifier_prebuilt",
+        sha256 = "e46c16180bc49487bfd0f1ffa7345364718c57334fa0b5b67cb5f27eba10f309",
+        strip_prefix = "buildifier-prebuilt-6.1.0",
+        urls = [
+            "http://github.com/keith/buildifier-prebuilt/archive/6.1.0.tar.gz",
```
jerrymarino, line 172: In the lines you've removed we already pulled the offical sources:
https://github.com/bazel-ios/rules_ios/pull/703/files#diff-5493ff8e9397811510e780de47c57abb70137f1afe85d1519130dc3679d60ce5L66

Unless there is blocking reason to change this, I'd prefer to keep that. Is there a reason you just update the URL to use a newer version of the official cut?

rules/repositories.bzl
```diff
@@ -58,20 +58,52 @@ def _get_bazel_version():
...
+        _rules_ios_bzlmod_dependencies()
+
+    if load_xchammer_dependencies:
+        _rules_ios_xchammer_dependencies()
+
+    # This goes after xchammer because it also depends on XcodeGen repo.
+    # If legacy project generator is deprecated this can be removed.
+    _rules_ios_legacy_xcodeproj_dependencies()
+
+def rules_ios_dev_dependencies(
```
jerrymarino, line 80: Why do we need this `rules_ios_dev_dependencies` function - this just adds so much complexity to the setup and repo. Bazel should only download what it is using unless you start running wildcard queries which is a performance problem that this looks like you are attempting to fix

BUILD.bazel
luispadron: `buildifier-prebuilt` just handles downloading the right `buildifier` for your architecture. It's already built

rules/repositories.bzl
luispadron: The previous WORKSPACE used an `http_file` download, I thought it made sense to use the `buildifier-prebuilt` which supports switching based on platform and supports bzlmod

rules/repositories.bzl
luispadron: Because in bzlmod you mark things as being development only dependencies when they aren't required by consumers of the rules. No one consuming rules_ios needs to download buildifier, which bzlmod solves. 

This specific change lets us reuse `repositories.bzl` while we transition to bzlmod completely (Bazel 9 I believe).

luispadron: Some examples: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/MODULE.bazel#L30

luispadron: > this just adds so much complexity to the setup

fwiw this shouldn't need to be run by end users, it's just a way for us internally to fetch development repos in both bzlmod / non-bzlmod workflows.

luispadron:
Will be blocked by #773 if that merges.

rules/module_extensions.bzl
jerrymarino: Isn't it fair to say that bzlmod was updated to work with existing repository rules? e.g. the [`use_repo`](https://github.com/bazel-ios/rules_ios/pull/703/files#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR58) wrapper can help use repos?

jerrymarino (approved):
@luispadron I think it's time to land this and forward fix. For the rules_apple 3.x update - I will rebase the changes once we land it.

jszumski (approved):

rules/module_extensions.bzl
```diff
@@ -0,0 +1,43 @@
...
+    "rules_ios_dev_dependencies",
+)
+load(
+    "//tools/toolchains/xcode_configure:xcode_configure.bzl",
+    _xcode_configure = "xcode_configure",
+)
+
+def _non_module_deps_impl(_):
+    rules_ios_dependencies(
+        load_bzlmod_dependencies = False,
```
justinseanmartin, line 15: FYI - This confused me, because I thought "load_bzlmod_dependencies" means "dependencies that bzlmod needs" rather than "dependencies that bzlmod has already supplied". Just putting this here in case anyone else stumbles across this, and as always, naming is hard.

luispadron: I had to update it the other day and I felt the same way lol

Happy to change this to something more clear

PR #705, opened 2023-05-03
[Xcode] Update Xcode version for CI

shepting:
Let's update the Xcode version used to run tests in CI.

Screenshot 2023-05-04 at 2 36 05 PM

## Depends on
https://github.com/bazel-ios/rules_ios/pull/710


.github/workflows/tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         with:
           name: bazel-testlogs
           path: bazel-testlogs
+
```
shepting, line 45: I noticed that we have newlines between half of the CI jobs but not others. This makes it conistent.

mattrobmattrob (approved):
This is what we're using internally so ✅.

luispadron (approved):

mattrobmattrob:
Guess it doesn't exist on these GitHub Actions machines?

> xcode-select: error: invalid developer directory '/Applications/Xcode_14.3.app'

qyang-nj (commented):
I believe Xcode 14.3 requires macOS 13.0 and later. We probably need to update `run-on:` to `macos-13`

thiagohmcruz (approved):
Thx for the PR! I wouldn't consider hard blockers if CI is 🟢 but it would be nice to update other Xcode related dependencies that might eventually break if we don't (could be done in a follow up):

* Any changes to [xcspecs](https://github.com/bazel-ios/rules_ios/blob/master/data/xcspecs.bzl#L5)?
* We would want the new version to be selected [here](https://github.com/bazel-ios/rules_ios/blob/master/.github/workflows/xcode_select.sh#L18) right? Meaning updating [these](https://github.com/bazel-ios/rules_ios/blob/master/BUILD.bazel#L133-L164) otherwise it will cause a mismatch with `xcode-select -s`?

shepting:
@thiagohmcruz Good call.

When I run `bazel run data_generators:extract_xcspecs` I see this error: 
```
╰─ bazel run data_generators:extract_xcspecs
Starting local Bazel server and connecting to it...
INFO: Analyzed target //data_generators:extract_xcspecs (44 packages loaded, 317 targets configured).
INFO: Found 1 target...
Target //data_generators:extract_xcspecs up-to-date:
  bazel-bin/data_generators/extract_xcspecs.sh
INFO: Elapsed time: 30.581s, Critical Path: 0.07s
INFO: 6 processes: 5 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 6 total actions
INFO: Running command line: bazel-bin/data_generators/extract_xcspecs.sh
  File "/Users/stevenhepting/workspace/rules_ios_shepting/bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/data_generators/xcspec_extractor.runfiles/build_bazel_rules_ios/data_generators/xcspec_extractor.py", line 51, in <module>
    a.extend(xcspec_from_file(path))
  File "/Users/stevenhepting/workspace/rules_ios_shepting/bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/data_generators/xcspec_extractor.runfiles/build_bazel_rules_ios/data_generators/xcspec_extractor.py", line 21, in xcspec_from_file
    xml = subprocess.check_output(
  File "/Applications/Xcode-14.3.0.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/Applications/Xcode-14.3.0.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['plutil', '-convert', 'xml1', '-o', '-', '--', '/Applications/Xcode-14.3.0.app/Contents/PlugIns/XCBSpecifications.ideplugin/Contents/Resources/Ld.xcspec']' returned non-zero exit status 1.
```

BUILD.bazel
```diff
@@ -131,34 +131,34 @@ selects.config_setting_group(
 )
 
 xcode_version(
-    name = "version14_2_0_14C18",
+    name = "version14_3_0_14E222b",
```
shepting, line 134: @thiagohmcruz Updated these versions as suggested.

Screenshot 2023-05-04 at 2 35 39 PM

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   integration_tests:
     name: Build and Test
-    runs-on: macos-12
+    runs-on: macos-13
```
shepting, line 17: @qyang-nj I had these staged locally but forgot to commit. 🤦

data_generators/xcspec_extractor.py
```diff
@@ -32,7 +32,7 @@ def xcspec_from_file(path):
 eval_bzl_path = sys.argv[4]
 xcode = os.path.dirname(os.path.dirname(developer_dir))
 xcode_plugins = os.path.join(
-    xcode, "Contents/PlugIns/XCBSpecifications.ideplugin/Contents/Resources"
+    xcode, "Contents/Developer/Library/Xcode/Plug-ins/XCBSpecifications.ideplugin/Contents/Resources"
```
shepting, line 35: @thiagohmcruz I believe this was the source of my failure before. Guessing that Xcode 14.3 puts these in this new directory?

.bazelversion
```diff
@@ -1 +1 @@
-6.0.0
+6.1.0
```
shepting, line 1: We found locally that we needed Bazel 6.1 to get the Bazel toolchain detection working with Xcode 14.3 on Intel machines.

BUILD.bazel
thiagohmcruz: cool thx!

data_generators/xcspec_extractor.py
thiagohmcruz: I see, I don't know what others think but IMO this should be backwards compatible in case we need to run this for older versions? Can we conditionally change this value based on the version?

.bazelversion
thiagohmcruz: yeah IIRC to pick up [this](https://github.com/bazelbuild/bazel//commit/43dadb275b3f9690242bf2d94a0757c721d231a9) right?

qyang-nj: Yep, I believe it's in 6.1.**1**

data_generators/xcspec_extractor.py
luispadron: This changed in 14 as well and we didnt add a flag, think its okay to bump this without it if the idea is the xccspecs are just generated by the latest supported Xcode

.bazelversion
luispadron: Thoughts on splitting these PRs? We should probably bump to 6.1.1, tag a release then merge this and tag again

tools/toolchains/xcode_configure/xcode_configure.bzl
```diff
@@ -120,7 +120,7 @@ def run_xcode_locator(repository_ctx, xcode_locator_src_label):
         "--sdk",
         "macosx",
         "clang",
-        "-mmacosx-version-min=10.9",
+        "-mmacosx-version-min=11.0",
```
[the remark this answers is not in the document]
thiagohmcruz: Just curious, any specific reason to pick `11` here? The [fix in Bazel](https://github.com/bazelbuild/bazel//commit/43dadb275b3f9690242bf2d94a0757c721d231a9) uses `10.13`

mattrobmattrob:
Any interest in moving this forward, @shepting?

mattrobmattrob:
Closing for now since we'll likely complete this update in some form as part of the Bazel 7 work. Please reopen if this isn't the desired outcome. 🙏

PR #706, opened 2023-05-05
Clean up `_write_modulemap` params

mattrobmattrob:


thiagohmcruz:
I believe there was a similar attempt to do that [here](https://github.com/bazel-ios/rules_ios/pull/572/files#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478) and I ended up reverting [here](https://github.com/bazel-ios/rules_ios/pull/581)

mattrobmattrob:
Ah, had in my head that I should check if this is an input parameter or not given the "default" naming of `_DEFAULT_LIBRARY_TOOLS`. Didn't check. I wish there was a more expressive way to declare these interface requirements vs. leaving them dangling in `rules_ios`.

thiagohmcruz:
> 

+1 for having a more expressive way to declare these, seems doable, I think we just need someone to do it haha (should probably be tracked in an issue if it's not already)

PR #689, opened 2023-04-15
[Cleanup] Remove unused hmap

jerrymarino:
This is a followup to address the issue raised in #688. It looks like we have an unused hmap which removing it will cleanup many actions

PR #708, opened 2023-05-08
Only create headermap when necessary

qyang-nj:
Only create headermap when the header list is not empty. Otherwise we would create empty `.hmap` file. This optimization can reduce lots of actions for  Swift-dominant code base.


mattrobmattrob (commented):
👏 🎉 
```
# Before
$ bazel query 'kind(headermap,deps(//RedditApp))' | wc -l
Loading: 0 packages loaded
Loading: 0 packages loaded
    1688

# After
$ bazel query 'kind(headermap,deps(//RedditApp))' | wc -l
     687
```

thiagohmcruz:
> 👏 🎉
> 
> ```
> # Before
> $ bazel query 'kind(headermap,deps(//RedditApp))' | wc -l
>     1688
> 
> # After
> $ bazel query 'kind(headermap,deps(//RedditApp))' | wc -l
>      687
> ```

Wow for us that's `before=10624` => `after=4009` 🎉

thiagohmcruz (approved):
Amazing, thank you! If CI is 🟢 here we should be good to know. Even if this surfaces another transitive/unexpected issue later it's probably something we should fix instead of reverting this anyways as this change makes a lot of sense

qyang-nj:
Nice! For us, it's 5690 (before) => 140 (after).

mattrobmattrob (approved):
Does this replace https://github.com/bazel-ios/rules_ios/pull/689?

mattrobmattrob:
Try updating those generated Xcode projects using `./tools/xcodeproj-tests.sh --update`. It _should_ work:

https://github.com/bazel-ios/rules_ios/blob/aa0f237c0ebb91c7e2493d7c81b8085bdb775aeb/tests/xcodeproj-tests.sh#L83-L93

rules/library.bzl
```diff
@@ -841,62 +841,47 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
+        )
+        private_dep_names.append(private_hmap_name)
+        _append_headermap_copts(private_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
 
     ## END HMAP
 
-    _append_headermap_copts(private_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
-    _append_headermap_copts(public_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
-    _append_headermap_copts(private_angled_hmap_name, "-I", additional_objc_copts, additional_swift_copts, additional_cc_copts)
-    _append_headermap_copts(private_hmap_name, "-iquote", additional_objc_copts, additional_swift_copts, additional_cc_copts)
```
thiagohmcruz, line 898: Sorry noticed something I should've asked in my previous review.

Since this is being deleted mind clarifying how it maps to the final code in this PR? E.g. why `private_angled_hmap_name` and the `"-iquote"` line are not being used anymore?

ob:
Nice!

rules/library.bzl
```diff
@@ -841,62 +841,47 @@ def apple_library(name, library_tools = {}, export_private_headers = True, names
...
-        tags = _MANUAL,
-    )
-    native.filegroup(
-        name = private_angled_hdrs_filegroup,
-        srcs = objc_non_exported_hdrs + objc_private_hdrs,
-        tags = _MANUAL,
-    )
+        native.filegroup(
+            name = private_hdrs_filegroup,
+            srcs = objc_non_exported_hdrs + objc_private_hdrs,
```
ob, line 871: was dropping `objc_hdrs` from here intentional?

rules/library.bzl
ob: These are moved to the `if`'s above.

ob (approved):

mattrobmattrob:
> > Does this replace #689?
> 
> It's not touching the same code path so I don't think so?

🤦 Yeah, meant the comment that inspired part of that PR: https://github.com/bazel-ios/rules_ios/pull/688#discussion_r1167300415

qyang-nj:
> Does this replace https://github.com/bazel-ios/rules_ios/pull/689?

> Since this is being deleted mind clarifying how it maps to the final code in this PR? E.g. why private_angled_hmap_name and the "-iquote" line are not being used anymore?

No. it doesn't replace that PR, but it has the similar idea. `-I` basically includes `-iquote`, so it doesn't need to be specified twice.

rules/library.bzl
mattrobmattrob: Won't it just get it from `public_hdrs_filegroup` now?

qyang-nj: Yes, `objc_hdrs` was added to the `public_hdrs_filegroup` above.  I don't think we need to add it to the private one again.

tests/macos/xcodeproj/Single-Application-Project-AllTargets.xcodeproj/project.pbxproj
```diff
@@ -465,7 +465,7 @@
 				BAZEL_BUILD_TARGET_WORKSPACE = build_bazel_rules_ios;
 				BAZEL_LLDB_INIT_FILE = "$CONFIGURATION_TEMP_DIR/Single-Application-UnitTests.lldbinit";
 				BAZEL_LLDB_SWIFT_EXTRA_CLANG_FLAGS = "-D'REQUIRED_DEFINED_FLAG=1' -D'FLAG_WITH_VALUE_ZERO=0'";
-				BAZEL_SWIFTMODULEFILES_TO_COPY = "bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftmodule bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftdoc bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftsourceinfo";
+				BAZEL_SWIFTMODULEFILES_TO_COPY = "bazel-out/ios-sim_arm64-min11.0-applebin_ios-ios_sim_arm64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftmodule bazel-out/ios-sim_arm64-min11.0-applebin_ios-ios_sim_arm64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftdoc bazel-out/ios-sim_arm64-min11.0-applebin_ios-ios_sim_arm64-dbg-ST-103f2e67c6d8/bin/tests/macos/xcodeproj/Single_Application_UnitTests.swiftsourceinfo";
```
thiagohmcruz, line 468: I don't think we want the arch to change here? `x86_64` => `arm64`

qyang-nj: I just ran `./tests/xcodeproj-tests.sh --update`. And `arch -arch x86_64 /bin/bash -l -c "./tests/xcodeproj-tests.sh --update"` is the same 🤦🏻‍♂️. Let me see how I can fix this.

.bazelrc
```diff
@@ -6,7 +6,7 @@ build --strategy=Stardoc=standalone
 # build --genrule_strategy=standalone
 # build --spawn_strategy=standalone
 
-build --verbose_failures # Print the full command line for commands that failed
+build --verbose_failures   # Print the full command line for commands that failed
```
luispadron, line 9: nit:

```suggestion
build --verbose_failures # Print the full command line for commands that failed
```

luispadron (approved):
This is great, +1 to @thiagohmcruz comment around `private_angled_hmap_name`, it looks like it's no longer used after this change.

LGTM once we correct the fixture diffs + discuss `private_angled_hmap_name`.

.bazelrc
qyang-nj: oops, mistake. reverted.

rules/library.bzl
ob: Ah, I missed that... thanks!

jszumski:
Before:
```
bazel query 'kind(headermap,deps(//Code/Apps/CashApp))' | wc -l
    1718
```

After: 
```
bazel query 'kind(headermap,deps(//Code/Apps/CashApp))' | wc -l
     538
```

🎉

PR #710, opened 2023-05-08
[Deps] Update Podfile dependencies to build on Xcode 14.3

shepting:
Update these dependencies so they can build on Xcode 14.3. This PR:
* updates the version of `cocoapods`
* adds a `.ruby-version` file for our test that uses CocoaPods
* updates the Pods to versions that will build with Xcode 14.3

## CocoaPods bug fixes
Cocoapods 1.12.1 has the following bugfixes:
Screenshot 2023-05-08 at 2 07 22 PM
https://github.com/CocoaPods/CocoaPods/releases/tag/1.12.1

## Build errors
Before:
Screenshot 2023-05-08 at 1 53 07 PM

After:
Screenshot 2023-05-08 at 1 57 02 PM


tests/ios/frameworks/sources-with-prebuilt-binaries/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.4
```
shepting, line 1: Totally open to changing this to whatever folks want. But it was helpful in my case to fix the following error: 
```
╰─ bundle install         
Warning: the running version of Bundler (2.1.4) is older than the version that created the lockfile (2.3.7). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.3.7`.
Fetching gem metadata from https://rubygems.org/.......
activesupport-7.0.4.3 requires ruby version >= 2.7.0, which is incompatible with the current version, ruby 2.6.10p210
```

tests/ios/frameworks/sources-with-prebuilt-binaries/Podfile.lock
```diff
@@ -50,14 +64,14 @@ SPEC REPOS:
...
+  GoogleUtilities: 9aa0ad5a7bc171f8bae016300bfcfa3fb8425749
+  nanopb: b552cce312b6c8484180ef47159bc0f65a1f0431
+  PromisesObjC: 09985d6d70fbe7878040aa746d78236e6946d2ef
+  SnapKit: e01d52ebb8ddbc333eefe2132acf85c8227d9c25
 
-PODFILE CHECKSUM: effa09ca7a4286ad7bc8fbfeb4844754d60fb94e
+PODFILE CHECKSUM: dbe40c09119e5d201de7dfd4c31e0e6f1c85bd6b
 
-COCOAPODS: 1.8.4
+COCOAPODS: 1.12.1
```
shepting, line 77: Really not sure how this Lockfile got so out of date.

tests/ios/frameworks/sources-with-prebuilt-binaries/.ruby-version
mattrobmattrob: I'd probably optimize this for whatever the Github runners have on them for now: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#ruby

So perhaps `2.7.8`?
```suggestion
2.7.8
```

mattrobmattrob: `macos-11` also have `2.7.8` so I'd say we go with that: https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#ruby

tests/ios/frameworks/sources-with-prebuilt-binaries/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.8
```
luispadron, line 1: Thank you for this, always forget to setup the env

luispadron (approved):
Thank you for updating this!

Yet another reason imo to remove `cocoapods` and the `carthage` rules if they are no longer used by too many people. Makes simple Xcode updates more difficult

shepting:
@luispadron Do you have any ideas how to fix this latest error? 
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/cocoapods/BUILD.bazel", line 6, column 10, in <toplevel>
		filegroup(
Error in filegroup: filegroup rule 'UserMessagingPlatform' in package '' conflicts with existing filegroup rule, defined at /private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/cocoapods/BUILD.bazel:1:10
ERROR: /Users/runner/work/rules_ios/rules_ios/tests/ios/frameworks/sources-with-prebuilt-binaries/BUILD.bazel:20:31: no such target '@cocoapods//:GoogleMobileAds': target 'GoogleMobileAds' not declared in package '' defined by /private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/external/cocoapods/BUILD.bazel (Tip: use `query "@cocoapods//:*"` to see all the targets in that package) and referenced by '//tests/ios/frameworks/sources-with-prebuilt-binaries:GoogleMobileAds'
```
https://github.com/bazel-ios/rules_ios/actions/runs/4920143706/jobs/8788642789?pr=710#step:4:226

tests/ios/frameworks/sources-with-prebuilt-binaries/Gemfile
```diff
@@ -1,4 +1,5 @@
-source 'https://rubygems.org' do
-  gem 'cocoapods', '1.12.0'
-  gem 'cocoapods-binary', '0.4.4'
-end
+source 'https://rubygems.org'
```
shepting, line 1: Updated this in response to an error log on CI.

shepting:
I'm curious what folks think we should do about the `@cocoapods` feature if there's an `.xcframework` inside it.

cc: @jerrymarino @qyang-nj @mattrobmattrob

shepting:
It looks like the `cocoapods-binary` project hasn't been updated in ~4 years:
Screenshot 2023-05-08 at 7 12 10 PM

mattrobmattrob:
> I'm curious what folks think we should do about the `@cocoapods` feature if there's an `.xcframework` inside it.
> 
> 
> 
> cc: @jerrymarino @qyang-nj @mattrobmattrob 

I vote we ask in the Bazel/other Slacks about this functionality, get the Airbnb/Snap/Square/Block/LinkedIn/other's sign off that they don't use it, and then remove it. We shouldn't keep maintaining it if it's unused. 👍

luispadron:
I think this can be closed now that we've removed cocoapods and Carthage support. Maybe we can keep the ruby version file though

mattrobmattrob:
Closing since the Carthage/CocoaPods dirs were removed in https://github.com/bazel-ios/rules_ios/pull/716.

PR #714, opened 2023-05-09
Fix infoplist collection in the legacy Xcode project generator

thiagohmcruz:
AFAICT the generated plists should be the ones being picked up by the generator, otherwise the plist in Xcode means nothing as it is only the input for the macro and not the merged one: https://github.com/bazel-ios/rules_ios/blob/master/rules/framework.bzl#L123


thiagohmcruz:
These plist paths were added by me [here](https://github.com/bazel-ios/rules_ios/pull/661/commits/3d6eb5642cb6cac9e5a7065be285715c396151c6) and I don't remember the context 😓

We should root cause this differently and understand why CI is creating those in the first place, will iterate on that [here](https://github.com/bazel-ios/rules_ios/pull/717)

Issue #712, opened 2023-05-09
Bug: Prebuilt XCFramework errors at `import` statements


mattrobmattrob:
Branch: https://github.com/bazel-ios/rules_ios/compare/master...mattrobmattrob:mr/broken.xcframework.deps?expand=1

Repro:
```
$ bazel build --features apple.virtualize_frameworks //tests/ios/xcframeworks/Basic:XCFrameworksApp
INFO: Analyzed target //tests/ios/xcframeworks/Basic:XCFrameworksApp (24 packages loaded, 897 targets configured).
INFO: Found 1 target...
ERROR: /Users/matt.robinson/Developer/rules_ios/rules_ios/tests/ios/xcframeworks/Basic/BUILD.bazel:5:16: Compiling Swift module //tests/ios/xcframeworks/Basic:xcframeworks_swift failed: (Exit 1): worker failed: error executing command (from target //tests/ios/xcframeworks/Basic:xcframeworks_swift)
  (cd /private/var/tmp/_bazel_matt.robinson/0e64031a9efbb95e4728094ee9457143/execroot/build_bazel_rules_ios && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=16.4 \
    SWIFT_AVOID_WARNING_USING_OLD_DRIVER=1 \
    XCODE_VERSION_OVERRIDE=14.3.0.14E222b \
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-103902c0b028/bin/external/build_bazel_rules_swift/tools/worker/worker swiftc @bazel-out/ios-sim_arm64-applebin_ios-ios_sim_arm64-dbg-ST-2a0e8ccdcbc3/bin/tests/ios/xcframeworks/Basic/xcframeworks.swiftmodule-0.params)
# Configuration: a9e3074c64353107d7b3531998216c3eec5702a13967b6f12e804b2a63420e30
# Execution platform: @local_config_platform//:host
error: emit-module command failed with exit code 1 (use -v to see invocation)
external/Instabug/Instabug.xcframework/ios-arm64_i386_x86_64-simulator/Instabug.framework/Headers/Instabug.h:13:9: note: while building module 'Instabug_import_middleman' imported from external/Instabug/Instabug.xcframework/ios-arm64_i386_x86_64-simulator/Instabug.framework/Headers/Instabug.h:13:
#import "IBGNetworkLogger.h"
[...]

PR #718, opened 2023-05-10
[Retry] Fix transitive framework import in opt build

qyang-nj:
**Retry merging PR #570, which solves issue #569.** The original PR got reverted because of issue #712.

### Original description
`apple_*_(xc)framework_import` provides `_SwiftInteropInfo` instead of `SwiftInfo`. In `apple_framework_packaging`, when [we merge the SwiftInfo](https://github.com/bazel-ios/rules_ios/blob/b1499fc08a7e4b7a7ba238fb4c9b089e6648ecbd/rules/framework.bzl#L901), the info from the framework import get lost there.

To fix the problem, we can add `swift_clang_module_aspect` to the `transitive_deps` attribute of `apple_framework_packaging`, which converts `_SwiftInteropInfo` to `SwiftInfo`. `swift_library` does this for `deps` as well.

@mattrobmattrob @thiagohmcruz

### DIFF
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -264,15 +264,9 @@ def _file_collector_rule_impl(ctx):
     )
 
     additional_providers = []
-    if not virtualize_frameworks:
-        dep_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep]
-        cc_info = cc_common.merge_cc_infos(direct_cc_infos = [], cc_infos = dep_cc_infos)
-        additional_providers.append(cc_info)
-    else:
-        cc_info = objc_provider_utils.merge_cc_info_providers(
-            cc_info_providers = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
-        )
-        additional_providers.append(cc_info)
+    dep_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep]
+    cc_info = cc_common.merge_cc_infos(direct_cc_infos = [], cc_infos = dep_cc_infos)
+    additional_providers.append(cc_info)
 
     return [
         DefaultInfo(files = depset(dynamic_framework_dirs + replaced_frameworks)),
--- a/rules/internal/objc_provider_utils.bzl
+++ b/rules/internal/objc_provider_utils.bzl
@@ -1,4 +1,4 @@
- def _add_to_dict_if_present(dict, key, value):
+def _add_to_dict_if_present(dict, key, value):
     if value:
         dict[key] = value
 
--- a/rules/internal/objc_provider_utils.bzl
+++ b/rules/internal/objc_provider_utils.bzl
@@ -1,4 +1,4 @@
-def _add_to_dict_if_present(dict, key, value):
+ def _add_to_dict_if_present(dict, key, value):
     if value:
         dict[key] = value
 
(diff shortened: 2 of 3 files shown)


rules/import_middleman.bzl
```diff
@@ -264,15 +264,9 @@ def _file_collector_rule_impl(ctx):
...
     additional_providers = []
-    if not virtualize_frameworks:
-        dep_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep]
-        cc_info = cc_common.merge_cc_infos(direct_cc_infos = [], cc_infos = dep_cc_infos)
-        additional_providers.append(cc_info)
-    else:
-        cc_info = objc_provider_utils.merge_cc_info_providers(
-            cc_info_providers = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep],
-        )
-        additional_providers.append(cc_info)
```
qyang-nj, line 275: This custom merging logic caused a problematic `modulemap` is generated. I'm not sure why we are merging `CcInfo` differently depending on whether VFS is used, but all CI jobs are passed.

mattrobmattrob:
Running an integration test in our repo now. Do you mind doing the same when you have time, @thiagohmcruz?

mattrobmattrob (approved):
LGTM in our project. Let's let the Square/Block folks give it a test in their codebase before we merge though. 🙏

rules/internal/objc_provider_utils.bzl
```diff
@@ -55,29 +55,9 @@ def _merge_dynamic_framework_providers(dynamic_framework_providers):
 
     return apple_common.new_dynamic_framework_provider(**fields)
 
-def _merge_cc_info_providers(cc_info_providers, merge_keys = [
```
qyang-nj, line 58: Actually this is not used anywhere else.

jszumski (approved):
This worked for Cash 👍

mattrobmattrob:
We can merge this and revert if Block folks have any issues. Thanks for fixing the previous problem, @qyang-nj!

thiagohmcruz:
> We can merge this and revert if Block folks have any issues. Thanks for fixing the previous problem, @qyang-nj!

+1 thx for the PR @qyang-nj, sorry about the delay I'm intermittently available this week, will try to run some tests on this sometime this week

jerrymarino (commented):
Thanks for the PR - left a comment inline and will be able to followup on this at a later day if necessary!

rules/import_middleman.bzl
thiagohmcruz: We should be able to though right? If we can't rely on CI being 🟢 here then I think we should look into it and make CI start checking for whatever is necessary?

qyang-nj: Yeah, I have the repro example in the issue #569 and the repro code [here](https://github.com/qyang-nj/BazelPlayground/tree/main/Issue0). I thought about adding it to CI tests, but realized we don't have any tests run under `--compilation_mode=opt`. Maybe we can add that.

jerrymarino: SGTM! Yeah it'd be nice to improve the coverage for the things we can, I can dig more into this early next week.

rules/internal/objc_provider_utils.bzl
jerrymarino: Overall this is "internal" but this overall utility module clearly has many useful functions none the less - we should clarify API contracts with this code.

PR #716, opened 2023-05-09
Remove Carthage and Cocoapod rules

luispadron:
These two rules require a fair amount of maintenance and constantly break between large changes like Bazel / Xcode upgrades. In the interest of the health of `rules_ios` we believe these should be deprecated in favor of other community rules which do similar things.

This change should help the `rules_ios` contributors more easily make changes without maintaining rules that aren't used by many folks.

If you do use these rules and depend on it and don't agree with this direction, please let us know!

PR #719, opened 2023-05-10
Add remote cache and CI configs

luispadron:
Before this change the CI and cache configs were being generated by `.github/workflows/xcode_select.sh` which makes it difficult for a contributor to find these and to then apply them.

Changes:
 - New Bazel configurations that make it easier to reproduce CI setup and debug issues locally.
 - New configs for testing specific tests that were previously only in the workflow file.
 - Removes the `user.bazelrc` generation CI was doing in favor of the new `ci` and `cache` configs.
 - Updates the workflows to use the new configs where needed.

Depends on #716

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -54,9 +54,23 @@ build:ios_multi_arch_test --ios_minimum_os=10.2
 build:ios_multi_arch_test --ios_multi_cpus=i386,x86_64
 
 # - Remote cache configuration
+build:remote_cache --bes_backend=grpcs://remote.buildbuddy.io
+build:remote_cache --bes_results_url=https://app.buildbuddy.io/invocation/
+build:remote_cache --experimental_build_event_upload_strategy=local
+build:remote_cache --experimental_remote_cache_compression
+build:remote_cache --experimental_remote_merkle_tree_cache
+build:remote_cache --modify_execution_info=^(BitcodeSymbolsCopy|BundleApp|BundleTreeApp|DsymDwarf|DsymLipo|GenerateAppleSymbolsFile|ObjcBinarySymbolStrip|CppLink|ObjcLink|ProcessAndSign|SignBinary|SwiftArchive|SwiftStdlibCopy)$=+no-remote,^(BundleResources|ImportedDynamicFrameworkProcessor)$=+no-remote-exec
+build:remote_cache --nolegacy_important_outputs
+build:remote_cache --noslim_profile
 build:remote_cache --remote_cache=grpcs://remote.buildbuddy.io
 build:remote_cache --remote_timeout=3600
-build:remote_cache --modify_execution_info=^(BitcodeSymbolsCopy|BundleApp|BundleTreeApp|DsymDwarf|DsymLipo|GenerateAppleSymbolsFile|ObjcBinarySymbolStrip|CppLink|ObjcLink|ProcessAndSign|SignBinary|SwiftArchive|SwiftStdlibCopy)$=+no-remote,^(BundleResources|ImportedDynamicFrameworkProcessor)$=+no-remote-exec
+
+# Enable remote_cache config by default
+build --config=remote_cache
+
+# By default don't upload local results to remote cache, only CI does this.
+build --noremote_upload_local_results
+build:ci --remote_upload_local_results
 
 # - CI Configuration
 # disable curses control since CI logs are not interactive
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,9 +2,11 @@
 # so circumvent by not using the sandbox
 build --strategy=Stardoc=standalone
 
-# Debugging does not work in sandbox mode. Uncomment these lines to turn off sandboxing.
-# build --genrule_strategy=standalone
-# build --spawn_strategy=standalone
+# Spawn strategy required for some tests
+build --spawn_strategy=standalone
+
+# Setup Xcode configuration for xcodeproj tests
+build --xcode_version_config=//:host_xcodes
 
 build --verbose_failures # Print the full command line for commands that failed
 build --test_output=errors # Prints log file output to the console on failure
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -141,19 +165,26 @@ jobs:
...
       - uses: actions/checkout@v3
       - name: Select Xcode 14.2.0
         run: .github/workflows/xcode_select.sh
       - name: Run tests
-        run: bazelisk test tests/ios/lldb/app:objc_app_po_test  tests/ios/lldb/app:objc_app_variable_test --config lldb_ios_test
+        run: |
+          bazelisk test \
+            --config=ci \
+            --config=ios_lldb_test \
+            tests/ios/lldb/app:objc_app_po_test \
```
jszumski, line 179:
```suggestion
            -- \
            tests/ios/lldb/app:objc_app_po_test \
```

.github/workflows/tests.yml
```diff
@@ -162,4 +193,8 @@ jobs:
       - name: Select Xcode
         run: .github/workflows/xcode_select.sh
       - name: Build App
-        run: bazelisk build -s tests/ios/app/App --apple_platform_type=ios --ios_minimum_os=10.2  --ios_multi_cpus=i386,x86_64
+        run: |
+          bazelisk build \
+            --config=ci \
+            --config=ios_multi_arch_test \
+            tests/ios/app/App
```
jszumski, line 200:
```suggestion
            -- \
            tests/ios/app/App
```

jszumski (approved):
This was always a pain point for me, thank you!

.github/workflows/tests.yml
```diff
@@ -67,18 +77,23 @@ jobs:
           path: bazel-testlogs
 
   lts_ios_integration_tests:
-    name: Build and Test ( Virtual Frameworks + LTS )
+    name: Bazel 5 - Build and Test (Virtual Frameworks)
```
mattrobmattrob, line 80: If we're going to be updating these, I'd prefer it stay beside the `Virtual Frameworks` portion since it's the "modifiers" to the build. I think the prefix makes the build list hard to parse:
Screenshot 2023-05-10 at 9 29 18 AM
So I guess:
```suggestion
    name: Build and Test (Virtual Frameworks, Bazel 5)
```

luispadron: That makes sense, went back and forth on it

luispadron:
Someone with Admin access will need to update the required jobs for the branch, if we think the name changes are worth it

.github/workflows/tests.yml
```diff
@@ -162,4 +202,9 @@ jobs:
       - name: Select Xcode
         run: .github/workflows/xcode_select.sh
       - name: Build App
-        run: bazelisk build -s tests/ios/app/App --apple_platform_type=ios --ios_minimum_os=10.2  --ios_multi_cpus=i386,x86_64
```
mattrobmattrob, line 165: Was `-s` intentionally removed? What did it do?

mattrobmattrob (approved):
Mostly nits. Take it or leave it. This is a great improvement for folks working on the ruleset. 👏

thiagohmcruz:
> Someone with Admin access will need to update the required jobs for the branch, if we think the name changes are worth it

Probably @justinseanmartin could help here? Where exactly we would have to update those? I've never done that myself 😓

luispadron:
> > Someone with Admin access will need to update the required jobs for the branch, if we think the name changes are worth it
> 
> Probably @justinseanmartin could help here? Where exactly we would have to update those? I've never done that myself 😓

It under Settings > Branches > Branch protection > `master` > Required checks

.github/workflows/tests.yml
luispadron: That prints the subcommands, i meant to add this `config:ci`. Do we think thats useful, it could be very verbose but its for CI so?

.github/workflows/tests.yml
```diff
@@ -67,18 +80,24 @@ jobs:
...
-    name: Build and Test ( Virtual Frameworks + LTS )
+    name: Build and Test (Virtual Frameworks, Bazel 5)
     runs-on: macos-12
+    env:
+      USE_BAZEL_VERSION: 5.3.2
     steps:
       - uses: actions/checkout@v3
       - name: Select Xcode
         run: .github/workflows/xcode_select.sh
       - name: Build and Test
```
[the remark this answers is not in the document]
luispadron: For sure, this confused me when i saw it was just running `build`

.bazelrc
```diff
@@ -28,14 +28,48 @@ build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
...
-
 # Delete the VM test suite for github
 build --deleted_packages tests/ios/vmd
 
 # Allow for server to start (GitHub actions can timeout)
 startup --local_startup_timeout_secs=240
 
+# - iOS Configuration
+# Required for tests/ios/... targets
+build:ios --apple_platform_type=ios
```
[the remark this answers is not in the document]
luispadron: I prefer the shorter name in this case but if we think platform is more clear thats fine too!

.github/workflows/tests.yml
luispadron: Added this to `config=ci` as well as `--verbose_failures` figure its better to be verbose in CI.

.bazelrc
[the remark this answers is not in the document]
mattrobmattrob: All good then. 👍

luispadron:
Weird @thiagohmcruz seeing a similar plist diff now:

```sh
-				INFOPLIST_FILE = ../../../tests/ios/frameworks/dynamic/a/Info.plist;
```

Let me double check I didnt change any underlying flags but its odd

thiagohmcruz:
> 

Yeah, the question is if we even want that plist there since we set it to empty for apps and extensions we might as well be ok with setting it to empty here as well. The things is that removing it causes other issues apparently, I was trying to iterate on that [here](https://github.com/bazel-ios/rules_ios/pull/717)

luispadron:
@thiagohmcruz Will wait for that PR then!

.github/workflows/tests.yml
```diff
@@ -21,44 +21,60 @@ jobs:
...
         run: |
-          # Host config
-          bazelisk test --local_test_jobs=1 -- //... -//tests/ios/...
+          # non-iOS tests
+          bazelisk test \
+            --config=ci \
+            -- \
+            //... \
+            -//tests/ios/...
```
qyang-nj, line 29: nit: I like this to be on a single line, so that it's easier to copy it into to terminal. Same as other commands as well.

qyang-nj (commented):
Thanks! This hidden `user.bazlerc` bite me before.

luispadron:
@thiagohmcruz fixed the Xcode tests, realized the ci generated rc file set the spawn strategy by default. I've enabled it by default in the main rc file now.

luispadron:
This is ready to merge as soon as we can disable the required jobs that have been renamed. 

Once this merges we can enable the new ones. 

@thiagohmcruz @justinseanmartin whenever y'all have time or if you want to make me admin I can do it

mattrobmattrob (approved):

luispadron:
Actually moving the renaming to a separate PR to not block this further: #722

PR #722, opened 2023-05-12
Rename CI jobs to be uniform

luispadron:
Some general organization of the CI jobs, let me know your thoughts.


mattrobmattrob (approved):

justinseanmartin (approved):
Thanks for the housekeeping!

luispadron:
One thing we could consider is adding [job matrix ](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) instead of separate jobs for bazel 5/6 and other jobs like vfs/non-vfs. 

Might make it more obvious it tests the same thing but under a different configuration.

.github/workflows/tests.yml
```diff
@@ -113,8 +113,8 @@ jobs:
           name: bazel-testlogs
           path: bazel-testlogs
 
-  buildifier:
-    name: Check Starlark and Docs
+  lint:
+    name: Check Formatting and Docs
     runs-on: macos-12
```
jerrymarino, line 118: On a side note, I just wonder if this even has to run on the macOS

jerrymarino (approved):
🚀

jerrymarino:
@luispadron job matrix seems like a nice cleanup, I wonder we can ( also ) use Bazel to make this a bit nicer. 

Consider adding a config tree in `ci.bazlerc` to make these easier to repro locally:

```
common: test_targets_ios tests/ios/...
common: test_targets_ios_virtual_frameworks --config test_targets_ios
common: test_targets_ios_incompatible_no_virtual_frameworks --config test_targets_ios
```

jerrymarino:
Today I constantly end up copy/pasting the contents of this file when I have an issue.. it's deff not ideal.

.github/workflows/tests.yml
luispadron: Probably not good point, I'll update

jerrymarino: Cool you might just need to pull a different `buildifier` binary in, otherwise we can land this and followup ;

jerrymarino:
@luispadron love this overall, want to land it? We can add more improvements later.

luispadron:
> @luispadron love this overall, want to land it? We can add more improvements later.  

Yeah I'd like to get it in if we want it. Think we just need to update the required checks to match the new job names though. Don't have access to do that for this repo though

luispadron:
I'll resolve conflicts and ping you when I'm back next week if you can change the settings (or give me access)

PR #724, opened 2023-05-18
support .inc file for func `_private_headers`

Su-Xiaofeng:
apple_library has has supported include files with the suffix `.inc`. 
But this type of file is not allowed when creating a `PrivateHeadersInfo` provider


mattrobmattrob (approved):

mattrobmattrob:
Mind adding some test/validation of this functionality in a follow up PR, @Su-Xiaofeng?

PR #727, opened 2023-06-01
Update rules_apple to 2.3.0 + remove bitcode support

mattrobmattrob:
Update `rules_apple` to https://github.com/bazelbuild/rules_apple/releases/tag/2.3.0 (https://github.com/bazelbuild/rules_apple/commit/915ac30a9fa1fd3809599a5ab90fa1c6640fe8dc) and handle https://github.com/bazelbuild/rules_apple/pull/1534.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -721,16 +721,6 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
         ),
     )
     processor_partials.append(
-        partials.bitcode_symbols_partial(
-            actions = actions,
-            binary_artifact = binary_artifact,
-            bitcode_symbol_maps = debug_outputs.bitcode_symbol_maps,
-            dependency_targets = dep_frameworks,
-            label_name = label.name,
-            platform_prerequisites = platform_prerequisites,
-        ),
-    )
-    processor_partials.append(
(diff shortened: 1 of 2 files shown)


luispadron (approved):

thiagohmcruz (approved):
Do we expect the code being removed to break the "Bazel 5 + old rules_apple" code path? Any way we can verify that?

luispadron:
I'm not entirely sure why we're supporting two Bazel versions on main. Should we not instead create a branch specifically for Bazel 5?  This makes maintenance really tough...

mattrobmattrob:
I was relying on `Tests / Build and Test ( Virtual Frameworks + LTS )` to verify the Bazel 5/LTS codepath. Do you think it's not good enough, @thiagohmcruz?

thiagohmcruz:
Oh my bad, forgot that there's an explicitly job to cover that, it should be enough yeah 

@luispadron there's a README IIRC to cover the LTS support, maybe a start is to review that and propose changes / a diff approach? 

I def can see the pain points in maintaining it, at the same time it was introduced to help enforce there is at least one code path we consider "stable". That said, there's other ways to achieve the same thing and any approach will have trade-offs to discuss IMO.

PR #729, opened 2023-06-02
Fix xcconfig passthrough for `ios_application`

jszumski:
Fix issue where `xcconfig` and `xcconfig_by_build_setting ` values for non-plist settings (e.g., Swift compiler flags) were not applied for `ios_application`.

I found this because we had `SWIFT_TREAT_WARNINGS_AS_ERRORS` set but didn't see `-warnings-as-errors` set on the swiftc invocation.

### DIFF
--- a/rules/app.bzl
+++ b/rules/app.bzl
@@ -82,6 +82,8 @@ def ios_application(
         namespace_is_module_name = False,
         platforms = {"ios": application_kwargs.get("minimum_os_version")},
         testonly = testonly,
+        xcconfig = xcconfig,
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):

thiagohmcruz (approved):
Nice catch! Maybe this is a regression from when we explicitly added these as params in the macro 🤔

luispadron (approved):
Thanks! I broke this with the argument change for the macros. Double checked and this looks like the only one since the other macros doesn't need to forward to apple_library

PR #734, opened 2023-07-06
Add support for forwarding symbol graph

luispadron:
`rules_swift` has support for `symbol_graph` in `SwiftInfo`: https://github.com/bazelbuild/rules_swift/pull/838

The framework target should forward that as well

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -330,6 +330,8 @@ def _get_framework_files(ctx, deps):
     swiftdoc_out = None
     infoplist_in = None
     infoplist_out = None
+    symbol_graph_in = None
+    symbol_graph_out = None
 
     # collect files
     for dep in deps:
@@ -386,6 +388,11 @@ def _get_framework_files(ctx, deps):
                     elif hdr.path.endswith(".modulemap"):
                         modulemap_in = hdr
 
+        # If theres a symbol graph, we need to copy it over
+        if dep.output_groups and hasattr(dep.output_groups, "swift_symbol_graph"):
+            symbol_graph_in = dep.output_groups.swift_symbol_graph.to_list()[0]
+            symbol_graph_out = paths.join(framework_dir, framework_name + ".symbolgraph")
+
         if not has_header:
             # only thing is the generated module map -- we don't want it
             continue
@@ -435,6 +442,7 @@ def _get_framework_files(ctx, deps):
     swiftinterface_out = _framework_packaging_single(ctx, "swiftinterface", [swiftinterface_in], swiftinterface_out, framework_manifest)
     swiftdoc_out = _framework_packaging_single(ctx, "swiftdoc", [swiftdoc_in], swiftdoc_out, framework_manifest)
     infoplist_out = _framework_packaging_single(ctx, "infoplist", [infoplist_in], infoplist_out, framework_manifest)
+    symbol_graph_out = _framework_packaging_single(ctx, "symbol_graph", [symbol_graph_in], symbol_graph_out, framework_manifest)
 
     outputs = struct(
         binary = binary_out,
@@ -446,6 +454,7 @@ def _get_framework_files(ctx, deps):
         swiftdoc = swiftdoc_out,
         swiftinterface = swiftinterface_out,
         manifest = framework_manifest,
+        symbol_graph = symbol_graph_out,
     )
 
     inputs = struct(
@@ -456,6 +465,7 @@ def _get_framework_files(ctx, deps):
         swiftmodule = swiftmodule_in,
         swiftdoc = swiftdoc_in,
         swiftinterface = swiftinterface_in,
+        symbol_graph = symbol_graph_in,
     )
     return struct(inputs = inputs, outputs = outputs)
 
@@ -528,6 +538,7 @@ def _copy_swiftmodule(ctx, framework_files):
         swiftdoc = outputs.swiftdoc,
         swiftmodule = outputs.swiftmodule,
         swiftinterface = outputs.swiftinterface,
+        symbol_graph = outputs.symbol_graph,
     )
 
     if swiftmodule_name != ctx.attr.framework_name:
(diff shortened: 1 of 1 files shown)


rules/framework.bzl
```diff
@@ -522,12 +522,25 @@ def _copy_swiftmodule(ctx, framework_files):
...
+    # and add it to the swift info
+    symbol_graph = None
+    for dep in ctx.attr.deps[1:]:
+        if SwiftInfo in dep:
+            for module in dep[SwiftInfo].direct_modules:
+                if module.swift and module.swift.symbol_graph:
+                    symbol_graph = module.swift.symbol_graph
+                    break
+            if symbol_graph:
+                break
```
luispadron, line 535: Not sure this is totally correct (because it looks different than how we do it for the other fields), but it works for the examples I tried.

mattrobmattrob (approved):

rules/framework.bzl
```diff
@@ -522,12 +522,22 @@ def _copy_swiftmodule(ctx, framework_files):
     # only add a swift module to the SwiftInfo if we've actually got a swiftmodule
     swiftmodule_name = paths.split_extension(inputs.swiftmodule.basename)[0]
 
+    # find the first swift symbol_graph in the direct_modules of the deps of this target and forward it.
+    symbol_graph = None
+    for dep in ctx.attr.deps:
+        if SwiftInfo in dep and dep[SwiftInfo].direct_modules:
+            module = dep[SwiftInfo].direct_modules[0]
+            if module.swift and module.swift.symbol_graph:
+                symbol_graph = module.swift.symbol_graph
```
jerrymarino, line 531: Is there a reason that the symbol graph shouldn't be part of `inputs` and `outputs` / adjacent to the other components? The existing compiler artifacts are collected onto these struct(s) / follow this pattern and would suggest to keep the `.symbolgraph` similar to how we process `.swiftmodule` output

luispadron: The symbol graphs are added as extra outputs. They don't seem to be available in the rule last I checked

luispadron: Is there a way to access extra outputs from the rule? What you suggest is what I tried originally but the symbol graph wasn't in the list the rule gets. 

If there's a way to query the extra output groups though that would work

luispadron: Updated this to use `dep.output_groups`, in there it will have a `swift_symbol_graph` if the `--features=swift.emit_symbol_graph` is provided (which the .docc rule transitions on)

rules/framework.bzl
```diff
@@ -528,6 +538,7 @@ def _copy_swiftmodule(ctx, framework_files):
         swiftdoc = outputs.swiftdoc,
         swiftmodule = outputs.swiftmodule,
         swiftinterface = outputs.swiftinterface,
+        symbol_graph = outputs.symbol_graph,
```
jerrymarino, line 541: Changes LGTM 👍 .

Seems like plumbing in the `.symbolgraph` here good enough to get it correctly propagating end to end for you?

jerrymarino (approved):

rules/framework.bzl
luispadron: Yep this worked in testing, it correctly forwards the libraries symbol graphs (if they were requested)

thiagohmcruz:
Looks like this is not backwards compatible:
```sh
        File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 1042, column 40, in _apple_framework_packaging_impl
                swift_info = _get_merged_swift_info(ctx, framework_files, transitive_deps)
        File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 566, column 57, in _get_merged_swift_info
                swift_info_fields["modules"] = _copy_swiftmodule(ctx, framework_files)
        File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 537, column 52, in _copy_swiftmodule
                swift_module = swift_common.create_swift_module(
        File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_swift/swift/internal/providers.bzl", line 357, column 5, in create_swift_module
                def create_swift_module(
Error: create_swift_module() got unexpected keyword argument: symbol_graph
```

Was that intentional? We can bump `rules_swift` internally but curious if there's any suggestion on how to bump `rules_ios` with this change. I can also look into how to make this optional in `rules_ios` if that's what we want.

thiagohmcruz:
> Looks like this is not backwards compatible:
> 
> ```shell
> Traceback (most recent call last):
>         File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 1042, column 40, in _apple_framework_packaging_impl
>                 swift_info = _get_merged_swift_info(ctx, framework_files, transitive_deps)
>         File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 566, column 57, in _get_merged_swift_info
>                 swift_info_fields["modules"] = _copy_swiftmodule(ctx, framework_files)
>         File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_ios/rules/framework.bzl", line 537, column 52, in _copy_swiftmodule
>                 swift_module = swift_common.create_swift_module(
>         File "/private/var/tmp/_bazel_thiago/90ec49cd0a280be0e29f918d4d51c649/external/build_bazel_rules_swift/swift/internal/providers.bzl", line 357, column 5, in create_swift_module
>                 def create_swift_module(
> Error: create_swift_module() got unexpected keyword argument: symbol_graph
> ```
> 
> Was that intentional? We can bump `rules_swift` internally but curious if there's any suggestion on how to bump `rules_ios` with this change. I can also look into how to make this optional in `rules_ios` if that's what we want.

Nvm, we're internally using an old version of `rules_swift` overriding the version `rules_ios` uses so that's on us and I'll take a look!

PR #736, opened 2023-07-18
Make symbol_graph optional

thiagohmcruz:
For those on old `rules_swift` SHAs `symbol_graph` does not exist. This PR makes it "optional" to trigger the behaviour introduced in https://github.com/bazel-ios/rules_ios/pull/734.


thiagohmcruz:
Actually hold, we might not want this.

Issue #737, opened 2023-07-18
 Error: create_swift_module() got unexpected keyword argument: symbol_graph


jerrymarino:
Getting an error when I pull a SHA of rules_swift ` Error: create_swift_module() got unexpected keyword argument: symbol_graph`

```
[2023-07-17T13:18:06.292Z]     File "/private/var/tmp/_bazel_build/a38278faa597bf6d4e647970e120bcdb/external/build_bazel_rules_ios/rules/framework.bzl", line 1042, column 40, in _apple_framework_packaging_impl
[2023-07-17T13:18:06.292Z]         swift_info = _get_merged_swift_info(ctx, framework_files, transitive_deps)
[2023-07-17T13:18:06.292Z]     File "/private/var/tmp/_bazel_build/a38278faa597bf6d4e647970e120bcdb/external/build_bazel_rules_ios/rules/framework.bzl", line 566, column 57, in _get_merged_swift_info
[2023-07-17T13:18:06.292Z]         swift_info_fields["modules"] = _copy_swiftmodule(ctx, framework_files)
[2023-07-17T13:18:06.292Z]     File "/private/var/tmp/_bazel_build/a38278faa597bf6d4e647970e120bcdb/external/build_bazel_rules_ios/rules/framework.bzl", line 537, column 52, in _copy_swiftmodule
[2023-07-17T13:18:06.292Z]         swift_module = swift_common.create_swift_module(
[2023-07-17T13:18:06.292Z]     File "/private/var/tmp/_bazel_build/a38278faa597bf6d4e647970e120bcdb/external/build_bazel_rules_swift/swift/internal/providers.bzl", line 357, column 5, in create_swift_module
[2023-07-17T13:18:06.292Z]         def create_swift_module(
[2023-07-17T13:18:06.292Z] Error: create_swift_module() got unexpected keyword argument: symbol_graph
```

It seems like the call to `create_swift_module` needs to have affordances to deal with the older version of the API

jerrymarino:
Maybe we can vet the minimum SHA on CI on the [existing LTS](https://github.com/bazel-ios/rules_ios/blob/master/rules/repositories.bzl#L77) job to atleast guard against breaking compat

We're running:
```
github_repo(
    name = "build_bazel_rules_swift",
    project = "bazel-ios",
    ref = "e0272df7d98a563c07aa2e78722cd8ce62549864",
    repo = "rules_swift",
[...]

PR #738, opened 2023-07-18
Repro Error: create_swift_module() got unexpected keyword argument: sy…

jerrymarino:
…mbol_graph

Test Minimum commit vetted of rules_swift - not necessarily exclusive with LTS bazel version or LTS rules_apple version but tested on CI as so.

Relates to #737

### DIFF
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -64,17 +64,19 @@ def rules_ios_dependencies(
     Args:
         load_xchammer_dependencies: if `True` loads XCHammer and xcbuildkit dependencies (this is considered "alpha" software at the moment, see `README`)
     """
-    _maybe(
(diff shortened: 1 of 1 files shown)


justinseanmartin (commented):
Do we need to move the `build_bazel_rules_swift` from above the `if bazel_version.major == "5":` into the else block here?

thiagohmcruz (approved):
+1 to Justin's comment on moving the other `build_bazel_rules_swift` to the `else` block

jerrymarino:
Yep - we definitely need to do that 🏃‍♂️  

@thiagohmcruz we can figure out how to land the fix as a followup, I'm going to land this with the LTS error which was already broken and we can pu

PR #740, opened 2023-07-19
Introduce wrapped_bazelisk in tests and add docs for updating Xcode fixtures

thiagohmcruz:
The command suggested [here](https://github.com/bazel-ios/rules_ios/pull/664) is not working for me on Apple Silicon (had to fix missing double quotes there)
```sh
arch -arch x86_64 /bin/bash -l -c "./tests/xcodeproj-tests.sh --update"
```
and is updating fixtures that shouldn't be updated (`x86_64` is being updated to `arm64`), e.g.:
```sh
-                               HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-d31cf6d5fbab/bin/tests/ios/frameworks/objc/ObjcFramework_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-d31cf6d5fbab/bin/tests/ios/frameworks/objc/ObjcFramework_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
+                               HEADER_SEARCH_PATHS = "\"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-sim_arm64-min10.0-applebin_ios-ios_sim_arm64-dbg-ST-4994c467dfba/bin/tests/ios/frameworks/objc/ObjcFramework_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT/bazel-out/ios-sim_arm64-min10.0-applebin_ios-ios_sim_arm64-dbg-ST-4994c467dfba/bin/tests/ios/frameworks/objc/ObjcFramework_public_hmap.hmap\" \"$BAZEL_WORKSPACE_ROOT\"";
```

I believe that's happening because `arch-arch x86_64` does not propagate to all underlying `bazelisk` invocations. This PR "wraps" `bazelisk` to fix that. Specifically changes are:

* Add `tests/wrapped_bazelisk.sh` and `source` it where needed, `bazelisk` is conditionally wrapped if `x86_64` is not found (it will happen in both AS and in the `bazelisk` invocations we do in `genrule` targets to test project re-generation)
* Update `CONTRIBUTING.md` with instructions (caveats on `bazelisk` being `FAT` or not)

With the changes in this PR the command (no need to prefix with `arch`)
```sh
./tests/xcodeproj-tests.sh --update
```
should no-op if no other changes were made in the repo and update fixtures otherwise (on both Intel and AS).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,23 @@ If you are a maintainer of the repository and would like to tag a new release, f
 - Type in a version number, follow SemVer whenever possible.
 - After a few minutes the release should show up in the [Releases](https://github.com/bazel-ios/rules_ios/releases) page.
 
+## Updating Xcode fixtures
+
+Check the output of `find . -name '*.xcodeproj'` and note that there are many `.xcodeproj` fixtures used to ensure we catch regressions as changes that affect the Xcode project generator are made. If you made such changes and want to update the fixtures before opening a PR run:
+```sh
+./tests/xcodeproj-tests.sh --update
+```
+
+**IMPORTANT**: If you're on Intel the command above should generally "just work" but if you're on Apple Silicon you have to ensure your `bazelisk` installation is a `FAT` file so the script can select the `x86_64` architecture and generate the correct fixtures. In other words the command
+```sh
+lipo -info $(which bazelisk)
+```
+should return something like this
+```sh
+Architectures in the fat file: path/to/bazelisk are: x86_64 arm64
+```
+otherwise you won't be able to update fixtures locally.
+
 ## Bazel 6 & LTS Support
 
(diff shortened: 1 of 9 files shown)


thiagohmcruz:
Hold on, I think this was a mistake on my end and this PR might not be needed 🤦

jerrymarino:
@thiagohmcruz it's not very idiomatic to refer to it as `bazelisk` anyhow - we should call it `bazel` and allow the user to run whatever they want - or not use `bazelisk`

PR #741, opened 2023-07-19
Fix product_type when using `link_dynamic`

thiagohmcruz:
We are currently bundling dynamic frameworks with `product_type` = `apple_product_type.static_framework` due to [this default](https://github.com/bazel-ios/rules_ios/blob/master/rules/framework.bzl#L1212) and the fact that `rule_descriptor` [uses that same default](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/rule_support.bzl#L900).

This PR set's `product_type` explicitly for each `_bundle_dynamic_framework` and `_bundle_static_framework` and sets the correct respective `MACH_O_TYPE` in the Xcode project generator.

Before:
Screenshot 2023-07-18 at 4 52 49 PM

After:
Screenshot 2023-07-19 at 3 40 57 PM

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -625,7 +625,7 @@ def _attrs_for_split_slice(attrs_by_split_slices, split_slice_key):
     else:
         return attrs_by_split_slices[split_slice_key]
 
-def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
+def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps, product_type):
     """Packages this as dynamic framework
 
     Currently, this doesn't include headers or other interface files.
@@ -714,6 +714,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
 
     # TODO(jmarino) - consider how to better handle frameworks of frameworks
     processor_partials = []
+
     processor_partials.append(
         partials.apple_bundle_info_partial(
             actions = actions,
@@ -724,7 +725,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
             label_name = label.name,
             platform_prerequisites = platform_prerequisites,
             predeclared_outputs = predeclared_outputs,
-            product_type = rule_descriptor.product_type,
+            product_type = product_type if product_type else ctx.attr._product_type,
         ),
     )
     processor_partials.append(
@@ -921,7 +922,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
         ] + processor_result.providers,
     )
 
-def _bundle_static_framework(ctx, is_extension_safe, current_apple_platform, outputs):
+def _bundle_static_framework(ctx, is_extension_safe, current_apple_platform, outputs, product_type):
     """Returns bundle info for a static framework commonly used intra-build"""
     partial_output = partial.call(
         partials.extension_safe_validation_partial(
@@ -945,7 +946,7 @@ def _bundle_static_framework(ctx, is_extension_safe, current_apple_platform, out
             minimum_os_version = str(current_apple_platform.target_os_version),
             minimum_deployment_os_version = ctx.attr.minimum_deployment_os_version,
             platform_type = str(current_apple_platform.platform.platform_type),
-            product_type = ctx.attr._product_type,
+            product_type = product_type if product_type else ctx.attr._product_type,
             uses_swift = outputs.swiftmodule != None,
         ),
     ] + partial_output.providers)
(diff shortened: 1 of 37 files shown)


tools/xcodeproj_shims/install.sh
```diff
@@ -11,6 +11,9 @@ find "${BAZEL_DIAGNOSTICS_DIR}" -type f -atime +7d -delete
 case "${PRODUCT_TYPE}" in
 com.apple.product-type.framework)
     input_options=("bazel-bin/$BAZEL_BIN_SUBDIR/${TARGET_NAME}/${FULL_PRODUCT_NAME}")
+    if [ ! -d "$input_options" ]; then
```
thiagohmcruz, line 14: When `link_dynamic` is `True` it's placed under this path instead

rules/framework.bzl
```diff
@@ -724,7 +725,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
             label_name = label.name,
             platform_prerequisites = platform_prerequisites,
             predeclared_outputs = predeclared_outputs,
-            product_type = rule_descriptor.product_type,
+            product_type = product_type if product_type else ctx.attr._product_type,
```
thiagohmcruz, line 728: @jerrymarino let me know if my interpretation here is correct and we don't want static as a type here. I don't know all the details about `link_dynamic` don't wanted to make sure this change is consistent with the original intent of this flag.

rules/framework.bzl
```diff
@@ -625,7 +625,7 @@ def _attrs_for_split_slice(attrs_by_split_slices, split_slice_key):
     else:
         return attrs_by_split_slices[split_slice_key]
 
-def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
+def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps, product_type):
```
jerrymarino, line 628: In this function ` _bundle_dynamic_framework` - what is the use case to make it accept a `product` type that's not for a dynamic framework? The code is specific to packing a dynamic one

rules/framework.bzl
```diff
@@ -921,7 +922,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
         ] + processor_result.providers,
     )
 
-def _bundle_static_framework(ctx, is_extension_safe, current_apple_platform, outputs):
+def _bundle_static_framework(ctx, is_extension_safe, current_apple_platform, outputs, product_type):
```
jerrymarino, line 925: Similar comment as above, do we have a need to allow the function `_bundle_static_framework` to accept other product types than a static framework?

rules/framework.bzl
thiagohmcruz: Oh we could just set it directly internally without exposing as a param in this function indeed. I can update that.

rules/framework.bzl
thiagohmcruz: We don't have to, this was just for better readability and to be explicit about the type. Happy to remove and just let the default to be used here (which is static).

rules/framework.bzl
jerrymarino: It's probably fine to just hardcode this to the dynamic product type on this line specifically because it's only for `_bundle_dynamic framework`(s).

That said, we might need to contend with setting set `_product_type` argument too

jerrymarino: It's probably fine to just hardcode this to the dynamic product type on this line specifically because it's only for `_bundle_dynamic framework`(s).

That said, we might need to contend with setting set `_product_type` attr too?

jerrymarino (commented):
I'd suggest to remove the `_product_type` attr if we can too - that is always set to static ATM 🤔

rules/framework.bzl
thiagohmcruz: We can remove the default value from the private _product_type and the set it explicitly in the rule?

thiagohmcruz:
> I'd suggest to remove the `_product_type` attr if we can too - that is always set to static ATM 🤔 

Ok, I'll check if it can be removed without breaking things.

luispadron (approved):
+1 to some of the comments about having each of the static/dynamic macros just use `static` or `dynamic` directly as the product type.

We could also go one step further and fail if either is sent the wrong product type, it might have more easily caught this issue without having to look at the rule

thiagohmcruz:
> > > I'd suggest to remove the `_product_type` attr if we can too - that is always set to static ATM 🤔
> > 
> > 
> > Ok, I'll check if it can be removed without breaking things.
> 
> @jerrymarino this [call site](https://github.com/bazel-ios/rules_ios/blob/master/rules/framework.bzl#L651) in `rules_ios` depends on `rule_descriptor` [here](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/bundling_support.bzl#L56C24-L56C24) in `rules_apple`. So I don't think we can just remove `_product_type` easily. Instead I'll set its value to static/dynamic based on the value of `ctx.attr.link_dynamic`.

The problem with keeping it is that we would need two rules `apple_framework_packaging_static`/`apple_framework_packaging_dynamic` each one initializing `ctx.attr._product_type` differently. Similar to the [rule_factory](https://github.com/bazelbuild/rules_apple/blob/96fa885dd8310ddc65289e590f31d1f3b25f46d9/apple/internal/rule_factory.bzl#L1344-L1345) logic in `rules_apple`. And then places depending on this and assuming that the rule name is just `apple_framework_packaging` will definitely break...

rules/framework.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/741/commits/c209ea96a3ccd838c37a94cfc42a77345bb7a20f

rules/framework.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/741/commits/c209ea96a3ccd838c37a94cfc42a77345bb7a20f

rules/framework.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/741/commits/c209ea96a3ccd838c37a94cfc42a77345bb7a20f

thiagohmcruz:
@jerrymarino I'll handle `ctx.attr._product_type` in a follow up after we discuss a good approach here: https://github.com/bazel-ios/rules_ios/issues/742

For now I'll keep this PR as is, it fixes one of the bundling steps and does not change previous behaviour for the other ones. It's effectively one step towards https://github.com/bazel-ios/rules_ios/issues/742 so IMO it's worth landing as is so we can iterate on the rest.

PR #744, opened 2023-07-20
Remove invalid product type

### LINKED ISSUE #742: `ctx.attr._product_type` always being set to `apple_product_type.static_framework` in `apple_framework_packaging`
Due to [this default](https://github.com/bazel-ios/rules_ios/blob/39eb26647dab78ad9c93a909c5aa616b71bc7af0/rules/framework.bzl#L1212) some bundling steps of `apple_framework_packaging` are running with `product_type` = `apple_product_type.static_framework` always. Even when the intent is bundling a dynamic framework (i.e. `link_dynamic` is `True`).

https://github.com/bazel-ios/rules_ios/pull/741 helps with one of the steps and fixes `MACH_O_TYPE` in Xcode but we still rely on the hard coded value of `ctx.attr._product_type` in a few places:

* [bundle_name, bundle_extension](https://github.com/bazel-ios/rules_ios/blob/39eb26647dab78ad9c93a909c5aa616b71bc7af0/rules/framework.bzl#L651)
* [executable_name](https://github.com/bazel-ios/rules_ios/blob/39eb26647dab78ad9c93a909c5aa616b71bc7af0/rules/framework.bzl#L652C5-L652C20)
* [platform_prerequisites](https://github.com/bazel-ios/rules_ios/blob/39eb26647dab78ad9c93a909c5aa616b71bc7af0/rules/framework.bzl#L658C5-L658C27)

Maybe there's more but these are the ones I've found on a first scan. Essentially, if you follow the code paths above, any step that under the hood instantiates [rule_descriptor](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/rule_support.bzl#L894-L901) in `rules_apple`.

To fix this I see two paths (each with its own trade-offs):

1. Remove `ctx.attr._product_type` from `apple_framework_packaging`
    * The problem here is that the steps above that rely on `rule_descriptor` and hence `ctx.attr._product_type` being set will break so we'll have to get the same values some other way
2. Keep `ctx.attr._product_type` but conditionally instantiate `apple_framework_packaging_*` with different `product_type` values depending on `link_dynamic` being `True/False`
    * The problem here is that  we'll need a `rule_factory` approach [similar to rules_apple](https://github.com/bazelbuild/rules_apple/blob/96fa885dd8310ddc65289e590f31d1f3b25f46d9/apple/internal/rule_factory.bzl#L1344-L1362) and instantiate two rules `apple_framework_packaging_static` and `apple_framework_packaging_dynamic` and instantiate those using the factory [similar to what rules_apple does](https://github.com/bazelbuild/rules_apple/blob/96fa885dd8310ddc65289e590f31d1f3b25f46d9/apple/internal/ios_rules.bzl#L2092-L2113). That would break many things AFAICT, even if we use macro/alias to keep the name `apple_framework_packaging` and conditionally instantiate `_static`/`_dynamic` other rules/aspects/tools implemented by folks (in OSS or internally) would break due to the actual rule name changing and `apple_framework_packaging` not being a concrete rule anymore.

jerrymarino:
It was previously hardcoded as as static library and led to confusion even though it was mainly unused at the time.

Fixes #742

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -22,6 +22,7 @@ load("@build_bazel_rules_apple//apple/internal:resource_actions.bzl", "resource_
 load("@build_bazel_rules_apple//apple/internal:resources.bzl", "resources")
 load("@build_bazel_rules_apple//apple/internal:rule_support.bzl", "rule_support")
 load("@build_bazel_rules_apple//apple/internal:apple_toolchains.bzl", "AppleMacToolsToolchainInfo", "AppleXPlatToolsToolchainInfo")
+load("@build_bazel_rules_apple//apple/internal:swift_support.bzl", "swift_support")
 load("@build_bazel_rules_apple//apple/internal/utils:clang_rt_dylibs.bzl", "clang_rt_dylibs")
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "IosFrameworkBundleInfo")
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_clang_module_aspect", "swift_common")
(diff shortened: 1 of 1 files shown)


rules/framework.bzl
```diff
@@ -582,7 +582,7 @@ def _merge_root_infoplists(ctx):
     current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
     platform_type = str(current_apple_platform.platform.platform_type)
     apple_mac_toolchain_info = ctx.attr._toolchain[AppleMacToolsToolchainInfo]
-    rule_descriptor = rule_support.rule_descriptor(ctx)
+    rule_descriptor = rule_support.rule_descriptor_no_ctx(platform_type, apple_product_type.static_framework)
```
thiagohmcruz, line 585: Right, in my other PR I could've passed the attribute we need directly like this 🤦, nice one

thiagohmcruz (approved):
At a high level LGTM, only one concern. Are we confident this is not changing behaviour by potentially bundling things with different names in some scenarios that might not be covered on CI?

Specifically these values now map from
```py
    bundle_name, bundle_extension and executable_name <= bundling_support internal logic
```
to
```py
    bundle_name <= ctx.attr.framework_name
    bundle_extension <= ".framework"
    executable_name <= bundle_name
```

How to we know these names match before and after?

jerrymarino:
It'd be a decent amount of studying the code in general, but this is a lot more explicit and removes possible values for it which is a beneficial part of it too.

rules/framework.bzl
jerrymarino: NP @thiagohmcruz it just seemed like a small thing to fix up and make a bit easier

PR #745, opened 2023-07-20
Add ability to disable passing the -import-underlying-module copt to swift_library targets

thiagohmcruz:
**Note**: note that this is an optional behaviour so backwards compatible.

* Introduce `swift.swift_disable_import_underlying_module` so consumers can opt-in to not always pass `-import-underlying-module` to `swift_library` targets [here](https://github.com/bazel-ios/rules_ios/blob/cf7d493d9cdcff9671fe074c4cfcdcbe540b4cd4/rules/library.bzl#L896-L898)
* Because `additional_swift_copts` above is now a `select` [_append_headermap_copts](https://github.com/bazel-ios/rules_ios/blob/cf7d493d9cdcff9671fe074c4cfcdcbe540b4cd4/rules/library.bzl#L896-L898) can't mutate `swift_copts` via `extend` so adopting the `+= ` pattern there, did the same for the other params just for consistency and so that `_append_headermap_copts` returns all updated `*_copts` values instead of only `swift_copts`

My particular use case is "SwiftUI Previews" as Xcode will compile Swift files generated under DerivedData that are specific to "SwiftUI Previews" and if this flag is always passed it will get into the "Build for Previews" build in Xcode and fail to compile since there's no underlying ObjC module to load in that case.

This proposed change is generic though, and it could be useful to anyone looking to conditionally trigger this `-import-underlying-module` behaviour (e.g. it affects indexing as well).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -33,6 +33,13 @@ config_setting(
     },
 )
 
+config_setting(
+    name = "swift_disable_import_underlying_module",
+    values = {
+        "features": feature_names.swift_disable_import_underlying_module,
+    },
+)
+
 selects.config_setting_group(
     name = "arm64_simulator_use_device_deps",
     match_any = [
--- a/rules/features.bzl
+++ b/rules/features.bzl
@@ -7,4 +7,7 @@ feature_names = struct(
 
     # Use the ARM deps for the simulator - see rules/import_middleman.bzl
     arm64_simulator_use_device_deps = "apple.arm64_simulator_use_device_deps",
+
+    # When set disable passing the `-import-underlying-module` copt to `swift_library` targets
+    swift_disable_import_underlying_module = "swift.swift_disable_import_underlying_module",
 )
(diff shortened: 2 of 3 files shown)


jszumski (approved):

jerrymarino:
> My particular use case is "SwiftUI Previews" as Xcode will compile Swift files generated under DerivedData that are specific to "SwiftUI Previews" and if this flag is always passed it will get into the "Build for Previews" build in Xcode and fail to compile since there's no underlying ObjC module to load in that case.

Is it fair to say they don't need to `-import-underyling-module` anymore? Otherwise It sounds like you will cause `Build for Previews` to re-compile the transitive deps too - they won't cache hit disk or shared RBE either right?

thiagohmcruz:
> -im

We do rely on it, if we simply don't pass it even outside the context of "SwiftUI Previews", so normal builds, will fail both in `rules_ios` and internally for us.

I haven't spent time understanding why this flag seems to be so fundamental at the moment and I'd rather not assume we don't need it without a way to vet this assumption first

thiagohmcruz:
`Build for Previews` can't build with that flag set in the top level target being used for previews in Xcode, I'll investigate but even if this causes cache misses we can't fix that in 🔴 builds so we'll have to find another way

jerrymarino:
Ah makes sense - I have tested previews yet - but this seemed to have other compilations! Despite it the feature / PR is good in general 👍

PR #747, opened 2023-07-28
Pass xccurrentversions explicitly in AppleResourceInfo.unprocessed

thiagohmcruz:
#### Changes in this PR

* In `_precompiled_apple_resource_bundle` explicitly pass `.xccurrentversion` files to `AppleResourceInfo` provider in the `datamodels` attribute
* Add a test case to exercise `.xccurrentversion` being picked up by `rules_ios`'s Xcode project generator (just because this wasn't being exercise before, the previous item is not a prerequisite for this to work in the `rules_ios`'s generator):

Screenshot 2023-07-28 at 11 04 56 AM

#### Why

This is part of the work to make `rules_ios` work better with [rules_xcodeproj](https://github.com/MobileNativeFoundation/rules_xcodeproj) for those using it.

If one's `.xcdatamodeld` contains more than one version a `.xccurrentversion` file is used to set the current active one (represented by the ✅ in Xcode). When using `rules_xcodeproj` because `_precompiled_apple_resource_bundle` bundles just a `.plist` and a `.bundle` `.xccurrentversion` is not visible to `rules_xcodeproj` by the time it gets [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/xcodeproj/internal/resources.bzl#L62).

The issue with that is that if one's project contains two `Foo.xcdatamodel` and `Foo 2.xcdatamodel` and the `.xccurrentversion` points to `Foo 2.xcdatamodel` the generated Xcode project using `rules_xcodeproj` won't select the correct one and pick `Foo.xcdatamodel` instead since it couldn't collect the `.xccurrentversion` file. 

This will cause Xcode to mutate the `.pbxproj` and select `Foo.xcdatamodel` instea of `Foo 2.xcdatamodel` on launch causing an undesired non-empty `git diff`, plus the code itself will use the wrong version when building in Xcode.

While I was here added a test case to exercise the `.xccurrentversion` file in `rules_ios`'s Xcode project generator, this is just to exemplify how this change is not changing any behaviour but instead just making it easier to integrate with `rules_xcodeproj`.

#### FAQ

> Why this is not happening in `rules_ios`'s Xcode project generator?

`rules_ios`'s Xcode project generator was intentionally built to work with the rules and and the `.xccurrentversion` file is collected [here](https://github.com/bazel-ios/rules_ios/blob/master/rules/legacy_xcodeproj.bzl#L339).

> Why not fix this in `rules_xcodeproj` instead?

We're collaborating with the `rules_xcodeproj` maintainers to better integrate `rules_ios` and when a change is generic enough and applies to common use cases too (pure `rules_apple`, other custom rule sets, etc) and not only to `rules_ios` we're prioritizing making the changes in `rules_xcodeproj`.

I don't think the proposed change in this PR falls in the above category. `rules_ios` has its own approach for some things and it doesn't make sense to introduce special handling in `rules_xcodeproj` if it was a decision in `rules_ios` to diverge from what other rules are doing (e.g. not populating all fields in `AppleResourceInfo`, including the `datamodels` attr which is the one I'm updating here to pass the `.xccurrentversion` file).

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -196,15 +196,17 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         outputs = [output_bundle_dir],
     )
 
+    # See https://github.com/bazel-ios/rules_ios/pull/747 for context
     xccurrentversions = [
-        f
+        (None, None, depset([f]))
         for resource_files in ctx.attr.resources
         for f in resource_files.files.to_list()
-        if f.path.count("xccurrentversion")
+        if f.extension == "xccurrentversion"
     ]
 
     return [
         AppleResourceInfo(
+            datamodels = xccurrentversions,
             unowned_resources = depset(),
             owners = depset([
                 (output_bundle_dir.short_path, ctx.label),
@@ -220,9 +222,6 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             #    Foo.bundle directory that contains our real resources
             unprocessed = [
                 (output_bundle_dir.basename, None, depset([output_bundle_dir, output_plist])),
-                # Handles an edge case when using `rules_xcodeproj`
-                # TODO: Link the PR here
-                (None, None, depset(xccurrentversions)),
             ],
         ),
         AppleResourceBundleInfo(),
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -196,6 +196,13 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         outputs = [output_bundle_dir],
     )
 
+    xccurrentversions = [
+        f
+        for resource_files in ctx.attr.resources
+        for f in resource_files.files.to_list()
+        if f.path.count("xccurrentversion")
+    ]
+
     return [
         AppleResourceInfo(
             unowned_resources = depset(),
@@ -213,6 +220,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             #    Foo.bundle directory that contains our real resources
             unprocessed = [
                 (output_bundle_dir.basename, None, depset([output_bundle_dir, output_plist])),
+                (None, None, depset(xccurrentversions)),
             ],
         ),
         AppleResourceBundleInfo(),
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -220,6 +220,8 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             #    Foo.bundle directory that contains our real resources
             unprocessed = [
                 (output_bundle_dir.basename, None, depset([output_bundle_dir, output_plist])),
+                # Handles an edge case when using `rules_xcodeproj`
+                # TODO: Link the PR here
                 (None, None, depset(xccurrentversions)),
             ],
         ),
--- a/tests/ios/frameworks/core-data-resource-bundle/CoreDataExample/Model.xcdatamodeld/.xccurrentversion
+++ b/tests/ios/frameworks/core-data-resource-bundle/CoreDataExample/Model.xcdatamodeld/.xccurrentversion
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>_XCCurrentVersionName</key>
+    <string>Model 2.xcdatamodel</string>
+  </dict>
+</plist>
--- a/tests/ios/frameworks/core-data-resource-bundle/CoreDataExample/Model.xcdatamodeld/Model 2.xcdatamodel/contents
+++ b/tests/ios/frameworks/core-data-resource-bundle/CoreDataExample/Model.xcdatamodeld/Model 2.xcdatamodel/contents
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0"
+    lastSavedToolsVersion="16119" systemVersion="19G2021" minimumToolsVersion="Automatic"
+    sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="Person" representedClassName=".Person" syncable="YES">
+        <attribute name="name" optional="YES" attributeType="String" />
+        <attribute name="name2" optional="YES" attributeType="String" />
+    </entity>
+    <elements>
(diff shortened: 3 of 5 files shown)


mattrobmattrob:
I feel like the hacks or near hacks are mounting in the context of LTS, which is unfortunate. It'd be nice to move to a place where we could have "a supported rules_apple/rules_swift/apple_support version" that's pinned to a release, IMO. Perhaps an audit to really understand how much you all have in your LTS/forks would be nice since rules_iOS is inevitably pretty implementation detail dependent on them.

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -196,6 +196,13 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         outputs = [output_bundle_dir],
     )
 
+    xccurrentversions = [
+        f
+        for resource_files in ctx.attr.resources
+        for f in resource_files.files.to_list()
+        if f.path.count("xccurrentversion")
```
mattrobmattrob, line 203: Are you using this approach vs. endswith or in or similar for a specific reason?

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -213,6 +220,9 @@ def _precompiled_apple_resource_bundle_impl(ctx):
             #    Foo.bundle directory that contains our real resources
             unprocessed = [
                 (output_bundle_dir.basename, None, depset([output_bundle_dir, output_plist])),
+                # Handles an edge case when using `rules_xcodeproj`
+                # TODO: Link the PR here
```
mattrobmattrob, line 224: Reminder to put PR link.

mattrobmattrob (approved):

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: nope, actually `endswith` or `extension` is probably a better check, will update

thiagohmcruz:
> I feel like the hacks or near hacks are mounting in the context of LTS, which is unfortunate. It'd be nice to move to a place where we could have "a supported rules_apple/rules_swift/apple_support version" that's pinned to a release, IMO. Perhaps an audit to really understand how much you all have in your LTS/forks would be nice since rules_iOS is inevitably pretty implementation detail dependent on them.

We've been getting this feedback relatively often and I agree with you. I think it would require some planning and a detailed audit with tasks tracked in an "epic"-like issue. Long term ideally we would not need LTS once we manage to always run on the latest releases.

thiagohmcruz:
To be transparent I'm open to follow a diff approach here, keeping it as a draft while I collect feedback from folks

thiagohmcruz:
Actually I might have found a way to  `.xccurrentversion` to `datamodels`, will run some tests and update this PR.

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/747/commits/2eb797e05ba9f2416b6b3a8f7923c7b5d041bebc

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/747/commits/2eb797e05ba9f2416b6b3a8f7923c7b5d041bebc

justinseanmartin (approved):

thiagohmcruz:
Pushed a change using `datamodels` [here](https://github.com/bazel-ios/rules_ios/pull/747/commits/2eb797e05ba9f2416b6b3a8f7923c7b5d041bebc) and updated the PR description to remote the `datamodels` related topics since that's now resolved.

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -196,8 +196,17 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         outputs = [output_bundle_dir],
     )
 
+    # See https://github.com/bazel-ios/rules_ios/pull/747 for context
+    xccurrentversions = [
+        (None, None, depset([f]))
```
mattrobmattrob, line 201: Is there only ever going to be a single xccurrentversion? I'm not too familiar with this part of CoreData but just trying to think through a more efficient depset creation if necessary.

thiagohmcruz: AFAIK yes that file is unique under one `.xcdatamodeld` directory, I couldn't find any docs confirming that though, just based on my experience. It would be hard to know what file determines the current version if there was more than one I think?

thiagohmcruz: The "plural" pattern here is just to be consistent with the type list, but I could make it be a single file indeed. Suggestions welcomed!

jerrymarino:
Is it accurate to say the point of this thread is to make  `rules_xcodeproj` work with Bazel 5? I updated to Bazel 6.x.x today does it go away?

AFAIK the community still does use rules_ios use the LTS release and Bazel 5.x.x with rules_ios. IME good to align with Bazel's LTS vision - that said if I'm the last user of 5.x.x I'm ok to part ways with it.  On a side note, Bazel 6.x.x is finally looking like a reality, but I have blocking issues for 7 would need LTS support for 6.x.x until I can fix those 🙃

jerrymarino (approved):
Thanks for adding the test case, not sure where the LTS version comes into play here but would like to know

thiagohmcruz:
> Is it accurate to say the point of this thread is to make `rules_xcodeproj` work with Bazel 5? If I updated to Bazel 6.x.x today does it go away?

@jerrymarino Not exactly, the fact that `.xccurrentversion` is not present in `AppleResourceInfo` => `datamodels` would also cause issues in Bazel 6. Previously I couldn't find a way to set in `datamodels` without hitting a build failure in `rules_apple`. But after the latest commit I shared above I was able to do just that and set it in `datamodels` so this change now is just passing more info via `AppleResourceInfo` in the correct way (i.e. data model files set in `datamodels` attribute) and it has an additional benefit of fixing an issue with `rules_xcodeproj` I've found (detailed in the PR descr).

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: I can also `fail` if more than one is found and then access index `[0]` and call it `xccurrentversion` instead of `xccurrentversions`. 

I honestly don't have a preference here, let me know if I should treat this as a nit or not. Otherwise will land this soon!

jerrymarino:
Sure, the PR seems reasonable in general if it helps fix that 👍

rules/precompiled_apple_resource_bundle.bzl
mattrobmattrob: Nah, all good on my side. I was just thinking passing the list to a single depset instead of N depset but I might be misunderstanding the datamodels input.

thiagohmcruz: Landing, happy to nit in a follow up!

thiagohmcruz:
I might have found a bug here... preping a PR

PR #748, opened 2023-07-28
Set xccurrentversion owners

thiagohmcruz:
Follow up to https://github.com/bazel-ios/rules_ios/pull/747

While testing this in a more complex graph found that I've missed setting `owners` for the newly collected `.xccurrentversion` files.

@mattrobmattrob re our other thread on that, also learned that we can hit this code path with multiple `.xccurrentversion` files under the same target even though they're unique in their own `.xcdatamodeld` directories.

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -207,8 +207,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
     return [
         AppleResourceInfo(
             datamodels = [
-                (None, None, depset([f]))
-                for f in xccurrentversions
+                (None, None, depset(xccurrentversions)),
             ],
             unowned_resources = depset(),
             owners = depset(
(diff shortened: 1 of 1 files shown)


justinseanmartin (approved):

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -198,20 +198,28 @@ def _precompiled_apple_resource_bundle_impl(ctx):
...
         for f in resource_files.files.to_list()
         if f.extension == "xccurrentversion"
     ]
 
     return [
         AppleResourceInfo(
-            datamodels = xccurrentversions,
+            datamodels = [
+                (None, None, depset([f]))
+                for f in xccurrentversions
```
mattrobmattrob, line 211: And this is meant to not be like this (what I poorly asked in the past PR), correct?
```suggestion
                (None, None, depset(xccurrentversions))
```

mattrobmattrob (approved):

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: Oh I think it will be fine if it's written like this, and I def didn't have to write one `depset` per file 🤦. Will run a quick test and pull this in, thx for catching!

thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/748/commits/94621fb0eeab1c4b2f01f30789c0127ea4729ccd

PR #750, opened 2023-08-02
Collect all .xcdatamodeld instead of only .xccurrentversion

thiagohmcruz:
Follow up to https://github.com/bazel-ios/rules_ios/pull/747 and https://github.com/bazel-ios/rules_ios/pull/748.

Noticed that some files were not being copied into the bundle correctly if we collect only the `.xccurrentversion` so collecting all files under the `.xcdatamodeld` directory instead.

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -197,17 +197,17 @@ def _precompiled_apple_resource_bundle_impl(ctx):
     )
 
     # See https://github.com/bazel-ios/rules_ios/pull/747 for context
-    xccurrentversions = [
+    datamodel_files = [
         f
         for resource_files in ctx.attr.resources
         for f in resource_files.files.to_list()
-        if f.extension == "xccurrentversion"
+        if f.path.count(".xcdatamodeld")
     ]
 
     return [
         AppleResourceInfo(
             datamodels = [
-                (None, None, depset(xccurrentversions)),
+                (None, None, depset(datamodel_files)),
             ],
             unowned_resources = depset(),
             owners = depset(
@@ -216,7 +216,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
                     (output_plist.short_path, ctx.label),
                 ] + [
                     (f.short_path, ctx.label)
-                    for f in xccurrentversions
+                    for f in datamodel_files
                 ],
             ),
             # This is a list of the resources to propagate without changing further


justinseanmartin (approved):

mattrobmattrob (approved):

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -161,6 +161,17 @@ def _precompiled_apple_resource_bundle_impl(ctx):
...
+    # ps: removing `.momd` files from here since they'll be collected in `AppleResourceInfo` => `datamodels`
+    # below and added under the `.bundle` directory like other resources. This is so data model files
+    # are visible to other rules looking for information in the `AppleResourceInfo` provider. See this
+    # PR for more context: https://github.com/bazel-ios/rules_ios/pull/750
+    control_files = [
+        f
+        for f in control_files
+        if hasattr(f, "src") and not f.src.endswith(".momd")
+    ]
+
```
justinseanmartin, line 174: NIT: Feels like the comments are out of order now and could be a bit more terse. Also, bikeshedding a bit, I think the PR information can generally be left out, as it can be traced back from git commit metadata when useful, but usually doesn't add much value and reduced readability.

```suggestion
    # `*.momd` files will be collected in `AppleResourceInfo` below and added under the `.bundle` directory like other resources.
    control_files = [
        f
        for f in control_files
        if hasattr(f, "src") and not f.src.endswith(".momd")
    ]

    # Create a file for bundletool to know what files to copy
    # https://github.com/bazelbuild/rules_apple/blob/d29df97b9652e0442ebf21f1bc0e04921b584f76/tools/bundletool/bundletool_experimental.py#L29-L46
```

justinseanmartin (approved):
Still looks reasonable to me, as long as it seems to clear up the reported issue and still fixes the original issue this was attempting to fix.

rules/precompiled_apple_resource_bundle.bzl
thiagohmcruz: Makes sense to me, def being more verbose here than necessary. These changes are still not right though and I'll continue to investigate. The failures in the latest run show me that something is missing since a test that was passing before is not anymore.

Will make the changes once I root cause that.

thiagohmcruz:
... this is not right. This is changing things in the bundle still and that's not the intent. This test
```
bazel test //tests/ios/frameworks/core-data-resource-bundle:CoreDataExampleTests
```
was 🟢 before and is 🔴 in this PR. 

I'll revisit my approach for this `.xccurrentversion` story and propose a better change or consider fixing this directly in `rules_xcodeproj`. Closing for now and sorry for the noise 🤦‍♂️.

PR #759, opened 2023-08-31
Update to rules_apple 3.0.0

jszumski:
## Changes
- **Uses new provider initialization pattern**
This is now private API in rules_apple as of https://github.com/bazelbuild/rules_apple/pull/2184 but visibility was relaxed in https://github.com/bazelbuild/rules_apple/pull/2225 to avoid breaking us.

- **Drops support for i386 slices in xcframeworks**
Dropped upstream in https://github.com/bazelbuild/rules_apple/pull/2155.

- **Passes through `minimum_os_version` to all test rules**
Required upstream in https://github.com/bazelbuild/rules_apple/pull/2124.

- **Uses new split transition**
Core's `apple_common.multi_arch_split` was removed in Bazel 7 and replaced by rules_apple's `apple_platform_split_transition`: https://github.com/bazelbuild/rules_apple/pull/2034

## Remaining work
- [ ] ~Update project generation rules~
- [x] Run buildifier
- [x] Decide if we should try to make these changes backwards-compatible **no**
- [x] Decide if this is a good time to drop Bazel 5 support **yes**

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -22,8 +22,10 @@ load("@build_bazel_rules_apple//apple/internal:resources.bzl", "resources")
 load("@build_bazel_rules_apple//apple/internal:rule_support.bzl", "rule_support")
 load("@build_bazel_rules_apple//apple/internal:apple_toolchains.bzl", "AppleMacToolsToolchainInfo", "AppleXPlatToolsToolchainInfo")
 load("@build_bazel_rules_apple//apple/internal:swift_support.bzl", "swift_support")
+load("@build_bazel_rules_apple//apple/internal:transition_support.bzl", rules_apple_transition_support = "transition_support")
 load("@build_bazel_rules_apple//apple/internal/utils:clang_rt_dylibs.bzl", "clang_rt_dylibs")
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "IosFrameworkBundleInfo")
+load("@build_bazel_rules_apple//apple/internal:providers.bzl", "new_applebundleinfo", "new_iosframeworkbundleinfo")
 load("@build_bazel_rules_swift//swift:swift.bzl", "SwiftInfo", "swift_clang_module_aspect", "swift_common")
 load(
     "@build_bazel_rules_apple//apple/internal/aspects:resource_aspect.bzl",
@@ -582,7 +584,14 @@ def _merge_root_infoplists(ctx):
     current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
     platform_type = str(current_apple_platform.platform.platform_type)
     apple_mac_toolchain_info = ctx.attr._toolchain[AppleMacToolsToolchainInfo]
-    rule_descriptor = rule_support.rule_descriptor_no_ctx(platform_type, apple_product_type.static_framework)
+    rule_descriptor = rule_support.rule_descriptor(
+        platform_type = platform_type,
+        product_type = apple_product_type.static_framework,
+    )
+    features = features_support.compute_enabled_features(
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
 
     resource_actions.merge_root_infoplists(
         actions = ctx.actions,
@@ -596,7 +605,7 @@ def _merge_root_infoplists(ctx):
         output_plist = output_plist,
         output_pkginfo = None,
         output_discriminator = "framework",
-        platform_prerequisites = _platform_prerequisites(ctx, rule_descriptor, platform_type),
+        platform_prerequisites = _platform_prerequisites(ctx, rule_descriptor, platform_type, features),
         resolved_plisttool = apple_mac_toolchain_info.resolved_plisttool,
         rule_descriptor = rule_descriptor,
         rule_label = ctx.label,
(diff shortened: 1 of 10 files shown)


rules/framework.bzl
```diff
@@ -25,6 +25,7 @@ load("@build_bazel_rules_apple//apple/internal:apple_toolchains.bzl", "AppleMacT
 load("@build_bazel_rules_apple//apple/internal:swift_support.bzl", "swift_support")
 load("@build_bazel_rules_apple//apple/internal/utils:clang_rt_dylibs.bzl", "clang_rt_dylibs")
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleBundleInfo", "IosFrameworkBundleInfo")
+load("@build_bazel_rules_apple//apple/internal:providers.bzl", "new_applebundleinfo")
```
jerrymarino, line 28: Do we need to rely on yet another internal provider from rules_apple here? It seems like it would continue to work well on 5.x.x pretty easily w/o the change?

jerrymarino: Also the issue with this line is everyone will have to run a commit later than:
https://github.com/bazelbuild/rules_apple/pull/2194

jszumski: It's incompatible because the initializer intentionally fails here in `_make_banned_init`: ```AppleBundleInfo, new_applebundleinfo = provider(
    doc = ...,
    fields = {
        ...
    },
    init = _make_banned_init(provider_name = "AppleBundleInfo"),
)```

Keeping `rules_ios` close to the head of `rules_apple` was intentional with this change.

rules/library.bzl
```diff
@@ -344,8 +344,10 @@ def _xcframework(*, library_name, name, slices):
 
         for arch in archs:
             if platform == "ios":
-                if (arch == "armv7s" or arch == "arm64e"):
-                    # unsupported platform-arch by rules_apple
+                # unsupported platform-arch by rules_apple
+                unsupported_platforms = ["armv7", "armv7s", "i386"]
```
jszumski, line 348: Dropping `arm64e` was intentional here since it exists in rules_apple

rules/framework.bzl
jerrymarino: Yeah what I'm wondering, is how to make it work backwards with older versions:

```
def _new_apple_ctx():
```

Though the initializer _moving to_ a private API which used to be public is another story 😮‍💨

jerrymarino (commented):
I'm going to review this update and see what it takes to support older versions.

rules/library.bzl
jerrymarino: Does this works on 2.5?

rules/framework.bzl
```diff
@@ -664,8 +675,11 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
     resource_deps = ctx.attr.deps + ctx.attr.transitive_deps + ctx.attr.data
     current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
     platform_type = str(current_apple_platform.platform.platform_type)
-    rule_descriptor = rule_support.rule_descriptor_no_ctx(platform_type, apple_product_type.framework)
-    platform_prerequisites = _platform_prerequisites(ctx, rule_descriptor, platform_type)
+    rule_descriptor = rule_support.rule_descriptor(
```
jerrymarino, line 678: Any way to plumb in the `rule_descriptor` from above?

rules/framework.bzl
```diff
@@ -694,6 +708,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
         ctx,
         avoid_deps = avoid_deps,
         entitlements = None,
+        exported_symbols_lists = ctx.files.exported_symbols_lists,
```
jerrymarino, line 711:
```
# If this isn't backwards compatible with 2.5.0 
if rules_apple_api_version_3:
    kwargs["exported_symbols_lists"] = ctx.files.exported_symbols_lists
else:
    kwargs[" executable_name"] = bundle_name
```

rules/framework.bzl
```diff
@@ -1053,23 +1073,23 @@ apple_framework_packaging = rule(
         ),
         "deps": attr.label_list(
             mandatory = True,
-            cfg = apple_common.multi_arch_split,
+            cfg = rules_apple_transition_support.apple_rule_transition,
```
jerrymarino, line 1076: Possible to add conditional assignment at the top of `framework.bzl` or in some other file? e.g. :
```
if rules_apple_api_version_3:
    apple_dep_transition = rules_apple_transition_support.apple_rule_transition
else:
    apple_dep_transition = apple_common.multi_arch_split
```

Then - consuming it here:
```
cfg = apple_dep_transition
```

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -52,6 +53,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
 
     platform_prerequisites = platform_support.platform_prerequisites(
         apple_fragment = ctx.fragments.apple,
+        build_settings = None,
```
jerrymarino, line 56: May or many not need fixing if it passes CI with 2.5 / LTS enabled

tools/vmd/vm_test_runner.bzl
```diff
@@ -19,7 +19,7 @@
 """iOS test runner rule."""
 
 load(
-    "@build_bazel_rules_apple//apple/testing:apple_test_rules.bzl",
```
jerrymarino, line 22: Might need other way to load it

jerrymarino (commented):
It seems like with few quick changes there isn't a reason it wouldn't support 2.5.x by making some of these conditional. Likely this can be done with wrappers to further centralize the calling of code from it to make it and clean it up.

tools/vmd/vm_test_runner.bzl
jszumski: Can you conditionalize load statements? I didn't think so but maybe I'm wrong

jerrymarino: Yeah sure - we can unify the sims in a starlark file we load in “rules ios  dependencies” - based on the api version.

jerrymarino: Shims*

jerrymarino:
@jszumski last night I pulled this down and looked at at bit. It seems like there are 2 distinct issues here:
1. making rules_ios work correctly with rules_apple 3.0.0 - this is easier said than done and the errors you have in the CI job are the tip of the iceberg. I'm trying to find the simplest way for to unblock you and others.  

2. cleanup and refactoring some stuff to better support pt1 and also consequently older rules_apple.

jerrymarino:
Submitting and end to end implementation of this.

PR #761, opened 2023-09-05
Using target name in bundle's plist path to avoid file writing clashes of Info.plist-root-control

tymurmustafaiev:
When I want to build 2 frameworks with same `module_name` `bazelisk build //Path/To/Target1 //Path/To/Target2`, there is an error thrown which says about clashing writing actions to the same destination: `/Path/ModuleName/Info.plist-root-control`.
In this PR, I've changed `module_name` from the path of the merging Info.plist to `target_name` so we could avoid this clash.
So the output would be:
`/Path/Target1/Info.plist-root-control` and `/Path/Target2/Info.plist-root-control`


tymurmustafaiev:
I'll push better solution

PR #762, opened 2023-09-06
Drops support for xchammer and xcbuildkit

jszumski:
Drops xchammer in favor of `legacy_xcodeproj` rule.  The goal is to clean up the unused portions of the repo to facilitate an update to Bazel 6, support for bzlmod, and newer rules_apple releases.


thiagohmcruz (approved):
XCHammer is in a "WIP" state for a while since I added these and most of my time now is focused on integrating with `rules_xcodeproj`. So it does make sense to me to drop these since AFAIK no one else was using it with `rules_ios` besides us (for readers, let me know if that's not the case and we can bring it back).

That said, such a WIP/experimental project should def not block things like bumping Bazel and `rules_apple` or adding support for new features that are getting traction like bzlmod. So I'm in favour of removing it.

luispadron (approved):

luispadron:
This greatly simplifies bzlmod migration

mattrobmattrob (approved):

jerrymarino:
Well that one landed pretty fast - what about this was actually broken @jszumski ?

jerrymarino:
In this PR description, I'd like to also know more about what you mean by this:
> facilitate an update to Bazel 6, 

We have used Bazel 6.0 in rules_ios to 6.0 about a year ago and didn't have an issue with it then either

jszumski:
Poorly worded... that was mean to convey moving the minimum supported version to 6.

PR #764, opened 2023-09-06
Drop Bazel 5 support

jszumski:
Removes the Bazel 5 repository pins and CI job now that all users have upgraded to Bazel 6.

Depends on https://github.com/bazel-ios/rules_ios/pull/762


luispadron (approved):
Wooo!

mattrobmattrob (approved):
Good with me as long as @thiagohmcruz and team give the 👍

thiagohmcruz:
Running some tests, will review soon!

rules/repositories.bzl
```diff
@@ -57,49 +57,23 @@ def _get_bazel_version():
 
 def rules_ios_dependencies():
     """Fetches repositories that are dependencies of `rules_ios`"""
-    bazel_version = _get_bazel_version()
-    if bazel_version.major == "5":
```
thiagohmcruz, line 61: cc @jerrymarino any concerns with the removal?

thiagohmcruz (approved):
@jszumski If LTS + Bazel 5.x is gone we should probably also update https://github.com/bazel-ios/rules_ios/blob/master/CONTRIBUTING.md?

jerrymarino (approved):
> that all users have upgraded to Bazel 6.

Curious to know what you are basing that off of. AFAIK the last issues for Bazel 6.0 are worked around so this can fall back to the [LTS policy](https://github.com/bazel-ios/rules_ios/pull/764/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L43) if someone wants to revert this / needs it back.

We don't really keep a close pulse on all the people using it so will wait to see how it goes.

jszumski:
> Curious to know what you are basing that off of. 

Slack and some asking around.  Bazel 5 folks can remain on 2.1.0 though.

jerrymarino:
I'm still sorting out some edge cases in Bazel 6.x.x. and would prefer to hold this till EOY, I'm still wondering what the adoption looks like on that verison outside of the folks who contribute here. Th

jerrymarino:
Per Bazel this is supported till 2025 so closing for now.

jerrymarino:
LTSReleaseSchedule

* per the [LTS release schedule](https://bazel.build/release)

luispadron:
> LTSReleaseSchedule
> * per the [LTS release schedule](https://bazel.build/release)

Can you explain why that should affect the minimum version we support in rules_ios 3.x?
By this logic we should be supporting Bazel 4 as well.

jerrymarino:
In practice it takes people different times to update and the changes in rules_ios hasn't ever been a significant burden for us. [The LTS README entry](https://github.com/bazel-ios/rules_ios/pull/764/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L23) covers some of our experiences maintaining it and keeping it up to date.

luispadron:
> In practice it takes people different times to update and the changes in rules_ios hasn't ever been a significant burden for us. [The LTS README entry](https://github.com/bazel-ios/rules_ios/pull/764/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L23) covers some of our experiences maintaining it and keeping it up to date. 

Maintaining those two Bazel versions (and now 2 rules Apple versions) does make it harder to contribute. And the proposed changes to support rules_apple 2 and 3 would block and make it more difficult to support bzlmod. 

I still don't understand why we can't use releases correctly. Bazel 5-6 changes can continue to be made to versions 2.x and Bazel 6+/rules_apple 3+ can go into 3.x

jerrymarino:
>Maintaining those two Bazel versions (and now 2 rules Apple versions) does make it harder to contribute. 


"it mades it harder to contribute to" is an interesting statement as most of the contributions came from people using Bazel 5, but these contributions also worked seamlessly with Bazel 6.

karim-alweheshy:
I would like to re-visit this issue again now we are trying to support bazel 7 as well

The main challenge here, would be supporting multiple variations of legacy ObjC while upgrading to CcInfo providers to compiling and linking

Also rules_apple 2.x and 3.x do not support bazel 5

luispadron:
Bazel 7 is the latest LTS, that means Bazel 6 and 7 are the two versions we'll support going forward.

We'll work on bringing this PR back up to date

thiagohmcruz:
IMO as long as we make some effort to communicate this to consumers (e.g. Bazel slack at least, maybe we should start using GC discussions too) that this is happening it does make sense to reduce maintenance burden by putting energy into supporting only the two latest versions (i.e. 6 and 7).

As discussed above, worst case folks on 5.x can always use older releases until they're ready to bump to 6 or 7.

Should we bring this PR back to a reviewable state in case it's not already?

luispadron:
Agreed, we can definitely communicate it and we did when it was originally proposed.
Folks who responded in Slack said Bazel 6+

John and/or I can work on getting this rebased and updated (or creating a new PR)

justinseanmartin:
With Bazelmod, being the new way to consume things, it seems like [this](https://github.com/bazel-ios/rules_ios/blob/4d9a2d003b771a906336b995f61d2817f87baf97/MODULE.bazel#L11) is the defacto place to define Bazel version support. Presumably things should "just work" for folks if they're still on Bazel 5 to get the older published rules_ios versions that support it.

To be explicit, we'll be losing some ability for folks to incrementally bump components independently. That said, I think it is a reasonable tradeoff to lessen work (support/testing/complexity) on the part of maintainers, especially in light of not having any active contributors known to still be on Bazel 5. Just because Google continues to support Bazel 5 doesn't mean every ruleset needs to IMO.

luispadron:
Superseded by: https://github.com/bazel-ios/rules_ios/pull/834

PR #765, opened 2023-09-12
[DNR] Move to patched_imports

jerrymarino:


rules/library.bzl
```diff
@@ -3,7 +3,7 @@
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:sets.bzl", "sets")
 load("@bazel_skylib//lib:selects.bzl", "selects")
-load("@build_bazel_rules_apple//apple:apple.bzl", "apple_dynamic_framework_import", "apple_static_framework_import")
+load("//rules:apple_patched.bzl", "apple_dynamic_framework_import", "apple_static_framework_import")
```
thiagohmcruz, line 6: Any chance this could slow us down in the future when trying to bump `rules_apple` to pick up changes that affect these symbols?

luispadron (commented):
Some more info on why this change is required would be great

rules/library.bzl
jerrymarino: I don't think so, we've been running with this file pretty much since the beginning of the repo.

justinseanmartin: Is the `apple_patched` behavior used anywhere?  It looks to me like it is dead code after we dropped Carthage/CocoaPods pre-building dependencies support. I don't think we ever exercised this code path in our consumption of rules_ios IIUC.

thiagohmcruz: Indeed AFAICT this was not being exercised? More details on your intent with this change would help.

jerrymarino: You've got it - I will followup with this PR in a few days, we are working through some updates to get 6.x.x really working well.

PR #767, opened 2023-09-14
[Bazel 6.x.x] sim_arm64 middleman fixes

jerrymarino:
In Bazel 6.x.x the `static_framework_file` was migrated onto `imported_library` this PR completes the migration and is followup to the Bazel 6 work.

The related rules_apple change which updated `static_framework_file`: https://github.com/bazelbuild/rules_apple/commit/8d841342c238457896cd7596cc29b2d06c9a75f0

### DIFF
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -1,3 +1,8 @@
+# Notes on Bazel 5.x.x VS 6.x.x and rules_apple rule_api_version changes:
+# The provider field that contains the `static_framework_file` changed to
+# `imported_library` - see
+# https://github.com/bazelbuild/rules_apple/commit/8d841342c238457896cd7596cc29b2d06c9a75f0
+
 load("@build_bazel_rules_apple//apple:providers.bzl", "AppleFrameworkImportInfo")
 load("//rules:features.bzl", "feature_names")
 load("//rules/internal:objc_provider_utils.bzl", "objc_provider_utils")
@@ -100,8 +105,6 @@ def _find_imports_impl(target, ctx):
         if use_lts_5_rules_apple_api:
             static_framework_file.append(target[apple_common.Objc].static_framework_file)
         else:
-            # The provider field that contains the `static_framework_file` changed in
-            # https://github.com/bazelbuild/rules_apple/commit/8d841342c238457896cd7596cc29b2d06c9a75f0
             static_framework_file.append(target[apple_common.Objc].imported_library)
 
         target_dynamic_framework_file = target[apple_common.Objc].dynamic_framework_file
@@ -217,7 +220,6 @@ def _file_collector_rule_impl(ctx):
     if use_lts_5_rules_apple_api:
         objc_provider_fields["static_framework_file"] = depset(replaced_static_framework.inputs)
     else:
-        # See comment in file above regarding this name in 6.x.x
         objc_provider_fields["imported_library"] = depset([], transitive = [
             depset(replaced_static_framework.inputs),
             objc_provider_fields.get("imported_library", depset([])),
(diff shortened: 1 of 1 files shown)


rules/import_middleman.bzl
```diff
@@ -210,8 +210,18 @@ def _file_collector_rule_impl(ctx):
...
     deduped_static_framework = depset(_deduplicate_test_deps(test_linker_deps[0], existing_static_framework.to_list()))
+
     replaced_static_framework = _replace_inputs(ctx, deduped_static_framework, input_static_frameworks, _update_framework)
-    objc_provider_fields["static_framework_file"] = depset(replaced_static_framework.inputs)
+    rules_apple_api_version = getattr(bundling_support, "rule_api_version", None)
+    use_lts_5_rules_apple_api = rules_apple_api_version == 1.0
+    if use_lts_5_rules_apple_api:
+        objc_provider_fields["static_framework_file"] = depset(replaced_static_framework.inputs)
+    else:
+        # See comment in file above regarding this name in 6.x.x
```
jszumski, line 220: Is this referring to L103? It might be more straightforward to repeat it here

rules/import_middleman.bzl
```diff
@@ -239,11 +249,17 @@ def _file_collector_rule_impl(ctx):
             dynamic_framework_dirs.append(f)
     objc_provider_fields["dynamic_framework_file"] = depset(dynamic_framework_file)
 
-    replaced_frameworks = replaced_dyanmic_framework.values() + replaced_static_framework.replaced.values()
+    all_replaced_frameworks = replaced_dyanmic_framework.values() + replaced_static_framework.replaced.values()
+
+    # This slightly varies in Bazel 6.x.x
```
jszumski, line 254: Can you expand this comment to explain what is different between the versions and needs unique handling for each?

rules/import_middleman.bzl
```diff
@@ -239,11 +249,17 @@ def _file_collector_rule_impl(ctx):
...
+    if use_lts_5_rules_apple_api:
+        replaced_frameworks = all_replaced_frameworks
+    else:
+        replaced_frameworks = replaced_dyanmic_framework.values()
 
     compat_link_opt = ["-L__BAZEL_XCODE_DEVELOPER_DIR__/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphonesimulator", "-Wl,-weak-lswiftCompatibility51"]
 
-    if len(replaced_frameworks):
+    if len(all_replaced_frameworks):
```
jszumski, line 262: Is it intentional that `linkopt` below is still using `replaced_frameworks` instead of `all_replaced_frameworks`?

```
["\"\"\"-F" + "/".join(f.path.split("/")[:-2]) + "\"\"\"" for f in replaced_frameworks] + compat_link_opt,
```

rules/import_middleman.bzl
thiagohmcruz: @jerrymarino correct me if I'm wrong but I believe what this is doing is handling only dynamic fmws in Bazel 6 since static fmws are handled here: https://github.com/bazel-ios/rules_ios/pull/766. If that's the case a small comment with this context would be preferred IMO.

rules/import_middleman.bzl
jerrymarino: That is the gist of it -  6.x.x `static_framework` was transitioned to `imported_library` and is in the `filelist` not here

rules/import_middleman.bzl
jerrymarino: Yes, I have added a `Notes` segment to the top of the doc WRT this: https://github.com/bazel-ios/rules_ios/pull/767/files#diff-e213489d3a65f173ffa46728826b06f786a5d63666e3fe7101493397ee12f948R2

rules/import_middleman.bzl
jerrymarino: Yeah you got it @thiagohmcruz - the `Notes` blurb now links this:  https://github.com/bazel-ios/rules_ios/pull/767/files#diff-e213489d3a65f173ffa46728826b06f786a5d63666e3fe7101493397ee12f948R2

thiagohmcruz (approved):

PR #768, opened 2023-09-19
Make update_in_place opt-in (true by default)

thiagohmcruz:
Due to the `codesign` issues being redirected [here](https://github.com/bazel-ios/rules_ios/blob/master/tools/m1_utils/update_dynamic.sh#L24-L29) SwiftUI Previews will crash in Xcode (using `rules_xcodeproj`) with
```sh
Exception Type:  EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
```

This happens only in `vendored_xcframeworks` that already provide the slices I need for Apple Silicon so planning to conditionally disable this `update_in_place` logic only in "SwiftUI Previews" builds.

More generally how are we supposed to handle this `codesign` errors in `Debug` builds where a `provisioning_profile` is not set? My suggestion is to allow this conditional behavior while we discuss this, happy to file an issue if needed.

### DIFF
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -16,6 +16,9 @@ _FindImportsAspectInfo = provider(fields = {
 })
 
 def _update_framework(ctx, framework):
+    if not ctx.attr.should_update_in_place:
+        return framework
+
     # Updates the `framework` for Apple Silicon
     out_file = ctx.actions.declare_file(ctx.attr.name + "/" + framework.basename + ".framework" + "/" + framework.basename)
     out_dir = ctx.actions.declare_file(ctx.attr.name + "/" + framework.basename + ".framework")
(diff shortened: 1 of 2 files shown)


jerrymarino (commented):
I wonder if theres `arm64_sim` slice why it's be running it through this in the first place. 

The other potential is it might fit be better to just give them an ad-hoc signature

thiagohmcruz:
> I wonder if theres `arm64_sim` slice why it's running it through this in the first place.

The internal dep contains these two slices in `vendored_xcframeworks`:
```
                    "identifier": "ios-arm64",
                    "platform": "ios",
                    "platform_variant": "",
```

```
                    "identifier": "ios-arm64_x86_64-simulator",
                    "platform": "ios",
                    "platform_variant": "simulator",
```
So IIUC it is supposed to hit this code path since [import_middleman](https://github.com/bazel-ios/rules_ios/blob/ac2ceab094c92cadeb1a115c92e1b1ba306901c8/rules/library.bzl#L780-L781) is created.

> The other potential is it might fit better to just give them an ad-hoc signature
It's not clear to me how we would sign this in a generic way if it's not a device build, what would be the inputs to `codesign`?

jerrymarino:
I was more of thinking since it was a `platform_variant` simulator it doesn't need to be called - do you see what I mean? Othwerwise could try to do a force'd ad-hoc signature sans identity.

thiagohmcruz:
Closing in favor of https://github.com/bazel-ios/rules_ios/pull/769

PR #772, opened 2023-09-29
Improves make_vfsoverlay performance

thiagohmcruz:
Comparing the top level `_apple_framework_packaging_impl` and `_framework_vfs_overlay_impl` nodes. Overall seems like `.append` and `.extend` are more expensive than `+=`? The delta goes up to the `root` node so I don't think this is just perceived improvement.

#### Before:
Screenshot 2023-09-29 at 4 55 41 PM
Screenshot 2023-09-29 at 4 55 48 PM
Screenshot 2023-09-29 at 5 15 43 PM

#### After:
Screenshot 2023-09-29 at 4 55 54 PM
Screenshot 2023-09-29 at 4 55 58 PM
Screenshot 2023-09-29 at 5 15 47 PM

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -49,6 +49,7 @@ def _find_top_swiftmodule_file(swiftmodules):
 
 # Builds out the VFS subtrees for a given set of paths. This is useful to
 # construct an in-memory rep of a tree on the file system
+# buildifier: disable=list-append
 def _build_subtrees(paths, vfs_prefix):
     # It uses a O(NVFSParts) sized helper dict to avoid O(NPathComponents^2)
     # worst case runtime
@@ -57,7 +58,18 @@ def _build_subtrees(paths, vfs_prefix):
     for path_info in paths:
         path = path_info.framework_path
 
-        parts = path.split("/")
+        # Avoid calling `.split` when possible for better performance
+        parts = []
+        parts_len = 0
+        if "/" in path:
+            parts = path.split("/")
+            parts_len = len(parts)
+        else:
+            parts = [path]
+            parts_len = 1
+
+        if parts_len == 0:
+            fail("[ERROR] Failed to build VFS subtrees, path with empty split: parts=%s, parts_len=%s" % (parts, parts_len))
 
         # current pointer to the current subdirs while walking the path
         curr_subdirs = subdirs
@@ -66,11 +78,10 @@ def _build_subtrees(paths, vfs_prefix):
         # Assume the last bit is a file then add it as a file
         idx = 0
 
-        parts_len = len(parts)
         for part in parts:
             if idx == parts_len - 1:
                 curr_subdirs.dict[part] = -1
-                curr_subdirs.json["contents"].append({"name": part, "type": "file", "external-contents": vfs_prefix + path_info.path})
+                curr_subdirs.json["contents"] += [{"name": part, "type": "file", "external-contents": vfs_prefix + path_info.path}]
                 break
 
             # Lookup a value for the current subdirs, otherwise append
@@ -80,19 +91,23 @@ def _build_subtrees(paths, vfs_prefix):
                 next_subdirs = struct(dict = {}, json = next_subdirs_json)
 
                 curr_subdirs.dict[part] = next_subdirs
-                curr_subdirs.json["contents"].append(next_subdirs_json)
+                curr_subdirs.json["contents"] += [next_subdirs_json]
 
             curr_subdirs = next_subdirs
             idx += 1
     return subdirs_json
 
 def _get_private_framework_header(path):
+    if "/PrivateHeaders/" not in path:
+        return None
     last_parts = path.split("/PrivateHeaders/")
     if len(last_parts) < 2:
         return None
     return last_parts[1]
 
 def _get_public_framework_header(path):
+    if "/Headers/" not in path:
+        return None
     last_parts = path.split("/Headers/")
     if len(last_parts) < 2:
         return None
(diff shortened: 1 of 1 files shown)


rules/framework/vfs_overlay.bzl
```diff
@@ -57,7 +58,18 @@ def _build_subtrees(paths, vfs_prefix):
...
+        parts_len = 0
+        if "/" in path:
+            parts = path.split("/")
+            parts_len = len(parts)
+        else:
+            parts = [path]
+            parts_len = 1
+
+        if parts_len == 0:
+            fail("[ERROR] Failed to build VFS subtrees, path with empty split: parts=%s, parts_len=%s" % (parts, parts_len))
```
luispadron, line 72:
```suggestion
            fail("[ERROR] Failed to build VFS subtrees, path with empty split: parts=%s" % parts)
```

nit: parts_len seems redundant since it's only ever the case this is == 0

luispadron (approved):
Thats a nice improvement, thanks!

I wonder if we could `disable=list-append` to some global style list so we don't end up needing it in a lot of places

rules/framework/vfs_overlay.bzl
ob: nit: a better error message would print the original `path` and say it was trying to split on `/`

thiagohmcruz:
> Thats a nice improvement, thanks!
> 
> I wonder if we could `disable=list-append` to some global style list so we don't end up needing it in a lot of places

For now I've profiled only this file but if I end up doing a scan in other places will move it to a global list 👍

rules/framework/vfs_overlay.bzl
thiagohmcruz: SGTM! Will update.

ob (commented):
Nice! 🚀

ob:
> > Thats a nice improvement, thanks!
> > I wonder if we could `disable=list-append` to some global style list so we don't end up needing it in a lot of places
> 
> @luispadron For now I've profiled only this file but if I end up doing a scan in other places will move it to a global list 👍

Is this a starlark bug? Seems a bit surprising that `+=` would that much faster than `.append`... maybe we should file a bug in the `bazel` repo?

rules/framework/vfs_overlay.bzl
thiagohmcruz: This is just me being paranoid in case someone uses `parts_len` in a way that was not intended, regardless makes sense to improve the err msg: https://github.com/bazel-ios/rules_ios/pull/772/commits/8a662017dc7a2a888dcfeebaf7a0d0c85c891b42

thiagohmcruz:
> > > Thats a nice improvement, thanks!
> > > I wonder if we could `disable=list-append` to some global style list so we don't end up needing it in a lot of places
> > 
> > 
> > @luispadron For now I've profiled only this file but if I end up doing a scan in other places will move it to a global list 👍
> 
> Is this a starlark bug? Seems a bit surprising that `+=` would that much faster than `.append`... maybe we should file a bug in the `bazel` repo?

@ob I could be misinterpreting the results. Tried to not change any other state when comparing the two and looking at the `root` node the improvement is there. Will keep an eye on perf in our repro after this but in order to report this to Bazel folks I think we would want a more methodical way to compare the two in isolation? Maybe a mock repo with many source files and deps like ours internally 🤔.

rules/framework/vfs_overlay.bzl
```diff
@@ -57,7 +58,18 @@ def _build_subtrees(paths, vfs_prefix):
     for path_info in paths:
         path = path_info.framework_path
 
-        parts = path.split("/")
+        # Avoid calling `.split` when possible for better performance
+        parts = []
+        parts_len = 0
+        if "/" in path:
+            parts = path.split("/")
```
jerrymarino, line 65: Making this faster is good but a larger problem is that we have to call this a lot of times right now - which is why this shows up in the perf. It looks like this PR makes a decent savings however - it still has a lot of room to improve.

Can we either:
1. rework `build_subtress` / related data structures to memoize more or all o fthis? 
or:
2. if 1. isn't possible - then cache the `split` call here to do it one time ever https://github.com/bazel-ios/rules_ios/pull/772/files#diff-450b1330485549712d2fd4e65b2b3fd0a7692fe8f4ec43a0b7b60bbb739313afR141

Pt1 seems ideal - and fixing this will help improve _bazel_ bootstrapping time after a checkout a bit - but after it's computed once it's less of an issue

jerrymarino (approved):
Nice find - left a suggestion to make this a bit better

jerrymarino:
> Is this a starlark bug? Seems a bit surprising that += would that much faster than .append... maybe we should file a bug in the bazel repo?

Huge ➕  - perhaps just apply the `split` fix and profile w/o the syntactical changes?

ob:
> Will keep an eye on perf in our repro after this but in order to report this to Bazel folks I think we would want a more methodical way to compare the two in isolation? Maybe a mock repo with many source files and deps, similar to ours internally 🤔.
> 
> (Plus, there are other changes here not involving .append/.extend too, one more reason to look for a more methodical approach)

I created https://github.com/ob/starlark-perf-test-1, which builds 1,000 paths with 10,000 path components using both methods and it looks like appending is 10% slower than using `+=`...

rules/framework/vfs_overlay.bzl
thiagohmcruz: I don't want to bloat this PR with more scope atm but I do think your suggestions make sense. I'm doing some small optimizations as I work on something else so not planning to spend much more time on this one for now.

That said, there's def more room for improvement here and will keep you suggestions in mind next time I tackle this.

thiagohmcruz:
> > Will keep an eye on perf in our repro after this but in order to report this to Bazel folks I think we would want a more methodical way to compare the two in isolation? Maybe a mock repo with many source files and deps, similar to ours internally 🤔.
> > (Plus, there are other changes here not involving .append/.extend too, one more reason to look for a more methodical approach)
> 
> I created https://github.com/ob/starlark-perf-test-1, which builds 1,000 paths with 10,000 path components using both methods and it looks like appending is 10% slower than using `+=`...

Started a thread here: https://bazelbuild.slack.com/archives/CDCEKVCHY/p1696356866950979

PR #773, opened 2023-09-29
Support rules_apple 3.x.x

jerrymarino:
1. update rules_ios to use rules_apple 3.x.x
2. Adding support for rules_apple version 3. I have added a simple starlark split transition to pass existing tests that failed in the original update.
3. Add explict rules_apple API wrapper. Eventually all uses and hacks to make rules_apple work can flow through here.
4. Add CI job for rules_apple 2 - ( default is now 3 )


.github/workflows/tests.yml
```diff
@@ -107,6 +107,31 @@ jobs:
           name: bazel-testlogs
           path: bazel-testlogs
 
+  rules_apple_2_ios_integration_tests:
+    name: Build and Test ( Virtual Frameworks + rules_apple 2.x )
+    runs-on: macos-12
+    env:
+      USE_BAZEL_VERSION: 6.1.2
```
thiagohmcruz, line 114: Q: is this supposed to match `.bazelversion`?

.github/workflows/tests.yml
```diff
@@ -200,6 +225,10 @@ jobs:
   multi_arch_support:
     name: Build iOS App for Multiple Architecture
     runs-on: macos-12
+    env:
+      # i386 was removed on rules_apple 3.x.x - this test case needs reworking
+      # to exemplify fat binaries on the latest version
```
thiagohmcruz, line 230: Nit: Why is this comment in the Bazel version line? Potentially confusing as it looks like an issue with the version of Bazel itself? Maybe move it up?

rules/framework.bzl
```diff
@@ -699,70 +726,130 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
...
     )
-    link_result = linking_support.register_binary_linking_action(
-        ctx,
-        avoid_deps = avoid_deps,
-        entitlements = None,
-        extra_linkopts = extra_linkopts,
-        platform_prerequisites = platform_prerequisites,
-        stamp = ctx.attr.stamp,
-    )
+    if apple_api_version == "3.0":
```
thiagohmcruz, line 729: For future readers it might get confusing to understand what `apple_api_version` means, should we explicitly give it the rule set's name e.g. `rules_apple_api_version`?

rules/framework.bzl
```diff
@@ -699,70 +726,130 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
...
-    processor_partials.append(
-        partials.apple_bundle_info_partial(
+    if apple_api_version == "3.0":
+        archive_for_embedding = outputs.archive_for_embedding(
             actions = actions,
-            bundle_extension = bundle_extension,
-            bundle_id = bundle_id,
             bundle_name = bundle_name,
-            executable_name = bundle_name,
+            bundle_extension = bundle_extension,
```
thiagohmcruz, line 756: Is `executable_name` not needed here or was it removed?

thiagohmcruz (commented):
I paused mid review as I wanted to clarify one thing first. What is the problem we're solving by adding all this additional handling for old versions? What is preventing us from simply moving on to Bazel 6 + rules_apple 3.x at HEAD?

I assume the intent is to support older versions for a longer period of time in case someone needs it. Why not do what other rule sets are doing then and (1) define the latest versions as "minimum" or "current" and (2) agree on a process to update older releases of `rules_ios` with critical bug fixes if necessary?

Not against the backwards-compatibility changes but wanted to understand if we absolutely need it to be implemented this way or not? Have you considered alternatives like mentioned above?

Maintaining two Bazel versions and two `rules_apple` versions at HEAD will increase maintenance cost for all of us and from a readability perspective there's def some overhead being added here for maintainers to have to understand what the split is doing exactly and why they can't just use known provider names in favor of the `new_{SomeProvider}` pattern.

rules/test.bzl
jszumski, line 1: `platform_type` and `minimum_os_version` need to be set in this file to avoid iOS version mismatches in tests:

> INFO: From Linking tests/ios/app/AppTestOnly_bin:
> 
> ld: warning: object file (bazel-out/ios-x86_64-applebin_ios-ios_x86_64-dbg-ST-78e2e727b17e/bin/tests/ios/frameworks/testonly/libMixedSourceFramework_objc.a(AngleBracketNamespacedLogger.o)) was built for newer iOS Simulator version (16.2) than being linked (10.0)
> 
> ld: warning: object file (bazel-out/ios-x86_64-applebin_ios-ios_x86_64-dbg-ST-78e2e727b17e/bin/tests/ios/frameworks/testonly/libMixedSourceFramework_swift.a(SwiftLibrary+extension.swift.o)) was built for newer iOS Simulator version (16.2) than being linked (10.0)

This worked for me:

```diff
-    framework_middleman(name = fw_name, framework_deps = kwargs.get("deps", []) + library.lib_names, testonly = testonly, tags = ["manual"])
+    framework_middleman(
+        name = fw_name,
+        framework_deps = kwargs.get("deps", []) + library.lib_names,
+        testonly = testonly,
+        tags = ["manual"],
+        platform_type = "ios",
+        minimum_os_version = ios_test_kwargs.get("minimum_os_version"),
+    )
     frameworks = [fw_name] + ios_test_kwargs.pop("frameworks", [])
 
     dep_name = name + ".dep_middleman"
-    dep_middleman(name = dep_name, deps = kwargs.get("deps", []) + library.lib_names, testonly = testonly, tags = ["manual"], test_deps = host_args)
+    dep_middleman(
+        name = dep_name,
+        deps = kwargs.get("deps", []) + library.lib_names,
+        testonly = testonly,
+        tags = ["manual"],
+        test_deps = host_args,
+        platform_type = "ios",
+        minimum_os_version = ios_test_kwargs.get("minimum_os_version"),
+    )
```

rules/framework.bzl
jerrymarino: Yeah it was removed

rules/framework.bzl
jerrymarino: SGTM

.github/workflows/tests.yml
jerrymarino: Ok fair enough will move it up then

.github/workflows/tests.yml
jerrymarino: It is how `bazelisk` uses multiple versions different from the `.bazelversion`

jerrymarino: So no - `bazelisk` allows you to override the version which doesn't match the text file

rules/test.bzl
jerrymarino: Good catch will update

jszumski: I pushed it, feel free to edit as necessary

rules/framework.bzl
jerrymarino: Updated

jerrymarino:
@thiagohmcruz I've pushed the latest updates based on what we discussed offline. For now, we've determined to continue to support the older versions on HEAD and will have to re-evaluate at a later day

.github/workflows/tests.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           name: bazel-testlogs
           path: bazel-testlogs
 
-  lts_ios_integration_tests:
+  bazel_lts_ios_integration_tests:
```
luispadron, line 85: IMO the name here is confusing, if it's going to change in this PR could we add bazel5 to the name instead of LTS. 

LTS applies to multiple bazel versions including 6

.github/workflows/tests.yml
```diff
@@ -201,12 +226,16 @@ jobs:
...
+    # to exemplify fat binaries on the latest version
     name: Build iOS App for Multiple Architecture
     runs-on: macos-12
+    env:
+      USE_BAZEL_VERSION: 6.1.2
     steps:
       - uses: actions/checkout@v3
       - name: Preflight Env
-        run: .github/workflows/preflight_env.sh
+        run: .github/workflows/preflight_env.sh --no-bzlmod
```
luispadron, line 238: Bzlmod support was merged with 6.1.2, why should this be disabled here?

luispadron (commented):
Some comments, I still don't understand why we gate the rules_apple version based on the Bazel version.

luispadron (commented):
Can we keep using the release versions, same for rules_swift.

luispadron (commented):
Can we keep using the release versions, same for rules_swift.

rules/repositories.bzl
```diff
@@ -116,30 +133,59 @@ def _rules_ios_bzlmod_dependencies():
             repo = "rules_apple",
             sha256 = "1618fc82e556ebc97ea360b8cacd3365ca3b0e0a85ccb32422468204843e752d",
         )
+        _maybe(
+            rules_apple_api,
+            name = "rules_apple_api",
+            version = "1_0",
+        )
+    elif bazel_version.major == "6" and bazel_version.minor == "1":
```
luispadron, line 141: Why is this needed? rules_apple 2 and 3 both support Bazel 6.x.

luispadron (commented):
In general, I'm confused why rules_apple version 2 is used in any of these checks. Don't we just need to check `is this rules_apple 3?` and use the appropriate provider methods?

rules/repositories.bzl
```diff
@@ -116,30 +133,59 @@ def _rules_ios_bzlmod_dependencies():
...
+        )
+        _maybe(
+            rules_apple_api,
+            name = "rules_apple_api",
+            version = "2_0",
+        )
     else:
+        # Bazel 6+ codepath - you can use the 2_x version of rules_apple here if
+        # you'd like to set it. If you want to use a different rules_apple than
+        # we have here - define rules_apple_api - to the value you'd like
```
luispadron, line 166: Any Bazel user can already override this to be whatever version they want. They can simply add a rules_apple dependency before rules_ios to override whatever rules_ios provides

.github/workflows/tests.yml
jerrymarino: [Please find in](https://github.com/bazel-ios/rules_ios/pull/773/files/9b0fe3fd7d2df44651e094308452e928aee83898#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR229) in the proceeding comment that it uses 6.1.2 and rules_apple 2 - where bzlmod is hardcoded to use 3.0.0.

rules/repositories.bzl
jerrymarino: We currently test rules_apple 2 with `6.1.2` on CI

jerrymarino:
> In general, I'm confused why rules_apple version 2 is used in any of these checks.

We agreed to support 2.x.x on HEAD offline as a team as it's what we use in some key use case

luispadron:
> > In general, I'm confused why rules_apple version 2 is used in any of these checks.
> 
> We agreed to support 2.x.x on HEAD offline as a team as it's what we use in some key use case

But now we seem to have no ios integration tests running Bazel 6 with rules_apple 3 & bzlmod: https://github.com/bazel-ios/rules_ios/blob/9b0fe3fd7d2df44651e094308452e928aee83898/.github/workflows/tests.yml#L85-L112 or am I missing it in there. Just want to make sure we still test everything with the "latest" i.e. Bazel 6.3.2 and rules_apple 3

I wasn't part of that specific discussion but my understanding ya'll agreed on not dropping Bazel 5 tests from the repository. This seems to add another CI requirement that rules_apple 2 is also supported with this specific Bazel 6.1.x configuration.

.github/workflows/tests.yml
luispadron: Gotcha so it doesnt work yet with rules_apple 3. I guess I got confused because we're essentially using Bazel `6.1.2` to mean rules_apple 2.

Not for this PR, but in the future I wonder if specifying these Bazel version + rules_apple versions wouldn't be better in a specific `WORKSPACE` for that specific test. Some repos do this with their integration tests and it allows them to test the different bazel version / rule versions they support without setting one version/feature for all their tests

.github/workflows/tests.yml
luispadron: i think adding `bazel5` like i said would it make clear this specifically is a bazel5 job.

thiagohmcruz: +1 on being more explicit on the naming here

.github/workflows/tests.yml
thiagohmcruz: Nit: I'd still explicitly say in that comment why that implies `--no-bzlmod` is needed. The same way it's confusing during review it will be once we're changing this in the future.

Saw what @luispadron is talking about in other repos too and also like the idea. Wondering if we should create "Feature Request" templates for issues in `rules_ios` to track things like this 🤔

jerrymarino:
> This seems to add another CI requirement that rules_apple 2 is also supported with this specific Bazel 6.1.x configuration.

Yes, it does have to to support rules_apple 2 is because we will use this version and have to support it for us. We can move the _bazel version_ forward from 6.1.2 but will require more followups to the testing infra and it doesn't impact anything beyond our testing infra.

.github/workflows/tests.yml
jerrymarino: Ok sounds good to me - this is a good improvement in the github UI

.github/workflows/tests.yml
jerrymarino: Yeah it basically exists because people aren't using bzlmod on 5 and 6 so it's good to have some coverage of both ways to test this. Though, we could bump to a later version but needs more consideration.

Also, the way it's configured now get's us a larger matrix of bzlmod vs non bzlmod across a few versions. Though bzlmod support works on all the versions in the repo 5.4.1, 6.x.x we'd still like like the coverage to we don't break it - but probably not testing 100% of permutations

jerrymarino:
@jszumski thanks for raising it!

It's possible that we might need to tackle some followups or other improvements to it in general - I'm happy to review a PR to tackle it. We did this in a way that you can still use rules_apple 2.x.x and may have smaller incremental improvements to 3.x.x. this is done

jerrymarino:
Also perhaps [starlark validation of versions](https://github.com/bazel-ios/rules_ios/pull/582) may be useful here outside of linker errors in general

jszumski:
Somewhat false alarm, that warning has been present in `master` for a long time (as far back as our GitHub Action log retention, at least!).  I'll add a todo to bisect the original cause

PR #776, opened 2023-10-16
Update stardoc for bzlmod

luispadron:
This PR updates stardoc for use with `bzlmod`. This is required as the versions of stardoc on the registry require these stricter `deps` changes.

This updates our documentation (i.e. `stardoc` and `bzl_library` targets) to work in a similar way to rules_apple: https://github.com/bazelbuild/rules_apple/blob/master/apple/BUILD#L33

Each file we want documentation generated for defines it's own `bzl_library` with it's own set of `deps`. These target names are then used within the `stardoc` rule to generate documentation for each of them.

The main change is going from one giant `bzl_library` to multiple `bzl_library`. 

Depends on the following stack:

- #777 
- #778


luispadron:
I plan to PR two things before this PR is merged but that it depends on:

- Updating Bazel verison, we need `>= 6.3.2` to fix issues with bzlmod and stardoc
- Update to use `bazel_skylib`'s `write_file` rule vs our custom copy of it, reduces code duplication and amount of `bzl_library` needed to create

rules/library.bzl
```diff
@@ -4,7 +4,7 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:sets.bzl", "sets")
 load("@bazel_skylib//lib:selects.bzl", "selects")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
-load("@build_bazel_rules_apple//apple:apple.bzl", "apple_dynamic_framework_import", "apple_static_framework_import")
+load("@build_bazel_rules_apple//apple/internal:apple_framework_import.bzl", "apple_dynamic_framework_import", "apple_static_framework_import")
```
luispadron, line 7: Changing from `apple.bzl` umbrella import to specific internal import fixes works around an issue i reported in stardoc + bzlmod: https://github.com/bazelbuild/stardoc/issues/192

rules/repositories.bzl
```diff
@@ -168,6 +168,16 @@ def _rules_ios_bzlmod_dev_dependencies():
...
 
+    _maybe(
+        http_archive,
+        name = "io_bazel_stardoc",
+        sha256 = "62bd2e60216b7a6fec3ac79341aa201e0956477e7c8f6ccc286f279ad1d96432",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/stardoc/releases/download/0.6.2/stardoc-0.6.2.tar.gz",
+            "https://github.com/bazelbuild/stardoc/releases/download/0.6.2/stardoc-0.6.2.tar.gz",
+        ],
+    )
```
luispadron, line 179: Moved this from `WORKSPACE` to the dev_dependencies section so that it matches the other rules_ios deps.

WORKSPACE
```diff
@@ -47,22 +47,28 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
...
+
+rules_jvm_external_setup()
+
+load("@io_bazel_stardoc//:deps.bzl", "stardoc_external_deps")
+
+stardoc_external_deps()
+
+load("@stardoc_maven//:defs.bzl", stardoc_pinned_maven_install = "pinned_maven_install")
+
+stardoc_pinned_maven_install()
```
luispadron, line 70: This is required by new stardoc versions

WORKSPACE.bzlmod
```diff
@@ -4,24 +4,3 @@
...
-git_repository(
-    name = "io_bazel_stardoc",
-    commit = "6f274e903009158504a9d9130d7f7d5f3e9421ed",
-    remote = "https://github.com/bazelbuild/stardoc.git",
-    shallow_since = "1667581897 -0400",
-)
-
-load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
-
-stardoc_repositories()
```
luispadron, line 27: Now fully bzlmod 🎉

docs/vfs_overlay.doc_doc.md
justinseanmartin, line 1: The file name here and `xcconfig.doc_doc.md` seem odd

rules/xcconfig.doc.bzl
```diff
@@ -0,0 +1,10 @@
+load(
+    "//rules/library:xcconfig.bzl",
```
justinseanmartin, line 2: Should this live under `rules/library/xcconfig.doc.bzl` to be next to its associated rule? I presume we don't want all of `//rules/library:xcconfig.bzl` to be considered public API, so we're cherry-picking a subset of it here? Should we mark the rest as internal symbols with an underscore prefix or something, if that prevents it from being documented?

Same feedback for `vfs_overlay`.

justinseanmartin (approved):
Take comments with a grain of salt and nothing that I'd consider a blocker to landing this.

docs/vfs_overlay.doc_doc.md
luispadron: Just fixed this, should be generated correctly now

rules/xcconfig.doc.bzl
luispadron: The use of `.doc.bzl` matches what `rules_apple` does, it's exporting the public api of `xcconfig.bzl` in the `//rules` package specifically. This makes it so we can create the `stardoc` target for these in one list vs. conditionalizing the path to the `bzl_library`

See this in `docs/BUILD.bazel`:

```
stardoc(
   ...
   input = "//rules:%s.bzl" % file,
)
```

docs/BUILD.bazel
```diff
@@ -1,50 +1,48 @@
...
 
 package(default_visibility = ["//visibility:private"])
 
-_RULES = [
+# Targets to document.
+# Must have a coressponding `bzl_library` target in `//rules:BUILD.bazel`.
+_DOC_SRCS = [
     "app_clip",
     "app",
-    "apple_patched",
```
jerrymarino, line 9: This change looks orthogonal to the change

docs/BUILD.bazel
```diff
@@ -1,50 +1,48 @@
...
     "extension",
+    "features",
+    "force_load_direct_deps",
```
jerrymarino, line 12: A number of these of these are internal ones which I don't think we really want to provide as an API contract or have starlark docs about what they do but if other people really want to have this I'm don't think it will change the lack of an API beyond `apple_framework` to the project

docs/BUILD.bazel
luispadron: Missed this one

I'll work on supporting this with new stardoc version and update the PR

docs/BUILD.bazel
luispadron: Features seems useful and imo more of the public API than some of the other things we've said are. 

Do you have a specific list you'd want me to remove?

docs/BUILD.bazel
luispadron: Fixed this up, its now included again

docs/BUILD.bazel
```diff
@@ -1,50 +1,49 @@
 load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
-load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
 package(default_visibility = ["//visibility:private"])
 
-_RULES = [
+# Targets to document.
+# Must have a coressponding `bzl_library` target in `//rules:BUILD.bazel`.
```
thiagohmcruz, line 6:
```suggestion
# Must have a corresponding `bzl_library` target in `//rules:BUILD.bazel`.
```

thiagohmcruz (approved):
Thx so much for tackling this!

PR #778, opened 2023-10-16
Move to using `write_file` from Bazel Skylib

luispadron:
Makes #776 a bit easier as one less rule to document and saves some code.


.bazelversion
```diff
@@ -1 +1 @@
-6.1.2
+6.3.2
```
luispadron, line 1: Oops stacked PR issue, I'll rebase this once Bazel update PR is merged

mattrobmattrob (approved):
👏 This should probably be a major version bump since it removes a macro/rule.

luispadron:
> 👏 This should probably be a major version bump since it removes a macro/rule.

Sounds good, I'll release a 3.0 bundled with the stardoc bzlmod upgrade (which also changes some api around bzl_library) and it'll be the first version released to the registry

jszumski (approved):

justinseanmartin (approved):
FYI - On a quick GH search, I couldn't find any public users of this API, but there still could presumably be some private ones. I don't feel super strongly about the major version bump personally, but not opposed to it either to follow semver to the letter. I don't think this was ever intended as part of the public API surface and is easily replaceable. It might be nice if we could be more strict about the public API surface so we don't end up needing to make a breaking API change for something like this.

thiagohmcruz (approved):

PR #780, opened 2023-10-17
Add Bazel central registry files

luispadron:
Depends on #776.

This adds the required template files for https://github.com/bazel-contrib/publish-to-bcr an action which will publish to the Bazel central registry when new releases are tagged.

We'll need someone with repository admin access to approve the action configuration I requested.

### DIFF
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -1,0 +1,14 @@
+{
+  "homepage": "https://github.com/bazel-ios/rules_ios",
+  "maintainers": [
+    {
+      "email": "rules_ios@squareup.com",
+      "name": "rules_ios Maintainers"
+    }
+  ],
+  "repository": [
+    "github:bazel-ios/rules_ios"
+  ],
+  "versions": [],
+  "yanked_versions": {}
+}
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,0 +1,12 @@
+bcr_test_module:
+  module_path: "e2e/bzlmod"
+  matrix:
+    platform: ["macos"]
+  tasks:
+    build_flags:
+      - "--config=ios"
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      build_targets:
+        - "//..."
--- a/.bcr/source.template.json
+++ b/.bcr/source.template.json
@@ -1,0 +1,5 @@
+{
+  "integrity": "",
+  "strip_prefix": "",
+  "url": "https://github.com/{OWNER}/{REPO}/releases/download/{TAG}/{REPO}.{TAG}.tar.gz"
+}
--- a/.github/workflows/generate_release_notes.sh
+++ b/.github/workflows/generate_release_notes.sh
@@ -14,6 +14,12 @@ readonly release_archive="rules_ios.$new_version.tar.gz"
 sha=$(shasum -a 256 "$release_archive" | cut -d " " -f1)
 
 cat <<EOF
+### Bzlmod Snippet
+
+\`\`\`bzl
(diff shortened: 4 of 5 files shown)


.bcr/metadata.template.json
```diff
@@ -0,0 +1,15 @@
+{
+  "homepage": "https://github.com/bazel-ios/rules_ios",
+  "maintainers": [
+    {
+      "email": "heyluispadron@gmail.com",
+      "github": "luispadron",
+      "name": "Luis Padron"
+    }
+  ],
```
luispadron, line 9: @jszumski @thiagohmcruz @mattrobmattrob @jerrymarino @justinseanmartin please provide the information needed here, you're the first that come to mind as active maintainers here but if we have more people we'd like listed please provide those here too.

I dont have access to the repo settings but maybe listing all users with approval / merge access here makes sense.

.bcr/config.yml
```diff
@@ -0,0 +1,3 @@
+fixedReleaser:
+  login: luispadron
+  email: heyluispadron@gmail.com
```
luispadron, line 3: For now we'll need to mark a single person as releaser. See an example PR for what this does:

- PR Example: https://github.com/bazelbuild/bazel-central-registry/pull/1010
- Information on this config: https://github.com/bazel-contrib/publish-to-bcr#a-note-on-release-automation

In the future we could create a bot account to handle this.

.bcr/presubmit.yml
```diff
@@ -0,0 +1,12 @@
...
+  matrix:
+    platform: ["macos"]
+  tasks:
+    build_flags:
+      - "--config=ios"
+    run_tests:
+      name: "Run test module"
+      platform: ${{ platform }}
+      build_targets:
+        - "//..."
```
luispadron, line 12: Our release archive doesnt contain the `tests` directory so we just care about building here. This is similar to other repos that maintain minimal release archives. 

Relying on rules_ios CI is good enough IMO and ensuring the release stuff builds is probably good enough as a starting point for this check.

.bcr/config.yml
jerrymarino: Perhaps for rules_ios the release author is a shared google group email?

.bcr/metadata.template.json
jerrymarino: Is this 100% necessary to add this field? We may have to be mindful of what emails to add to the repo - will try to start a thread offline

.bcr/config.yml
luispadron: Makes sense, though it needs to be a GitHub account with a fork of the BCR since that's how it makes the PR to release. If you want to make that account that would be great and I can get this PR updated

.bcr/metadata.template.json
luispadron: Seems to be required from the template/other repos I'm seeing. We could make this that bot account though if we make it

jerrymarino: Yeah ideally we can setup a google group / account for it

mattrobmattrob: Would be interesting to try excluding the `maintainers` but just guessing that's a possibility based on https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#validations

luispadron:
Are we okay with merging this with me as the releaser and re-visiting this to add more info / use a shared account when that's possible.

This essentially just marks me as the maintainer for the BCR part of rules_ios which I'm okay with. I'd just like to unblock this if possible

luispadron:
Updated PR to:

- Remove config.yml, this means no one will get tagged since it'll tag the bot like [here](https://github.com/bazelbuild/bazel-central-registry/pull/990).
- Set the shared Google workspace email as the sole maintainer.

We can attempt with this setup and followup as needed.

justinseanmartin (approved):

Issue #783, opened 2023-10-19
Issues w/stardoc + bzlmod: currently runnning stardoc w/o bzlmod 


jerrymarino:
I ran into some unexplainable bugs that didn't exist after rebasing the rules_apple 3.x.x bump on the latest changes which updated stardoc a couple days ago 

Please find some of the other issues which are also reported to stardoc: bazelbuild/stardoc#192

### Opaque errors when some "bzl_libraries" are missing transitive files
If you revert some of the fixes I made to the "bzl_library" doc targets we have you can find them. It isn't clear why some of these errors are popping and stardoc gives no actionable output to me. Perhaps it's related to the issue below "Mirroring the entire include structure into bzl_library"

This is the largest issue I ran into: with the latest change we broke down several "bzl_library" targets for each file so it is quite easy to run into this situation when changing imports.

### Non-actionable errors missing how to import misc files from transitive deps
> bazel-out/darwin-opt-exec-2B5CBBC6/bin/docs/app_clip_doc_stardoc '--input=@@//rules:app_clip.bzl' '--?> workspace_name=_main' '--output=bazel-out/darwin-dbg/bin/docs/app_clip_doc.raw')
># Configuration: 522a85f38684f2efaef2b90a82a9ce08d5355dfc2131858996e59ade94cbf161
> # Execution platform: @local_config_platform//:host
> com.google.devtools.build.skydoc.SkydocMain$StarlarkEvaluationException: File 
> /private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/execroot/_main/bazel-out/darwin-opt-exec-
> 2B5CBBC6/bin/docs/app_clip_doc_stardoc.runfiles/rules_apple3.0.0/apple/internal/transition_support.bzl imported 
> '@build_bazel_apple_support//configs:platforms.bzl', yet 
>/private/var/tmp/_bazel_runner/c2342bca25695567c8c3f3be166b56b3/execroot/_main/bazel-out/darwin-opt-exec-
>B5CBBC6/bin/docs/app_clip_doc_stardoc.runfiles/apple_support~1.10.1/configs/platforms.bzl 
>was not found.
>	at com.google.devtools.build.skydoc.SkydocMain.recursiveEval(SkydocMain.java:4[49](https://github.com/bazel-?>ios/rules_ios/actions/runs/6568570441/job/17843180059#step:4:50))
[...]

PR #785, opened 2023-10-19
[Stardoc] Cleanup new stardoc include issues, removing internal docs

jerrymarino:
1. Export all files to `bzl_srcs`
2. Remove mirror of includes into `bzl_library`
3. Remove documentation of internal sources

Related: #783 #776


rules/BUILD.bazel
```diff
@@ -51,38 +48,20 @@ bzl_library(
...
-    ],
-)
-
 bzl_library(
     name = "xcconfig.doc",
     srcs = ["xcconfig.doc.bzl"],
     tags = ["manual"],
     visibility = ["//visibility:public"],
-    deps = ["//rules/library:xcconfig"],
+    deps = ["bzl_srcs"],
```
jszumski, line 56:
```suggestion
    deps = [":bzl_srcs"],
```

rules/BUILD.bazel
```diff
@@ -126,101 +87,72 @@ bzl_library(
...
     name = "extension",
     srcs = ["extension.bzl"],
     tags = ["manual"],
     visibility = ["//visibility:public"],
+    deps = [":bzl_srcs"],
+)
+
+# A library for all of the bzl_srcs, to satisfy `stardoc`'s deps attribute.
+bzl_library(
+    name = "bzl_srcs",
```
jszumski, line 135: Would `name = "bzl_deps"` better reflect how this is used?

I'm not a stardoc expert, but is it an anti-pattern to just include the world here instead of being more targeted in the specific `bzl_library`s above?  Maybe we keep this to all external deps and leave in-repo ones on the individual libraries?

luispadron:
I think if the intention here is to not have to define all the `loads` correctly because its a maintenance cost we should look at doing something similar to this PR: https://github.com/bazelbuild/rules_swift/pull/1132 and just generate these targets.

jerrymarino:
I'm not tied to generating them with gazelle or starlark, the latter is likely a more simple way to tackle this but I'm happy to review a PR if you have a gazelle implementation for it.

PR #787, opened 2023-10-24
Update bzlmod files for first publish

luispadron:
Feedback from: https://github.com/bazelbuild/bazel-central-registry/pull/1052

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,12 +1,6 @@
-bcr_test_module:
-  module_path: "e2e/bzlmod"
-  matrix:
-    platform: ["macos"]
-  tasks:
-    build_flags:
-      - "--config=ios"
-    run_tests:
-      name: "Run test module"
-      platform: ${{ platform }}
-      build_targets:
-        - "//..."
+tasks:
+  verify_build_targets:
+    name: Verify Build targets on macOS
+    platform: macos
+    build_targets:
+      - "@rules_ios//rules/..."
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -23,13 +23,15 @@ jobs:
 
           # Archive the repository
           # Considerations:
-          # - Dont package hidden files (including the .git directory)
+          # - Dont package development files (vscode config, etc)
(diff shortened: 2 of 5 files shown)


mattrobmattrob (approved):

fmeum (commented):
This looks okay assuming that building everything doesn't require dev dependencies: They will not be available in the BCR test setup which works by creating a synthetic empty module depending on your root module - your module won't be the root module.

luispadron:
> This looks okay assuming that building everything doesn't require dev dependencies: They will not be available in the BCR test setup which works by creating a synthetic empty module depending on your root module - your module won't be the root module.

Makes sense, so i'll make this more strict to just build `//rules` since some other things require dev dependencies.

thiagohmcruz (approved):

luispadron:
Noticing some more changes I need to make to `MODULE.bazel`, testing with path override. Will update this PR.

luispadron:
Had to make two changes:

- Mark `bazel_skylib` as a non-dev dependency, this is loaded by our rules so it cant be development only.
- Move the buildifier target from `//:BUILD.bazel` to `//tools:BUILD.bazel`, this allows us to move the load out of the rules path and allows us to keep buildifier as a dev depenedency
  - I added an `alias` so you can still do `bazel run //:buildifier`

With these changes, I get a green build depending on rules_ios from another bzlmod module:

```sh
bazel build @rules_ios//rules/...        
                                                                                                     
INFO: Analyzed 67 targets (134 packages loaded, 17858 targets configured).
INFO: Found 67 targets...
INFO: From Linking external/rules_ios~override/rules/hmap/hmaptest:
ld: warning: ignoring duplicate libraries: '-lc++'
INFO: From Linking external/rules_ios~override/rules/hmap/hmaptool:
ld: warning: ignoring duplicate libraries: '-lc++'
INFO: From Linking external/rules_ios~override/rules/hmap/hmapdump:
ld: warning: ignoring duplicate libraries: '-lc++'
INFO: Elapsed time: 28.322s, Critical Path: 7.12s
INFO: 442 processes: 257 internal, 142 darwin-sandbox, 43 local.
INFO: Build completed successfully, 442 total actions
```

PR #788, opened 2023-10-24
Use remote cache when building in xcodeproj on CI

mattrobmattrob:
Attempt to speed up the slow `.xcodeproj Tests on Xcode 14.2.0` job on CI.

### DIFF
--- a/.github/workflows/preflight_env.sh
+++ b/.github/workflows/preflight_env.sh
@@ -2,7 +2,11 @@
 
 # If flag --no-bzlmod is passed, writes a user.bazelrc file to disable Bzlmod.
 if [[ "$*" == *--no-bzlmod* ]]; then
-  echo "build --noexperimental_enable_bzlmod" > user.bazelrc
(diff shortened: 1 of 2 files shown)


jszumski:
From the timing it looks like this ended up saving 7m?

mattrobmattrob:
> From the timing it looks like this ended up saving 7m?

These GitHub VMs are impossible to benchmark against I'm finding. Ha. Will look a bit deeper tomorrow.

mattrobmattrob:
In general, I think this is a positive change. But the slowness of this particular build is likely coming from lots of places.

For example, the serial generation of the Xcode projects is slow (even on my M1 Pro MacBook Pro), likely even worse than this benchmark on the IO starved Github Action VMs:
```
$ time bazelisk query 'kind("xcodeproj rule", tests/ios/xcodeproj/...)' | xargs -n 1 bazelisk run
...
xargs -n 1 bazelisk run  18.36s user 68.33s system 76% cpu 1:52.81 total
```

jszumski (approved):

PR #789, opened 2023-10-25
Pre-boot simulator for xcodebuild usage

mattrobmattrob:
This aims to fix the test time when running the iOS/macOS `xcodebuild ... test` commands:
```
2023-10-24 23:02:41.288 xcodebuild[2658:16239] [MT] IDETestOperationsObserverDebug: 0.000 sec, +0.000 sec -- start
2023-10-24 23:02:41.288 xcodebuild[2658:16239] [MT] IDETestOperationsObserverDebug: 153.510 sec, +153.510 sec -- end
```
It seems to be paid on each invocation but can be avoided when the sim is pre-booted. Looking like roughly a 50% reduction compared to before (20-25m => 10-15m).


thiagohmcruz (approved):
Awesome! This is a significant reduction 😯

PR #790, opened 2023-10-25
Remove dependency on rules_python

luispadron:
For bzlmod you must declare dependency on things you `load`, in WORKSPACE mode `rules_python` is brought in transitively but in bzlmod we need to define this in `MODULE.bazel`.

Given that `rules_python` has beta support for bzlmod and that the Python rules we use are already part of the native rules, this PR removes all `load` of `rules_python` in favor of using the rules defined in the native module.

It also updates them to define the `PY3` source version.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -28,10 +28,6 @@ bazel_dep(
     version = "0.0.6",
 )
 bazel_dep(
-    name = "rules_python",
-    version = "0.24.0",
-)
-bazel_dep(
     name = "rules_swift",
     version = "1.13.0",
     repo_name = "build_bazel_rules_swift",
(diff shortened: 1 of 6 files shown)


justinseanmartin (approved):

thiagohmcruz (approved):

mattrobmattrob (approved):
Nice

justinseanmartin:
Was this problematic for some reason? IIUC - I thought we wanted to move away from native rules to starlark in general so that things can be decoupled from Bazel bumps. Am I getting that right?

luispadron:
> Was this problematic for some reason?

I was seeing a rules_python toolchain resolution error, but it didn't turn out to be related to this specifically. It did make me question why we depend on rules_python though here when the other rulesets seem to just the native rules.

Since we only use these rules for essentially a few python scripts, it feels heavy handed

> IIUC - I thought we wanted to move away from native rules to starlark in general so that things can be decoupled from Bazel bumps. Am I getting that right?

Yeah I think so, although the stuff in rules_python is _currently_ just an alias for the native rules.

> Currently, the core rules build into the Bazel binary, and the symbols in this repository are simple aliases. However, we are migrating the rules to Starlark and removing them from the Bazel binary.
> https://github.com/bazelbuild/rules_python/#overview

Given rules_python is still beta for bzlmod and we don't use it for anything like the PyPi support it seemed best to not bring it in yet until the other rules we depend on start using / starlark transition happens

PR #793, opened 2023-10-26
Allow Bazel 6 users to explicitly opt-in to rules_apple 2.x

thiagohmcruz:
Regressed in https://github.com/bazel-ios/rules_ios/pull/773

### DIFF
--- a/rules/repositories.bzl
+++ b/rules/repositories.bzl
@@ -152,6 +152,10 @@ def _rules_ios_bzlmod_dependencies(load_rules_apple_2_dependencies = False):
             sha256 = "c6966ec828da198c5d9adbaa94c05e3a1c7f21bd012a0b29ba8ddbccb2c93b0d",
         )
     elif load_rules_apple_2_dependencies:
+        # This `load_rules_apple_2_dependencies` flag is simply a convenience for consumers on
+        # any supported Bazel version wanting to use the default versions of the deps below (e.g. `rules_apple` 2.x.x).
+        #
+        # For context see discussion in this PR: https://github.com/bazel-ios/rules_ios/pull/793
         _maybe(
             http_archive,
             name = "build_bazel_rules_swift",
@@ -180,9 +184,8 @@ def _rules_ios_bzlmod_dependencies(load_rules_apple_2_dependencies = False):
         )
 
     else:
-        # Bazel 6+ codepath - you can use the 2_x version of rules_apple here if
-        # you'd like to set it. If you want to use a different rules_apple than
-        # we have here - define rules_apple_api - to the value you'd like
+        # (Bazel 6.x.x, rules_apple 3.x.x) codepath - you can use the 2.x.x version of rules_apple if
+        # you'd like either by setting `load_rules_apple_2_dependencies = True` or defining `rules_apple_api` in your WORKSPACE.
         _maybe(
             http_archive,
             name = "build_bazel_rules_swift",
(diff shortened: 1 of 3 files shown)


luispadron (approved):

rules/repositories.bzl
```diff
@@ -149,7 +151,7 @@ def _rules_ios_bzlmod_dependencies():
             ],
             sha256 = "c6966ec828da198c5d9adbaa94c05e3a1c7f21bd012a0b29ba8ddbccb2c93b0d",
         )
-    elif bazel_version.major == "6" and bazel_version.minor == "1":
```
justinseanmartin, line 152: @thiagohmcruz and I discussed offline - This feels like a weird special behavior to have for handling the CI use-case ([per this comment](https://github.com/bazel-ios/rules_ios/pull/773/files#r1362497302)). I like that the proposed change makes the rules_apple 2.x support a bit more explicit in the API contract. Before this change, someone using 6.1 and accidentally getting some weird default dependencies would probably be confusing.

I think we're going to omit the Bazel version check here and have this be defaults that we know work for rules_apple 2.x based on an explicit `use_rules_apple_2 = True` param. We're also going to reset this back to the values used prior to #773 (rules_apple v2.5.0 and rules_swift v1.10.0) as the defaults. For CI, Thiago is going to mutate the WORKSPACE file in the CI script to pass `use_rules_apple_2 = True`. Separately to this, we could bump everything to run on Bazel 6.4 as well and things should continue to work.

thiagohmcruz:
@justinseanmartin To confirm the `sed` to use `rules_apple` 2.x on CI + Bazel 6.1.2 I checked the events [here](https://app.buildbuddy.io/invocation/bad5e9e7-c8c2-4023-9652-de9902368b49#raw) and saw this:
```json
  {
    "id": {
      "fetch": {
        "url": "https://github.com/bazelbuild/rules_apple/releases/download/2.5.0/rules_apple.2.5.0.tar.gz"
      }
    },
    "fetch": {
      "success": true
    }
  },
```

justinseanmartin (approved):
It might be worth adding a comment on on the elsif block and updating the one on the else block. You technically don't need to specify `load_rules_apple_2_dependencies` in order to load the rules_apple 2.x, given the `_maybe()`'s, this is just a convenience for folks that want to use the defaults and making the rules_apple 2.x support a bit more clear in the API contract.

rules/repositories.bzl
luispadron: Thank you both, that's easier to reason about

mattrobmattrob (approved):
Think this should be a major version bump since we're changing the public API? I'm curious how many folks are using this definition vs. the `maybe` support outside of Block.

luispadron:
> Think this should be a major version bump since we're changing the public API? I'm curious how many folks are using this definition vs. the `maybe` support outside of Block.

I personally wouldn't consider that public api, IMO it shouldn't have been merged to begin with since again users in Bazel can simply define the version they want to use. Ideally, we wouldn't have this at flag all. 

If we wanted to test in CI with 6.x and rules_apple 2 we can configure that in CI

thiagohmcruz:
+1 to treating this as an improved contract but not necessarily a major version bump (no strong opinions though if folks want to do so). To hard code it to "Bazel 6.1.x" was a bit hack-y and would bite folks on that version without them knowing. With this new flag at least it's now an opt-in for "I'm using whatever Bazel version <=6.x.x and I want to use `rules_apple` 2.x".

mattrobmattrob:
I don't feel strongly. Please just call it out in the release notes. 🙏

PR #801, opened 2023-10-31
Update precompiled_apple_resource_bundle to not include executable_name

### LINKED ISSUE #800: precompiled_apple_resource_bundle causing AppStore rejection
Hello!

My app started getting rejected from AppStore after updating to rules_ios 3.1.0 and rules_apple 3.1.1, with error:

```
ITMS-90207: Invalid Bundle - The bundle at '/Payload/MyApp.app/AccountStrings.bundle' does not contain a bundle executable.
```

The Info.plist I provide to `precompiled_apple_resource_bundle`:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>CFBundleName</key>
	<string>$(PRODUCT_NAME)</string>
	<key>CFBundleVersion</key>
	<string>1.0</string>
	<key>CFBundleShortVersionString</key>
	<string>1.0</string>
	<key>CFBundleIdentifier</key>
	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
	<key>CFBundlePackageType</key>
	<string>BNDL</string>
</dict>
</plist>
```

Checking the compiled Info.plist inside the .bundle, there is a new key which is probably causing the rejection:

```
    <key>CFBundleExecutable</key>
    <string></string>
```

This key doesn't exist when using `apple_resource_bundle` from rules_apple.

I used rules_ios 2.1.0 and rules_apple 2.5.0 before and never had this issue before.

luispadron:
Fixes #800

Recent changes have made it so that `bundle_name` OR `executable_name` are used for the `CFExecutable` plist substitution: https://github.com/bazelbuild/rules_apple/commit/f909b0b6f31dee7ee9ba9998a92404eaeb7af727#diff-fd21ff183af5a0af8861d50117a7a377b04a0b7d9924600f424261b7db57e2deL279

This causes App store rejection, we should set `include_executable_name` to `False` to ensure its not substituted/included in the plist.

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -56,10 +56,14 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         platform_prerequisites_version_args = {
             "build_settings": None,
         }
+        rules_api_3_resource_partials_args = {
+            "include_executable_name": False,  # Must be set to False or bundle_name is now used if executable_name is None
+        }
     else:
         platform_prerequisites_version_args = {
             "disabled_features": ctx.disabled_features,
         }
+        rules_api_3_resource_partials_args = {}
 
     platform_prerequisites = platform_support.platform_prerequisites(
         apple_fragment = ctx.fragments.apple,
@@ -90,6 +94,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         ),
         rule_label = fake_rule_label,
         version = None,
+        **rules_api_3_resource_partials_args
     )
 
     apple_mac_toolchain_info = ctx.attr._toolchain[AppleMacToolsToolchainInfo]


rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -90,6 +94,7 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         ),
         rule_label = fake_rule_label,
         version = None,
+        **partials_args
```
mattrobmattrob, line 97: Can you name this something slightly different than the dict var that you're creating for ease of understanding?

mattrobmattrob (approved):

thiagohmcruz (approved):

rules/precompiled_apple_resource_bundle.bzl
mattrobmattrob: I like that!

luispadron: Updated, and added comment for why this needs to be `False`

PR #802, opened 2023-11-01
Add option to disable modulemap generation

luispadron:
When maintaining two build systems (Xcode and Bazel) it's useful to be able to configure what is and isn't generated. In our case, we only generate modulemaps for frameworks which enable this feature in our Xcode setup. In bazel however, this file is generated regardless.

This leads to build inconsistency where Xcode fails to build because a missing modulemap is not found while Bazel continues to pass since these are always generated.

Instead, we should allow users to enable/disable this behavior. With this change `generate_default_modulemap` will be defaulted to `True` (so no changes here) unless you set to `False` in which case no modulemap is generated.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -552,6 +552,7 @@ def apple_library(
     # * Optionally, allow the consumers to set generate_default_umbrella_header to False, so the
     #   generated umbrella header does not contain any imports
     generate_default_umbrella_header = kwargs.pop("generate_default_umbrella_header", True)
+    generate_default_modulemap = kwargs.pop("generate_default_modulemap", True)
     cc_copts = kwargs.pop("cc_copts", [])
     additional_cc_copts = []
     swift_copts = kwargs.pop("swift_copts", [])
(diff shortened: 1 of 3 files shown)


luispadron:
Converting to a draft, not sure this solves our specific issue just yet.

rules/library.bzl
```diff
@@ -552,6 +552,7 @@ def apple_library(
     # * Optionally, allow the consumers to set generate_default_umbrella_header to False, so the
     #   generated umbrella header does not contain any imports
     generate_default_umbrella_header = kwargs.pop("generate_default_umbrella_header", True)
+    generate_default_modulemap = kwargs.pop("generate_default_modulemap", True)
```
justinseanmartin, line 555: Does the existing hook for being able to override the `modulemap_generator` enable this functionality in your `apple_library` wrapper? It'll still pass through kwargs, so if you need this set on a per-target basis, that should be possible.

luispadron: We do set `generate_default_umbrella_header` to `False` already, but since it does:

```bzl
generate_default_umbrella_header = kwargs.pop("generate_default_umbrella_header", True)
```

The kwargs don't reach the custom `modulemap_generator`. And if we pass a custom argument to the library target, rules_ios would attempt to forward to one of the libraries which fails too

luispadron:
We chose to instead catch this issue in our existing setup vs. in here. Closing this.

PR #803, opened 2023-11-07
Support more valid ios_application kwargs

luispadron:
Adds support for `codesign_inputs`, `codesignopts` and `toolchains`


rules/app.bzl
luispadron, line 1: Sorted this list too

mattrobmattrob (approved):
❤️

thiagohmcruz (approved):

PR #804, opened 2023-11-07
Fix pkg_resources module not found in CI

luispadron:
This script seems to be failing [here](https://github.com/bazel-ios/rules_ios/actions/runs/6788374457/job/18454172023?pr=803), this works locally so trying to re-install the module before.


jerrymarino (commented):
How does running this with Bazel change the invocation of the script WRT reliability

luispadron:
Looks like adding `pip install setuptools` fixes the issue

mattrobmattrob (approved):
LGTM assuming there's not some easier built in package to do the work of `pkg_resources`.

PR #805, opened 2023-11-08
Support Swift Macro

qyang-nj:
Make `apple_framework` support swift compiler plugins.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -1,5 +1,5 @@
 """Library rules"""
-
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:sets.bzl", "sets")
 load("@bazel_skylib//lib:selects.bzl", "selects")
@@ -582,7 +582,7 @@ def apple_library(
     defines = kwargs.pop("defines", [])
     testonly = kwargs.pop("testonly", False)
     features = kwargs.pop("features", [])
-    plugins = kwargs.pop("plugins", [])
+    plugins = kwargs.pop("plugins", None)
 
     for (k, v) in {"momc_copts": momc_copts, "mapc_copts": mapc_copts, "ibtool_copts": ibtool_copts}.items():
         if v:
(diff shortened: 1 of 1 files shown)


luispadron (approved):
Thank you! Mind adding a test fixture that exercises this attr 🙏🏼

jerrymarino (approved):
🚀

thiagohmcruz (approved):

rules/library.bzl
```diff
@@ -985,6 +986,7 @@ def apple_library(
             tags = tags_manual,
             defines = defines + swift_defines,
             testonly = testonly,
+            plugins = plugins,
```
jerrymarino, line 989: Perhaps adding this onto `kawrgs` `if plugins != []` will fix the CI issue

rules/library.bzl
```diff
@@ -960,6 +961,9 @@ def apple_library(
     )
 
     if swift_sources:
+        # To be backward compatible with rules_apple 2.x
+        swift_kwargs = dicts.add(kwargs, {"plugins" : plugins} if plugins else {})
```
qyang-nj, line 965: I don't want to modify `kwargs` in place as it may be used in the places later.

qyang-nj:
> Mind adding a test fixture that exercises this attr 🙏🏼
I'm not sure how to add a test just for "rules_apple 3.x" 🤔

luispadron:
> > Mind adding a test fixture that exercises this attr 🙏🏼
> 
> I'm not sure how to add a test just for "rules_apple 3.x" 🤔

@jerrymarino do you have suggestions

jerrymarino:
Is it possible you'd use a `select` statement for the `plugins` attribute you're setting; where the `select` contains a default condition for your plugins but nothing for the LTS stuff?

qyang-nj:
I'm seeing this error. Somehow I cannot use Xcode 15 for this repo. 🤔
```
Error in fail: Swift compiler plugins require Swift 5.9+
```

luispadron:
Ah we need to update CI to support that version. I think this PR is good as-is and we can follow up with a test when we can support it.

qyang-nj:
Sounds good. Thank you. I'm happy to add the tests after the CI supports it.

PR #806, opened 2023-11-08
Ignore MODULE.bazel.lock

mattrobmattrob:
This file is generated by default by bazel @ HEAD but we don't really want to manage it. This way it can be generated locally and folks can build offline if they have generated it at least once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/MODULE.bazel.lock
 bazel-*
 .idea
 tests/macos/xcodeproj/**/xcuserdata


thiagohmcruz (approved):

luispadron (approved):

luispadron:
seems fine for now but this looks to go against what's suggested here: https://bazel.build/external/lockfile

mattrobmattrob:
> seems fine for now but this looks to go against what's suggested here: https://bazel.build/external/lockfile
> 
> > Include the lockfile in version control to facilitate collaboration and ensure that all team members have access to the same lockfile, promoting consistent development environments across the project.

I'm following https://github.com/bazelbuild/rules_swift/commit/d0687c3c9957633376a7099d03e0ced183f64f75 for now. IMO, it's probably a different rationale for rulesets vs. normal repos but 🤷

luispadron:
Ah that's fair

Issue #795, opened 2023-10-27
Add support for Bazel 7


luispadron:
Bazel 7 is coming out soon (November 2023), building with `USE_BAZEL_VERSION=last_rc` we get the following error:

```sh
ERROR: /Users/lpadron/Development/rules_xcodeproj/examples/rules_ios/iOSApp/Source/BUILD:32:26: //iOSApp/Source:iOSApp: no such attribute 'output_discriminator' in 'ios_application' rule
ERROR: /Users/lpadron/Development/rules_xcodeproj/examples/rules_ios/iOSApp/Source/BUILD:32:26: errors encountered resolving select() keys for //iOSApp/Source:iOSApp
```

## Tasks

- [x] `output_discriminator` is not a valid attr: fixed in #796 
- [x] Remove `apple_grte_top` usage: https://github.com/bazelbuild/bazel/commit/fb4106bdbd23c365337ea99704921ada7b86c2df
  - `/Users/runner/work/rules_ios/rules_ios/rules/transition_support.bzl:125:36: transition inputs [//command_line_option:apple_compiler, //command_line_option:apple_grte_top] do not correspond to valid settings`
- [x] Remove `apple_compiler` usage: https://github.com/bazelbuild/bazel/commit/1acdfc422e724b4fe12c7bf5248086ab514ec4be
  - `/Users/runner/work/rules_ios/rules_ios/rules/transition_support.bzl:125:36: transition inputs [//command_line_option:apple_compiler, //command_line_option:apple_grte_top] do not correspond to valid settings`
- [x] Add `toolchains` argument to `rule`s
  - `Error in fail: In order to use find_cpp_toolchain, you must include the '@bazel_tools//tools/cpp:toolchain_type' in the toolchains argument to your rule.`
- [x] `sysroot` issues on CI
  - `clang: error: using sysroot for 'iPhoneSimulator' but targeting 'MacOSX' [-Werror,-Wincompatible-sysroot]`
  - [x] Add `manual` tags to `objc_library`: https://bazelbuild.slack.com/archives/CD3QY5C2X/p1701890904270059
- [x] `ld: Undefined symbols`, we'll need to audit `ObjcProvider` for use in Bazel 7 given: https://github.com/bazelbuild/bazel/issues/17377

mattrobmattrob:
Are you working on this actively, @luispadron?

luispadron:
[...]

PR #797, opened 2023-10-30
[DNR] Move CI to Bazel 7

mattrobmattrob:
Testing for https://github.com/bazel-ios/rules_ios/issues/795.

PR #807, opened 2023-11-10
Add `toolchains` to rules that use `@bazel_tools//tools/cpp:toolchain_type`

mattrobmattrob:
Fix the following from https://github.com/bazel-ios/rules_ios/issues/795:
```
Error in fail: In order to use find_cpp_toolchain, you must include the '@bazel_tools//tools/cpp:toolchain_type' in the toolchains argument to your rule.
```

### DIFF
--- a/rules/apple_patched.bzl
+++ b/rules/apple_patched.bzl
@@ -144,6 +144,7 @@ def _apple_framework_import_modulemap_impl(ctx):
 
 _apple_framework_import_modulemap = rule(
     implementation = _apple_framework_import_modulemap_impl,
+    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
     fragments = ["apple"],
     attrs = {
         "legacy_target": attr.label(
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -1191,6 +1191,7 @@ def _apple_framework_packaging_impl(ctx):
 
 apple_framework_packaging = rule(
     implementation = _apple_framework_packaging_impl,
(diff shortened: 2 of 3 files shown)


thiagohmcruz (approved):

rules/framework.bzl
```diff
@@ -1191,6 +1191,7 @@ def _apple_framework_packaging_impl(ctx):
 
 apple_framework_packaging = rule(
     implementation = _apple_framework_packaging_impl,
+    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
```
jerrymarino, line 1194: Any downsides or fallout from adding this? I wondered if we need to add it to other rules; hmap, middlemen, etc

mattrobmattrob: I don't know of any. But I still haven't narrowed down the reasons for 👇 when using Bazel 7 with this change (like in https://github.com/bazel-ios/rules_ios/pull/797):
```
clang: error: using sysroot for 'AppleTVSimulator' but targeting 'MacOSX' [-Werror,-Wincompatible-sysroot]
```

rules/apple_patched.bzl
luispadron, line 1: Unrelated but should we delete this file in a follow-up, it appears to be unused.

luispadron (approved):
I wonder if this has to do with the move to apple_support toolchain

> The Apple CC toolchain in this repository provides toolchains for building for Apple platforms besides macOS. Since Bazel 7 this toolchain is required when targeting those platforms but the toolchain also supports Bazel 6.

https://github.com/bazelbuild/apple_support#toolchain-setup

PR #816, opened 2023-12-25
Karim/fix objc library toolchain resolution transition

karim-alweheshy:
[Remove objc_library transition](https://github.com/bazelbuild/bazel/issues/16870)
[Bazel changes](https://github.com/bazelbuild/bazel/commit/0f34e76f45a2d9a6407431ea3a371a9d62814527)
[rules_apple changes](https://github.com/bazelbuild/rules_apple/pull/2257/files)


karim-alweheshy:
A step for bazel 7 migration https://github.com/bazel-ios/rules_ios/issues/795

rules/transition_support.bzl
```diff
@@ -7,6 +7,51 @@ load(
...
+    "ios": "//command_line_option:ios_multi_cpus",
+    "macos": "//command_line_option:macos_cpus",
+    "tvos": "//command_line_option:tvos_cpus",
+    "visionos": "//command_line_option:visionos_cpus",
+    "watchos": "//command_line_option:watchos_cpus",
+}
+
+# Could not load it directly from bazel_build_apple_support
+# As we support multiple versions of rules_apple_support
+# Where some don't have the file. So we copied it
```
jszumski, line 20: Can you add a TODO to switch to directly loading it when the minimum apple_support version is X?

karim-alweheshy: Added ✅

rules/transition_support.bzl
```diff
@@ -7,6 +7,52 @@ load(
...
+    "tvos": "//command_line_option:tvos_cpus",
+    "visionos": "//command_line_option:visionos_cpus",
+    "watchos": "//command_line_option:watchos_cpus",
+}
+
+# TODO: Should be loaded from bazel_build_apple_support when we migrate to v1.10.0
+# Could not load it directly from bazel_build_apple_support
+# As we support multiple versions of rules_apple_support (v2 and v3)
+# Where <v3.1.0 don't have the updated dependency. So we copied it
+# https://github.com/bazelbuild/apple_support/blob/d87e8b07f3345e750834dbb6ce38c7c7d3b8b44b/configs/platforms.bzl#L108
```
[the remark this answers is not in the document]
thiagohmcruz: IMO if dropping Bazel 5 makes this (and other stuff) easier we should consider doing it first then. My main concern are folks still relying on it but I guess there's not much we can do besides let them know in the Bazel slack that this is happening.

Should we focus on dropping Bazel 5 then before continuing this particular PR?

luispadron:
@karim-alweheshy I'll rebase and merge this once #835 lands.

MODULE.bazel
```diff
@@ -70,6 +70,12 @@ use_repo(
     "tart",
 )
 
+rules_ios_bazel_version_deps = use_extension("//rules:module_extensions.bzl", "rules_ios_bazel_version_deps")
+use_repo(
+    rules_ios_bazel_version_deps,
+    "rules_ios_bazel_version",
+)
+
```
luispadron, line 78: Can we instead use: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/versions_doc.md

rules/internal/bazel_version.bzl
luispadron, line 1: Similar comment here, let's remove this custom logic and use skylib: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/versions_doc.md

rules/transition_support.bzl
```diff
@@ -74,6 +113,21 @@ def _min_os_version_or_none(attr, attr_platforms, platform, attr_platform_type):
 
     return None
 
+def _bazel_7_objc_library_transition_values(settings, platform_type):
```
luispadron, line 116: Do you know if this can be done for Bazel 6 too? Thats the min. version we support now so if we can remove this conditionality that would be best

rules/internal/bazel_version.bzl
karim-alweheshy: I added a comment explaining why we need it and why skylib versions won't work for us

MODULE.bazel
karim-alweheshy: I added a comment explaining why we need it and why skylib versions won't work for us

rules/transition_support.bzl
karim-alweheshy: apparently it works on bazel 6 as well.. check removed

rules/transition_support.bzl
```diff
@@ -112,10 +162,17 @@ def _apple_rule_transition_impl(settings, attr):
         "//command_line_option:tvos_minimum_os": _min_os_version_or_none(attr, attr_platforms, "tvos", platform_type),
         "//command_line_option:watchos_minimum_os": _min_os_version_or_none(attr, attr_platforms, "watchos", platform_type),
     }
+    ret.update(_objc_library_transition_values(settings, platform_type))
     return ret
 
 _supports_visionos = hasattr(apple_common.platform_type, "visionos")
 
+# `--apple_compiler` was removed in https://github.com/bazelbuild/bazel/commit/1acdfc422e724b4fe12c7bf5248086ab514ec4be
+_supports_clo_apple_compiler = _bazel_major_version < 7
```
luispadron, line 171: Can we just delete it completely, similarly to: https://github.com/bazelbuild/rules_apple/pull/2040

rules/transition_support.bzl
```diff
@@ -112,10 +162,17 @@ def _apple_rule_transition_impl(settings, attr):
...
+
+# `--apple_grte_top` was removed in https://github.com/bazelbuild/bazel/commit/fb4106bdbd23c365337ea99704921ada7b86c2df
+_supports_clo_apple_grte_top = _bazel_major_version < 7
```
luispadron, line 174: Same here: https://github.com/bazelbuild/rules_apple/pull/1608 I don't we need this when supporting Bazel 6+

luispadron (commented):
I'll get PRs up to remove the `apple_compiler`/`gre_top` usage

rules/transition_support.bzl
```diff
@@ -1,5 +1,25 @@
...
+}
+
+_CPU_TO_DEFAULT_PLATFORM_FLAG = {
+    cpu: "@build_bazel_apple_support//platforms:{}_platform".format(
+        platform_name,
+    )
+    for cpu, platform_name in CPU_TO_DEFAULT_PLATFORM_NAME.items()
+}
+
 def _current_apple_platform(apple_fragment, xcode_config):
```
luispadron, line 23: This now matches `rules_apple`, thank you!

rules/transition_support.bzl
```diff
@@ -74,6 +102,19 @@ def _min_os_version_or_none(attr, attr_platforms, platform, attr_platform_type):
 
     return None
 
+def _objc_library_transition_values(settings, platform_type):
```
luispadron, line 105: nit: im not sure the best name here but since this is related to the transition of apple platform moving to rules vs. defined in Bazel maybe something like this is more clear:

```suggestion
def _apple_platforms_base_values(settings, platform_type):
```

Alternatively we could try to match: https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/transition_support.bzl#L440-L455

luispadron (approved):
Thank you for iterating on this!

jszumski (approved):
Thank you!

PR #817, opened 2023-12-26
Karim/fix test framework archs

karim-alweheshy:
Rebuilding Basic and NestedFrameworks for ios_sim_arm64_x86_64
And repointing public TensorFlowLiteC to be on the same arch as well

Linking non matching archs with the target is an error on bazel 7


karim-alweheshy:
done in https://github.com/bazel-ios/rules_ios/pull/818

PR #818, opened 2023-12-26
Update CI to Xcode 15.1 on macos 13.0

karim-alweheshy:
Updated CI runners to [macos-13](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md)
Update Xcode to 15.1.0 from 14.2.0
Update min iOS version to 12.0 from 10.0 and 11.0 and remove Host-App pre 12.0
Updated iOS sdk to 17.2 and devices to sims from the runner
Update swift_version from 5.1 to 5.9
Update xcodeproj installer, as it was pointing to a static lib for dynamic lib compilation and fixed path to remove `//`
Updated compile frameworks to the correct arch `ios-sim-arm64-x86_64`
Updated third party libs to be on the right arch `ios-sim-arm64-x86_64`


karim-alweheshy:
A step for [bazel 7 migration](https://github.com/bazel-ios/rules_ios/issues/795)

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
luispadron, line 1: Was this change intentional? Seems to break the path

luispadron (commented):
The xcspec changes / ci look good to me. 

I'm not sure about the project changes though, it seems like the path was broken? @thiagohmcruz @jerrymarino

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/install.sh
karim-alweheshy: without this path change it was pointing to a static framework when it should point to the dynamic framework. This dynamic framework can at times be compressed into a zip i.e. doesn't exist yet on disk. in that case i added the zip to be unzipped later in the same script

PR #819, opened 2023-12-26
Allow frameworks parameter to be used in ios_application

codeman9:
- Before this change, using the "frameworks" parameter in ios_application resulted in an error: "Error in ios_application: rule(...) got multiple values for parameter 'frameworks'" because we are passing "application_kwargs" into the rules_apple rule which has the "frameworks" parameter copied over from "kwargs". Making this change actually pops the parameter from the correct dictionary which will remove the duplication error.

### DIFF
--- a/tests/ios/frameworks/dynamic/BUILD.bazel
+++ b/tests/ios/frameworks/dynamic/BUILD.bazel
@@ -17,6 +17,22 @@ ios_application(
 )
 
 ios_application(
+    name = "AppWithFrameworks",
+    srcs = ["App/main.m"],
+    bundle_id = "com.example.app",
+    frameworks = [
+        "//tests/ios/frameworks/dynamic/a",
+    ],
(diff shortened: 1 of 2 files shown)


tests/ios/frameworks/dynamic/BUILD.bazel
```diff
@@ -16,6 +16,22 @@ ios_application(
...
+    name = "AppWithFrameworks",
+    srcs = ["App/main.m"],
+    bundle_id = "com.example.app",
+    frameworks = [
+        "//tests/ios/frameworks/dynamic/a",
+    ],
+    minimum_os_version = "10.0",
+    visibility = ["//visibility:public"],
+    # This should automatically bake in a,b
+    deps = [
```
luispadron, line 29: Thanks!

luispadron (approved):

PR #820, opened 2023-12-26
Allow select in test_host parameter

codeman9:
Previously, using a `select` in the `test_host` parameter would result in the error:

```
ERROR: rules_ios/tests/ios/unit-test/BUILD.bazel:51:14: //tests/ios/unit-test:ExplicitHostedWithSelect.dep_middleman: expected value of type 'string' for element 0 of attribute 'test_deps' in 'dep_middleman' rule, but got select({"//tests/ios/unit-test:use_other_test_host": "//rules/test_host_app:iOS-9.3-AppHost", "//conditions:default": "//rules/test_host_app:iOS-14.0-AppHost"}) (select)
```

I simply removed the `test_deps` parameter as that wasn't really being used for anything of value that I could see. It was only ever being set to the `test_host` parameter or an empty list of which neither contain the `AvoidDepsInfo` information that was being deduplicated.

### DIFF
--- a/rules/internal/framework_middleman.bzl
+++ b/rules/internal/framework_middleman.bzl
@@ -216,11 +216,6 @@ def _dep_middleman(ctx):
             for lib_dep in dep[AvoidDepsInfo].libraries:
                 _collect_providers(lib_dep)
 
-    # Pull AvoidDeps from test deps
-    for dep in (ctx.attr.test_deps if ctx.attr.test_deps else []):
-        if AvoidDepsInfo in dep:
-            _process_avoid_deps(dep[AvoidDepsInfo].libraries)
-
     # Merge the entire provider here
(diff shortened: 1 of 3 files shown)


luispadron (approved):
Making this selectable again is nice, thank you!

@jerrymarino do you know why this was added / who is using it?

mattrobmattrob (approved):
@ethan-gates or @ob, do you all use `test_deps` in your setup?

ob:
> @ethan-gates or @ob, do you all use `test_deps` in your setup?

Not that I'm aware of, good riddance I say.

thiagohmcruz (approved):
Should be fine, I don't see us using `test_deps` anywhere in our codebase. Thx for adding a test case for it too!

mattrobmattrob:
I'm going to merge this since we have the confirmed lack of usage from the large stakeholders. Please comment here or open an issue if this change causes issues for any future readers.

PR #825, opened 2024-01-03
Ensure -import-underlying-module gets into the generated Xcode project

thiagohmcruz:
So [OTHER_SWIFT_FLAGS](https://github.com/bazel-ios/rules_ios/blob/883cc859daffb1f02bd0e153fca39177d2fa7eb4/rules/legacy_xcodeproj.bzl#L836) gets `-import-underlying-module` if present. Without this indexing might fail in mixed modules.

I'm not sure why the aspect is not propagating the direct `copts` here and I'd rather not propagate everything and risk breaking the experience for folks. This legacy generator is going to be less and less maintained over time so not worth the effort of digging more here IMO (open to suggestions if anyone feels strongly about fixing all things `copts` in follow PRs).

### DIFF
--- a/rules/legacy_xcodeproj.bzl
+++ b/rules/legacy_xcodeproj.bzl
@@ -127,6 +127,13 @@ def _xcodeproj_aspect_collect_swift_copts(deps, ctx):
     copts = None
     if ctx.rule.kind == "swift_library":
         copts = _make_swift_copts(deps)
+
+        # Ensures `-import-underlying-module` gets into the generated Xcode project if applied,
+        # otherwise indexing might fail in mixed modules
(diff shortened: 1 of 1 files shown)


jerrymarino (approved):

mattrobmattrob (approved):
We should probably work towards a plan where this is removed assuming you all are actually planning to migrate.

thiagohmcruz:
> We should probably work towards a plan where this is removed assuming you all are actually planning to migrate.

@mattrobmattrob We're dedicating less time to maintain it, realistically I think it will be less and less used over time unless there's someone actively improving it besides us. I'd like to know if there's anyone hard-relying on it today too as part of this so we don't break anything for them, so some kind of survey would help. Maybe this should be a discussion in GH and we can send links asking for folks input in Slack?

That said, we're still using it but in the process of migrating to `rules_xcodeproj`, so our our end I wouldn't feel the need to kick this discussion off at least until we're fully migrated.

PR #827, opened 2024-01-25
[DNR][DNM] LLDB crash Xcode 15.x

thiagohmcruz:
Works around a LLDB crash happening on Sonoma + Xcode 15.x.

Pulling into `rules_ios` a change @jerrymarino did internally to fix the crash for us. I don't think there's an easy way to find a small repro since it was hard to repro even in our internal setup.

@jerrymarino let me know if you want to add more details here in the comments, maybe you can share a more detailed explanation of what was going on for future readers?

### DIFF
--- a/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
+++ b/tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
@@ -103,3 +103,7 @@ then
 settings append target.source-map ./external/ "$BAZEL_EXTERNAL_DIRNAME"
 END
 fi
+
+# Xcode 15.x: Fixes a problem in the flag parser of LLDB 15.1 potentially causing LLDB to crash
+sed -i '' 's, -vfsoverlay, -vfsoverlay ,g' $BAZEL_LLDB_INIT_FILE
+sed -i '' 's, -ivfsoverlay, -ivfsoverlay ,g' $BAZEL_LLDB_INIT_FILE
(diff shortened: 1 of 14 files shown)


thiagohmcruz:
This was supposed to be a draft, my bad for the notifications others might have received

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
```diff
@@ -103,3 +103,7 @@ then
 settings append target.source-map ./external/ "$BAZEL_EXTERNAL_DIRNAME"
 END
 fi
+
+# Xcode 15.x: Fixes a problem in the flag parser of LLDB 15.1 potentially causing LLDB to crash
+sed -i '' 's, -vfsoverlay, -vfsoverlay ,g' $BAZEL_LLDB_INIT_FILE
```
jerrymarino, line 108: Is there a concern that this causes issues with indexing?

jerrymarino (commented):
Thanks for getting this in @thiagohmcruz

tests/ios/xcodeproj/Single-Static-Framework-Project.xcodeproj/bazelinstallers/lldb-settings.sh
thiagohmcruz: >Is there a concern that this causes issues with indexing?

I tested basic indexing stuff after applying this and it was fine, but I do want to run a more complete set of manual tests before calling it "safe"

>Should we conditionally set this for Xcode 15.1?

Good callout, yes we should

thiagohmcruz: This is not the final approach btw, once I remove it from draft I'll ask for new reviews. Ignore what's in here for now.

thiagohmcruz:
Closing this for now as I'm planning to re-visit it in a newer Xcode version

PR #828, opened 2024-01-26
Generate starlark_cpu_profile for builds in xcode

AngelaGuardia:
**What changed**
- Added the `--starlark_cpu_profile` flag to generate the profile for builds in xcode
- Updated fixture files

### DIFF
--- a/tools/xcodeproj_shims/build-wrapper.sh
+++ b/tools/xcodeproj_shims/build-wrapper.sh
@@ -18,6 +18,10 @@ if [ $BAZEL_PROFILE_ENABLED -gt 0 ]; then
     BAZEL_BUILD_OPTIONS+=("--profile=$BAZEL_PROFILE_FILENAME")
 fi
 
+if [ $BAZEL_STARLARK_CPU_PROFILE_ENABLED -gt 0 ]; then
+    BAZEL_BUILD_OPTIONS+=("--starlark_cpu_profile=$BAZEL_STARLARK_CPU_PROFILE_FILENAME")
+fi
+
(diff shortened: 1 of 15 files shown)


tools/xcodeproj_shims/build-wrapper.sh
```diff
@@ -16,6 +16,7 @@ fi
 
 if [ $BAZEL_PROFILE_ENABLED -gt 0 ]; then
     BAZEL_BUILD_OPTIONS+=("--profile=$BAZEL_PROFILE_FILENAME")
+    BAZEL_BUILD_OPTIONS+=("--starlark_cpu_profile=$BAZEL_STARLARK_CPU_PROFILE_FILENAME")
```
thiagohmcruz, line 19: I think my concern here is that the `--profile` is always generated even if we don't specify it via `--profile` so we're not introducing overhead by adding it.

That's different with `--starlark_cpu_profile` though, IIRC it's not always created by Bazel. Meaning this will introduce additional overhead (even if minimal depending on the workspace). But still, I think this should be separate opt-in behaviour instead of being part of `BAZEL_PROFILE_ENABLED`. Thoughts?

AngelaGuardia: I agree. This should definitely be opt-in. I can create a `BAZEL_STARLARK_CPU_PROFILE_ENABLED` flag to control that.

thiagohmcruz (approved):

thiagohmcruz:
The CI failure here is very likely unrelated to your changes. I also hit that same err in a draft PR of mine. We should land this PR to unblock you and tackle this specific CI issue separately. Mind filing an issue @AngelaGuardia?

justinseanmartin:
@AngelaGuardia is going to log an issue to fix CI, but admin merging this for now to unblock.

PR #834, opened 2024-02-13
Drop support for Bazel 5 & rules_apple 2

luispadron:
This stops testing Bazel 5/rules_apple 2 in CI and bumps the support Bazel version in MODULE.bazel. It does not include any of the Bazel 5 related cleanup that is now possible. This is for an upcoming version 4.0.0.

Adds a table like the following to the read me:

## Supported Bazel versions

| Bazel release | Minimum supported rules version | Final supported rules version
|:-------------------:|:-------------------------:|:-------------------------:
| 6.* | 2.0.0 | current
| 5.* | 1.0.0 | 3.2.2
| 4.* | 1.0.0 | 1.0.0

## Supported [rules_apple](https://github.com/bazelbuild/rules_apple) versions

| rules_apple release | Minimum supported rules version | Final supported rules version
|:-------------------:|:-------------------------:|:-------------------------:
| 3.* | 3.* | current
| 2.* | 2.* | 3.2.2
| 1.* | 1.0.0 | 3.2.2

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -28,10 +28,6 @@ build --features swift.index_while_building
 # Prevents leaking unexpected binaries via PATH to tests
 build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
-# Enable device deps when the build matches the arm64 simulator
-# Bazel 5 - matches ios_sim_arm64 CPU, Bazel 4 - needs "override" flags
-build --features apple.arm64_simulator_use_device_deps
-
 # Delete the VM test suite for github
 build --deleted_packages tests/ios/vmd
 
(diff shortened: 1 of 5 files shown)


.bazelrc
```diff
@@ -28,10 +28,6 @@ build --features swift.index_while_building
 # Prevents leaking unexpected binaries via PATH to tests
 build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
-# Enable device deps when the build matches the arm64 simulator
-# Bazel 5 - matches ios_sim_arm64 CPU, Bazel 4 - needs "override" flags
-build --features apple.arm64_simulator_use_device_deps
```
luispadron, line 33: I moved this to the arm64 CI job where it's tested. This way we test without it by default to ensure we catch issues without this flag enabled.

.bazelrc
```diff
@@ -53,7 +49,9 @@ build:ios_lldb_test --define=apple.experimental.tree_artifact_outputs=1
 # - iOS multi arch test configuration
 build:ios_multi_arch_test --config=ios
 build:ios_multi_arch_test --ios_minimum_os=10.2
-build:ios_multi_arch_test --ios_multi_cpus=i386,x86_64
+# i386 was removed on rules_apple 3.x.x - this test case needs reworking
+# to exemplify fat binaries on the latest version
+build:ios_multi_arch_test --ios_multi_cpus=x86_64
```
jszumski, line 54: Would adding `arm64` to the list suffice?

luispadron: I tried that and it failed with an unknown architecture issue, i think potentially theres a bug in `transition_support.bzl` within rules_ios we need to look at after this

thiagohmcruz (approved):

jszumski (approved):

mattrobmattrob:
Awesome job!

PR #835, opened 2024-02-13
Remove Bazel 5 and rules_apple 2 related code

luispadron:
This PR removes the existing Bzlmod extensions required to work around the API changes between rules apple 2 and rules apple 3. Additionally, removes any branching we were doing for either Bazel 5 or rules_apple < 3.


rules/transition_support.bzl
```diff
@@ -123,15 +116,6 @@ def _apple_rule_transition_impl(settings, attr):
...
 
-_bazel_version = get_bazel_version(bazel_version)
-_bazel_major_version = int(_bazel_version.major)
-
-# `--apple_compiler` was removed in https://github.com/bazelbuild/bazel/commit/1acdfc422e724b4fe12c7bf5248086ab514ec4be
-_supports_clo_apple_compiler = _bazel_major_version < 7
-
-# `--apple_grte_top` was removed in https://github.com/bazelbuild/bazel/commit/fb4106bdbd23c365337ea99704921ada7b86c2df
-_supports_clo_apple_grte_top = _bazel_major_version < 7
-
```
luispadron, line 134: These are both related and removed in Bazel and now rules_ios for the same reason as: https://github.com/bazelbuild/rules_apple/pull/2040

.bazelrc
```diff
@@ -28,10 +27,6 @@ build --features swift.index_while_building
 # Prevents leaking unexpected binaries via PATH to tests
 build --test_env=PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
-# Enable device deps when the build matches the arm64 simulator
-# Bazel 5 - matches ios_sim_arm64 CPU, Bazel 4 - needs "override" flags
-build --features apple.arm64_simulator_use_device_deps
```
luispadron, line 33: I moved this to the CI job where it's tested. This way we test without it by default to ensure we catch issues without this flag enabled.

jszumski (approved):

thiagohmcruz (approved):
LGTM. Questions:

*The idea is to cherry pick new things into older releases correct? Should this process be documented in the repo for current and future maintainers?
* Are we planning to PSA this in the Bazel slack? I feel like we should since it's a big-ish change?

luispadron:
I'll add some contributing docs on cherry picking and create a branch for v3. 

We've already asked folks if they're on Bazel 5 previously but I'll announce when I merge and release this

PR #836, opened 2024-02-23
Fix BCR presubmit for new requirements

luispadron:
Now requires bazel version to be set, also switches to arm64 for the machines

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,7 +1,9 @@
-bazel: 6.4.0
+matrix:
+  bazel: ["6.x"]
 tasks:
   verify_build_targets:
     name: Verify Build targets on macOS
-    platform: macos
+    platform: macos_arm64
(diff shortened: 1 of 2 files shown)


.bcr/presubmit.yml
```diff
@@ -1,7 +1,9 @@
-bazel: 6.4.0
+matrix:
+  bazel: ["6.x"]
```
mattrobmattrob, line 2: Want to be representative of our CI?
```suggestion
matrix:
  bazel: ["6.x", "7.x"]
```

mattrobmattrob (approved):

.bcr/presubmit.yml
luispadron: We dont support 7.x yet though

mattrobmattrob: 🤦

jszumski (approved):

PR #837, opened 2024-02-29
Update to use Xcode 15.2

luispadron:
This is a prerequisite for Bazel 7 (#795).

Changes: 

- Update xcspecs for Xcode 15.2 (see the second commit for those changes separately)
- Update CI machines to macOS 14 which includes Xcode 15 machines
- Update CI to use Xcode 15.2
- Update default runners for iOS 17.2 since thats now the default installed OS
- Update macOS xcconfig tests for Xcode 15
- Update .xcodeproj tests to use arm64 as CI now runs on M1 machines


data_generators/xcspec_extractor.py
```diff
@@ -32,7 +32,7 @@ def xcspec_from_file(path):
 eval_bzl_path = sys.argv[4]
 xcode = os.path.dirname(os.path.dirname(developer_dir))
 xcode_plugins = os.path.join(
-    xcode, "Contents/PlugIns/XCBSpecifications.ideplugin/Contents/Resources"
+    xcode, "Contents/Developer/Library/Xcode/Plug-ins/XCBSpecifications.ideplugin/Contents/Resources"
```
luispadron, line 35: This is the new path to `XCBSpecifications.ideplugin`

rules/test/lldb/lldb_sim_runner.py
```diff
@@ -209,8 +209,7 @@ def GetStartedAppPid(self):
     def Fail(self):
         traceback.print_exc()
         logger.error("FAIL")
-        with ctxlock:
-            self.status = -1
+        self.status = -1
```
luispadron, line 212: Unrelated but `ctxlock` is undefined so when a lldb test would fail this would mask the actual error.

tests/macos/xcconfig/tests.bzl
luispadron, line 1: These changes were required after updating xcspecs, they don't really make sense to me but 🤷🏼

luispadron:
Once this merges i'll mark the 15.2 jobs required in GitHub again

luispadron:
@thiagohmcruz Thoughts on why the .xcodeproj tests seem to fail due to an interrupt? Its coming from the `gets` call in `output-processor.rb`

mattrobmattrob (approved):

luispadron:
Got past the build failure by updating the sim to use iPhone 15, now just need to update the fixtures for arm64 since `macos-14` are M1 machines

jszumski (approved):

PR #841, opened 2024-03-04
Update CI for latest actions and to use matrix

luispadron:
Updates GitHub CI to make it easier to add more Bazel/Xcode Versions.

Changes:

- Update `checkout` and `upload-artifacts` actions to `v4`, this fixes warnings we get on every PR for deprecated Node versions
- Add `matrix` strategy for all jobs, this adds a `bazel_version` and `xcode_version` which each job will be configured to run against. We can add Bazel 7 once #795 is done.
- Merge the `--config=vfs` and non-vfs job, this can be set with a custom matrix without duplicating the job code.
- Remove `--remote-cache` flag, this was set by every job already.

## Example:

Screenshot 2024-03-04 at 10 59 36 AM

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -52,9 +52,6 @@ build:ios_multi_arch_test --ios_minimum_os=10.2
 # to exemplify fat binaries on the latest version
 build:ios_multi_arch_test --ios_multi_cpus=x86_64
 
-# - Remote cache configuration
-build:ci_with_caches --config=ci --config=remote_cache
-
 build:remote_cache --bes_backend=grpcs://remote.buildbuddy.io
 build:remote_cache --bes_results_url=https://app.buildbuddy.io/invocation/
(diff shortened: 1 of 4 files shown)


.bazelrc
```diff
@@ -52,9 +52,6 @@ build:ios_multi_arch_test --ios_minimum_os=10.2
 # to exemplify fat binaries on the latest version
 build:ios_multi_arch_test --ios_multi_cpus=x86_64
 
-# - Remote cache configuration
-build:ci_with_caches --config=ci --config=remote_cache
```
luispadron, line 56: This config was previously used by all jobs, seemed redundant to define another `ci` config so just using `ci` with `--remote_cache` by default now

luispadron:
If this is approved i'll update the required jobs to use these new ones in the settings

thiagohmcruz (approved):

PR #844, opened 2024-03-05
Fix duplicate symbols for test rules

luispadron:
This makes changes to `test.bzl` to remove the default `internal_test_deps` being added to all snapshot tests. This is causing duplicate symbol issues in Bazel 7 as framework search path logic has changed. In Bazel 7, the snapshot test target can't depend on the runfiles lib without introducing duplicate symbols at link time, this is because the `BazelRunfiles` target it depends on already links the runefiles lib. This updates the rule / usage of it to pass this dependency explicitly so it can be removed in the bazel 7 branch but maintained here for others who need it.

Changes:

- Dont set `interal_test_deps` to have a default value for snapshot tests
- Explicitly set each key word argument in each test factory, this stops issues like this from coming up again in the future

### DIFF
--- a/rules/test.bzl
+++ b/rules/test.bzl
@@ -142,8 +142,9 @@ def _ios_test(name, bundle_rule, test_rule, test_factory, apple_library, infopli
 
     Args:
         name: The name of the unit test.
-        test_rule: The underlying rules_apple test rule.
         bundle_rule: The underlying rules_apple test suite rule.
+        test_rule: The underlying rules_apple test rule.
+        test_factory: The factory object used to generate tests.
         apple_library: The macro used to package sources into a library.
         infoplists_by_build_setting: A dictionary of infoplists grouped by bazel build setting.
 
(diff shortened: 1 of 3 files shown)


rules/test.bzl
```diff
@@ -267,4 +298,18 @@ def ios_unit_snapshot_test(name, apple_library = apple_library, test_factory = d
         test_factory: Use this to generate other variations of tests.
         **kwargs: Arguments passed to the apple_library and ios_unit_test rules as appropriate.
     """
-    _ios_test(name, _ios_internal_unit_test_bundle, _ios_unit_test, test_factory, apple_library, internal_test_deps = ["@bazel_tools//tools/cpp/runfiles"], **kwargs)
```
luispadron, line 270: This is the bulk of the change, `@bazel_tools//tools/cpp/runfiles` was set for all snapshot test targets for some reason... this should be set by snapshot targets which need it not on in a global manner

luispadron:
Mode some changes to prepare the target for Bazel 7 along with adding a `bazel_version.is_bazel_7` constant similarly to how they determine this in rules_apple

jszumski (approved):

PR #845, opened 2024-03-06
Update: use `use_cpp_toolchain` for toolchains

luispadron:
This updates our `toolchains` usage to match rules_apple and use the `use_cpp_toolchains` helper.


thiagohmcruz (approved):

mattrobmattrob (approved):
This follows our supported Bazel versions guidelines, right?

luispadron:
> This follows our supported Bazel versions guidelines, right?

It should, rules_apple uses this directly and they support the same minimum Bazel version as us (6.0)

luispadron:
Confirmed `use_cpp_toolchain` was added in 5.1: https://blog.bazel.build/2022/03/24/bazel-5.1.html

PR #847, opened 2024-03-06
Update transition support

luispadron:
This completely updates the `transition_support.bzl` file. It now essentially is a copy of the one in here: https://github.com/bazelbuild/rules_apple/blob/f37305163dd7546f87839f3187e153c955cd20d8/apple/internal/transition_support.bzl

This is important as Bazel 7 has changed this a lot.

Changes:

- Delete old transition_support.bzl
- Add rules_apple transition_support.bzl
- Add rules_ios specific patches to transition_support.bzl (see the `PATCH:` comments)
- Update existing rules to support this new transition_support
- Remove unused transitions from transition_support.bzl that are used in rules_apple but not in rules_ios


rules/framework.bzl
```diff
@@ -1204,8 +1204,9 @@ that this target depends on.
             """,
         ),
         "_child_configuration_dummy": attr.label(
-            cfg = transition_support.split_transition,
-            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+            cfg = transition_support.apple_platform_split_transition,
+            providers = [cc_common.CcToolchainInfo, ApplePlatformInfo],
+            default = "@build_bazel_rules_apple//apple:default_cc_toolchain_forwarder",
```
luispadron, line 1209: After I updated `transition_support.bzl` to match rules_apple I was getting a Starlark crash. This change fixes it and matches rules_apple: https://github.com/bazelbuild/rules_apple/blob/f8fa4d9a32556ece0f7eee0ebdf997ca25bdc855/apple/internal/rule_attrs.bzl#L140-L149

rules/transition_support.bzl
```diff
@@ -32,15 +61,110 @@ def _current_apple_platform(apple_fragment, xcode_config):
...
+
+    if not platform_type:
+        fail("ERROR: platform_type not found")
+
+    return platform_type
+
+# PATCH: end
+
+# PATCH: start - private helper to get the `minimum_os_version` from the rules_ios attributes.
+def _rules_ios_minimum_os_version(platform_type, attr):
```
luispadron, line 90: This is essentially the custom stuff `rules_ios` needs, i.e. it uses `platforms` dict to determine this stuff. Longer term we should consider matching rules_apple here...

luispadron:
With this change + some toolchain updates i have in a follow up i got a target in our project to pass the build

rules/framework.bzl
luispadron: I'll follow up this PR with another to update toolchains, it'll fix toolchain resolution errors on Bazel 7

luispadron: Moving this to #848

mattrobmattrob (approved):

thiagohmcruz (approved):
LGTM, agree that ideally we should get rid of the custom `rules_ios` stuff and simply import/use what exists upstream. Should we file an issue for that if it doesn't exist already?

ps: edited your PR description to link to `rules_apple` in a specific SHA instead of `master` for future reference.

PR #850, opened 2024-03-11
Update rules for ObjcProvider deprecations

luispadron:
In Bazel 7+ `ObjcProvider` no longer supports/provides the required linking attributes.

The migrations is detailed here: https://github.com/bazelbuild/bazel/issues/16939. In summary, as part of the migration, the `ObjcProvider` fields which previously provided linking related information are now now longer providing that info. In addition to this, a new flag: `--incompatible_objc_linking_info_migration` was added to further delete these link attrs from the `ObjcProvider` making it an error if the attr is used or set.

The goal of this PR is to address support for `ObjcProvider` migration and to instead use the correct linking information from `CcInfo`. This will support both Bazel 6/7+. It does not try to support `--incompatible_objc_linking_info_migration` as that requires more changes and should be a separate PR

Depends on: 

- #848 
- #847

### DIFF
--- a/rules/force_load_direct_deps.bzl
+++ b/rules/force_load_direct_deps.bzl
@@ -1,36 +1,116 @@
+load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 load("//rules:providers.bzl", "AvoidDepsInfo")
 load("//rules:transition_support.bzl", "transition_support")
 
-def _impl(ctx):
+def _objc_provider_static_libraries(dep):
+    """Returns the ObjcProvider static libraries (.a) that should be force loaded.
+    """
+    if not apple_common.Objc in dep:
+        return []
+
+    return dep[apple_common.Objc].library.to_list()
+
+def _cc_info_static_libraries(dep):
+    """Returns the CcInfo static libraries (.a) that should be force loaded.
+
+    NOTE: CcInfo, unlike ObjcProvider, does not encode where the static library came from.
+    In the existing `_objc_provider_static_libraries` we only collect `.library` from ObjcProvider.
+    ObjcProvider `.library` list static library dependencies of the current target,
+    it does not include imported static libraries (such as those from `.framework` files).
+    CcInfo only provides `.static_library` and does not make this distinction.
+    To match this behavior, we only collect `.static_library` from CcInfo that are not from `.framework`s.
+    """
+    if not CcInfo in dep:
+        return []
+
+    static_cc_libraries = []
+    for linker_input in dep[CcInfo].linking_context.linker_inputs.to_list():
+        for library_to_link in linker_input.libraries:
+            if not library_to_link.static_library:
+                continue
+            containing_path = paths.dirname(library_to_link.static_library.path)
+            if containing_path.endswith(".framework"):
+                continue
+            static_cc_libraries.append(library_to_link.static_library)
+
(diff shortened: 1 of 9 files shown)


rules/force_load_direct_deps.bzl
```diff
@@ -1,36 +1,114 @@
...
+        return []
+
+    static_cc_libraries = []
+    for linker_input in dep[CcInfo].linking_context.linker_inputs.to_list():
+        for library_to_link in linker_input.libraries:
+            if not library_to_link.static_library:
+                continue
+            containing_path = paths.dirname(library_to_link.static_library.path)
+            if containing_path.endswith(".framework"):
+                continue
```
jszumski, line 32: Can you add a comment here or in the doc comment what/why we're attempting to exclude here?  Is having the `.framework` suffix a proxy for "an imported static library" that isn't a direct dep?

rules/import_middleman.bzl
```diff
@@ -260,23 +264,61 @@ def _file_collector_rule_impl(ctx):
         **objc_provider_fields
     )
 
-    additional_providers = []
+    # Create the deduplicated CcInfo provider information
+    static_libraries = objc_provider_fields.get("library", depset([])).to_list() + \
+                       objc_provider_fields.get("static_framework_file", depset([])).to_list() + \
+                       objc_provider_fields.get("imported_library", depset([])).to_list()
+    force_load_static_libraries = objc_provider_fields.get("force_load_library", depset([])).to_list()
+    dynamic_libraries = objc_provider_fields.get("dynamic_framework_file", depset([])).to_list()
```
jszumski, line 272: Could we merge these as `depset`s instead of dropping down to list?  This could change how duplicates are handled though

rules/force_load_direct_deps.bzl
```diff
@@ -1,36 +1,114 @@
...
+            static_cc_libraries.append(library_to_link.static_library)
 
-    force_load = []
+    return static_cc_libraries
 
+# TODO: We should deprecate this rule for Bazel 7+ as `--incompatible_objc_alwayslink_by_default` effectively
```
thiagohmcruz, line 37: Can we create an issue for this if it's not tracked somewhere else?

luispadron: https://github.com/bazel-ios/rules_ios/issues/862

rules/force_load_direct_deps.bzl
luispadron: Updated the doc comment to go into more details

rules/import_middleman.bzl
luispadron: We have to iterate on them below to create the `LibraryToLink` so don't think we can. I tried to make the bare minimum changes to this rule to get CI passing. 

I imagine were still dropping linking information / not deduping the inputs in Bazel 7+ because we the list is using information from `ObjcProvider`. But this rule's code is extremely confusing to me, so tried to keep it minimal. I'm hoping one of the Square folks can revisit this if it's required for other Bazel 7+ changes.

jszumski (approved):

rules/import_middleman.bzl
luispadron: I ended up reverting most of the changes here, it was regressing the build in Bazel 6 for Square folks. These current changes are enough to pass CI here in both 6 and 7 and fix that regression. I imagine another pass will be needed in this rule for Bazel 7 in other use cases though

jszumski (approved):

thiagohmcruz (approved):
Thx for improving the readability here as part of this 🙇

PR #852, opened 2024-03-25
Allow consumers to set always_include_developer_search_paths in swift_library

thiagohmcruz:
https://github.com/bazelbuild/rules_swift/pull/1162

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -959,10 +959,12 @@ def apple_library(
 
     if swift_sources:
         # To be backward compatible with rules_apple 2.x
-        swift_kwargs = dicts.add(kwargs, {"plugins": plugins} if plugins else {})
+        if plugins:
+            swift_kwargs = dicts.add(kwargs, {"plugins": plugins})
 
         # Allow consumers to opt-in to https://github.com/bazelbuild/rules_swift/pull/1162
-        swift_kwargs = dicts.add(swift_kwargs, {"always_include_developer_search_paths": swift_always_include_developer_search_paths} if swift_always_include_developer_search_paths else {})
+        if swift_always_include_developer_search_paths:
+            swift_kwargs = dicts.add(swift_kwargs, {"always_include_developer_search_paths": swift_always_include_developer_search_paths})
 
         swift_library(
             name = swift_libname,
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -958,9 +958,11 @@ def apple_library(
     )
 
     if swift_sources:
+        swift_kwargs = kwargs
+
         # To be backward compatible with rules_apple 2.x
         if plugins:
-            swift_kwargs = dicts.add(kwargs, {"plugins": plugins})
+            swift_kwargs = dicts.add(swift_kwargs, {"plugins": plugins})
 
         # Allow consumers to opt-in to https://github.com/bazelbuild/rules_swift/pull/1162
         if always_include_developer_search_paths:
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -579,6 +579,7 @@ def apple_library(
     testonly = kwargs.pop("testonly", False)
     features = kwargs.pop("features", [])
     plugins = kwargs.pop("plugins", None)
+    swift_always_include_developer_search_paths = kwargs.pop("swift_always_include_developer_search_paths", None)
 
     for (k, v) in {"momc_copts": momc_copts, "mapc_copts": mapc_copts, "ibtool_copts": ibtool_copts}.items():
         if v:
@@ -959,6 +960,9 @@ def apple_library(
     if swift_sources:
         # To be backward compatible with rules_apple 2.x
         swift_kwargs = dicts.add(kwargs, {"plugins": plugins} if plugins else {})
+
+        # Allow consumers to opt-in to https://github.com/bazelbuild/rules_swift/pull/1162
+        swift_kwargs = dicts.add(swift_kwargs, {"always_include_developer_search_paths": swift_always_include_developer_search_paths} if swift_always_include_developer_search_paths else {})
 
         swift_library(
             name = swift_libname,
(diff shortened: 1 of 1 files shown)


AngelaGuardia (approved):

rules/library.bzl
```diff
@@ -985,6 +985,10 @@ def apple_library(
             tags = tags_manual,
             defines = defines + swift_defines,
             testonly = testonly,
+            always_include_developer_search_paths = select({
```
mattrobmattrob, line 988: I suppose we only need the rules_swift bump because we're applying this directly vs. in the swift_kwargs below.

Is it easy to not require the rules_swift bump? I don't really feel strongly either way just wanted to ask.

mattrobmattrob (approved):

rules/library.bzl
thiagohmcruz: That's a good callout, aligned with feedback I got from @luispadron too. I guess one looking to use the flag can simply use kwargs indeed. 

For those looking to use the flag It would simplify things if `rules_ios` was on this `rules_swift` version though. So i guess I can repurpose this PR to simply bump `rules_swift` if folks don't have any concerns with that.

thiagohmcruz: @mattrobmattrob Q: What would be the issue with the `rules_swift` bump here specifically? I could make the changes in a way that for us everything works. But isn't this a good opportunity to add support for this flag for all consumers?

mattrobmattrob: Mostly just that this requires a rules_swift bump for consumers. Not necessarily a bad thing though.

thiagohmcruz:
Discussed offline and we agreed that instead of bumping `rules_swift` in `rules_ios` it provides a better DX for both consumers and maintainers to simply allow the flag `always_include_developer_search_paths`to be set in kwargs and propagated to the `swift_library`. Letting consumers decide if they want to bump to a newer `rules_swift` version or not to start setting that flag in their BUILD files.

luispadron (approved):
Thanks for iterating on this!

mattrobmattrob:
Shouldn't the flag be `always_include_developer_search_paths` instead of `swift_ always_include_developer_search_paths`?

rules/library.bzl
```diff
@@ -960,6 +961,9 @@ def apple_library(
         # To be backward compatible with rules_apple 2.x
         swift_kwargs = dicts.add(kwargs, {"plugins": plugins} if plugins else {})
 
+        # Allow consumers to opt-in to https://github.com/bazelbuild/rules_swift/pull/1162
+        swift_kwargs = dicts.add(swift_kwargs, {"always_include_developer_search_paths": swift_always_include_developer_search_paths} if swift_always_include_developer_search_paths else {})
```
mattrobmattrob, line 965: Want to do the if check first before mutating the `swift_kwargs` to minimize the penalty to all callers?

mattrobmattrob (approved):
I do wonder if we should come up with a general approach for passing rule attributes through to this point vs. a one off flag. Doesn't seem like we have a pattern for it at this point though.

thiagohmcruz:
> Shouldn't the flag be `always_include_developer_search_paths` instead of `swift_ always_include_developer_search_paths`?

@mattrobmattrob From a mixed frameworks perspective I think we should explicitly convey that this is a Swift thing and not Obj-C. Mostly thinking of a rules_ios user that is not aware of all the underlying impl details of the rules and want it to "just work". In that scenario the names used in the API are the main way to interact with the rule set and IMO we should be very explicit here. 

Open to change it though in case I am missing something here!

rules/library.bzl
thiagohmcruz: Makes sense, we should do the same in the `plugins` line above in that case

thiagohmcruz:
> I do wonder if we should come up with a general approach for passing rule attributes through to this point vs. a one off flag. Doesn't seem like we have a pattern for it at this point though.

@mattrobmattrob +1, I would still prefer to land this change to unblock ongoing work but I'm open to contribute to an issue/proposal on that. Def something I thought about in the past too while working on the rules.

mattrobmattrob:
> > Shouldn't the flag be `always_include_developer_search_paths` instead of `swift_ always_include_developer_search_paths`?
> 
> 
> 
> @mattrobmattrob From a mixed frameworks perspective I think we should explicitly convey that this is a Swift thing and not Obj-C. Mostly thinking of a rules_ios user that is not aware of all the underlying impl details of the rules and want it to "just work". In that scenario the names used in the API are the main way to interact with the rule set and IMO we should be very explicit here. 
> 
> 
> 
> Open to change it though in case I am missing something here!


Deleted my comment but wasn't fast enough. I thought you were passing it through to the swift_library with the prefix but realized I was wrong.

rules/library.bzl
```diff
@@ -579,6 +579,7 @@ def apple_library(
     testonly = kwargs.pop("testonly", False)
     features = kwargs.pop("features", [])
     plugins = kwargs.pop("plugins", None)
+    swift_always_include_developer_search_paths = kwargs.pop("swift_always_include_developer_search_paths", None)
```
luispadron, line 582: I agree this should be 
```suggestion
    swift_always_include_developer_search_paths = kwargs.pop("always_include_developer_search_paths", None)
```

we didnt make that distinction for `plugins`

rules/library.bzl
thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/852/commits/71de457039f733c1b871f15e3d8f5842939647c1

rules/library.bzl
thiagohmcruz: See [this](https://github.com/bazel-ios/rules_ios/pull/852#issuecomment-2018890660)

luispadron: Essentially a user would have to look at the code here to understand they should use `swift_always_include_developer_search_paths` vs. the actual flag `always_include_developer_search_paths`

thiagohmcruz: Ok, people will try to use flags by the original names. I'm convinced, cc @mattrobmattrob haha

thiagohmcruz: https://github.com/bazel-ios/rules_ios/pull/852/commits/4a57fd23f7b82dfb8d26c33599dacc866a6f6c0e

PR #853, opened 2024-03-26
Forward package_name to swift_library part of apple_library

luispadron:
Adds support for `package_name` arg of `swift_library`. I also updated the kwarg forwarding to match the way we do it in `framework.bzl` and the other macros.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -27,6 +27,12 @@ GLOBAL_INDEX_STORE_PATH = "bazel-out/_global_index_store"
 
 _MANUAL = ["manual"]
 
+_SWIFT_LIBRARY_KWARGS = [
+    "always_include_developer_search_paths",
+    "package_name",
+    "plugins",
+]
+
 def _private_headers_impl(ctx):
     return [
         PrivateHeadersInfo(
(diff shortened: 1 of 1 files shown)


jszumski (approved):

rules/library.bzl
```diff
@@ -958,15 +962,12 @@ def apple_library(
...
-            swift_kwargs = dicts.add(swift_kwargs, {"plugins": plugins})
-
-        # Allow consumers to opt-in to https://github.com/bazelbuild/rules_swift/pull/1162
-        if always_include_developer_search_paths:
-            swift_kwargs = dicts.add(swift_kwargs, {"always_include_developer_search_paths": always_include_developer_search_paths})
+        # Collect the swift_library related kwargs, these are typically only set when provided to allow
+        # for wider compatibility with rule_swift versions.
+        swift_kwargs = dicts.add(
+            kwargs,
+            {arg: kwargs.pop(arg) for arg in _SWIFT_LIBRARY_KWARGS if arg in kwargs},
```
thiagohmcruz, line 969: Have you verified if pop-ing this only here instead of up around line 581 causes problems? IIUC if we don't pop these early it means we'll pass these flags to the [modulemap_generator](https://github.com/bazel-ios/rules_ios/blob/eab038c4028156f7b3825ae9bf2a28675ac8ccb8/rules/library.bzl#L836) correct?

thiagohmcruz (approved):

rules/library.bzl
luispadron: Good point, updated this to collect/pop the swift related kwargs earlier and then add them to kwargs right before the library is created.

PR #858, opened 2024-04-01
Do not declare force_load and VFS targets unnecessarily

thiagohmcruz:
Saves analysis time since it implies less rules impls to go through and less configured targets to process

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -537,6 +537,7 @@ def apple_library(
     namespace = module_name if namespace_is_module_name else name
     module_map = kwargs.pop("module_map", None)
     swift_objc_bridging_header = kwargs.pop("swift_objc_bridging_header", None)
+    has_swift_sources = len(swift_sources) > 0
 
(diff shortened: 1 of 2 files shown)


rules/library.bzl
```diff
@@ -615,6 +616,9 @@ def apple_library(
             )
 
             swift_sources.append(intent_name)
+
+            # Refresh `has_swift_sources` since `swift_sources` has changed
+            has_swift_sources = len(swift_sources) > 0
```
luispadron, line 621: I dont think this could ever change right? Since to be in this `if` we must have had swift_sources and so adding a swift source file wont ever make this false.

luispadron (approved):
This is great thank you!

luispadron (approved):

rules/library.bzl
thiagohmcruz: Oh good catch, I'll remove it

thiagohmcruz: Force pushed with the change

PR #861, opened 2024-04-02
CI: improvements for stability

luispadron:


mattrobmattrob (approved):
https://github.com/bazel-ios/rules_ios/pull/677 plus utilizing the pre-boot sim will likely help as well but 🤷

rules/test/lldb/lldb_sim_runner.py
```diff
@@ -174,26 +171,11 @@ class TestContext():
...
-        else:
-            xcode_select_result = subprocess.run(["xcode-select", "-p"],
-                                                 encoding="utf-8",
-                                                 check=True,
-                                                 stdout=subprocess.PIPE)
-
-            self.developer_dir = xcode_select_result.stdout.rstrip()
-        os.environ["DEVELOPER_DIR"] = self.developer_dir
-        self.simctl_path = os.path.join(
-            self.developer_dir, "usr", "bin", "simctl")
```
luispadron, line 195: This setting of the environ was causing issues in testing, seemed to be changing the xcode selected path. Updated this to use `xcrun -f` to find the path instead and it seems to work more stably

jszumski (approved):

PR #864, opened 2024-04-05
Update and fix for rules_apple 3.5.0

luispadron:
rules_apple 3.5.0 broke `resolved_*` tools from `AppleMacToolchainInfo` (https://github.com/bazelbuild/rules_apple/pull/2438) This sets the new minimum supported version for the next rules_ios release to 3.5.0.

PR #865, opened 2024-04-08
Removed fake_rule_label from precompiled_apple_resource_bundle_impl and using swift_module instead

tymurmustafaiev:
**Why this is needed?**
`fake_rule_label` was as a workaround and I'd like to remove it and use swift_module.
This pr is a continuation of https://github.com/bazelbuild/rules_apple/pull/2335.
If we try to build 2 different targets with same module name - plist generation will fail due to rule name collision as `fake_rule_label` is used, which was created from `swift_module` or `bundle_name`

**What has changed?**
Using `swift_module` for compiling assets, plists, storyboards, etc. instead of `fake_rule_label`. Now, each plist generation action has it's own rule name which has no name collision

**NOTE**
This PR Requires `rules_apple` to be at least 3.2.0, so I'll fix tests once it's updated

### DIFF
--- a/rules/precompiled_apple_resource_bundle.bzl
+++ b/rules/precompiled_apple_resource_bundle.bzl
@@ -34,19 +34,6 @@ def _precompiled_apple_resource_bundle_impl(ctx):
     current_apple_platform = transition_support.current_apple_platform(apple_fragment = ctx.fragments.apple, xcode_config = ctx.attr._xcode_config)
     platform_type = str(current_apple_platform.platform.platform_type)
 
-    if apple_api_version == "3.0":
-        platform_prerequisites_version_args = {
-            "build_settings": None,
-        }
-        rules_api_3_resource_partials_args = {
-            "include_executable_name": False,  # Must be set to False or bundle_name is now used if executable_name is None
-        }
-    else:
-        platform_prerequisites_version_args = {
-            "disabled_features": ctx.disabled_features,
-        }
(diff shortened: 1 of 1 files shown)


luispadron:
Thanks, this is a nice cleanup! If this requires a new rules_apple version you'll have to rebase after I merge #864

tymurmustafaiev:
> Thanks, this is a nice cleanup! If this requires a new rules_apple version you'll have to rebase after I merge #864

Sure, I'll keep an eye on it and rebase/fix unit tests once it's ready 🙏

tymurmustafaiev:
@luispadron how can I trigger CI checks? 🙏 It seems waiting for status for a long time

luispadron (approved):

rules/precompiled_apple_resource_bundle.bzl
```diff
@@ -107,6 +91,8 @@ def _precompiled_apple_resource_bundle_impl(ctx):
         paths.join("%s-intermediates" % ctx.label.name, "Info.plist"),
     )
 
+    # No need to pass swift_module to merge_root_infoplists
+    partials_args.pop("swift_module")
```
luispadron, line 95: What happens if we do pass this?

tymurmustafaiev: @luispadron `merge_root_infoplists` fails because swift_module is unexpected param. I could make new dictionary but not sure if it's better solution

PR #873, opened 2024-06-05
Fix import_middleman in Bazel 7

thiagohmcruz:

PR #877, opened 2024-06-14
Bazel 7: Set framework_includes in import_middleman

thiagohmcruz:
Continuation of: https://github.com/bazel-ios/rules_ios/pull/873

While testing Bazel 7 hit some build failures due to missing `-F` flags in the invocation.

I could set other attributes here too but since I'm not super familiar with this rule I'd rather add what is missing progressively as I run tests in Bazel 7.

### DIFF
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -269,6 +269,15 @@ def _file_collector_rule_impl(ctx):
     cc_info = None
     if is_bazel_7:
         cc_info = CcInfo(
+            compilation_context = cc_common.create_compilation_context(
+                framework_includes = depset(
+                    transitive = [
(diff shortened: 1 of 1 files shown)


luispadron (approved):

mattrobmattrob (approved):

mattrobmattrob:
Is it easy to add a test that demonstrates this failure on th Bazel 7.X test jobs?

thiagohmcruz:
> Is it easy to add a test that demonstrates this failure on th Bazel 7.X test jobs?

@mattrobmattrob not sure but I'll take all look

thiagohmcruz:
@mattrobmattrob just FYI I'll add a test to this in a follow up. Couldn't find an easy way to repro what I see inside yet.

PR #878, opened 2024-06-18
Fix extension_safe attr not passing copts

luispadron:
When using `extension_safe` the `copt` for the objc/swift library must also supply `-fapplication-extension`/`-application-extension`

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -589,7 +589,9 @@ def apple_library(
 
     # Set extra linkopt for application extension safety
     if extension_safe:
-        linkopts.append("-fapplication-extension")
+        linkopts += ["-fapplication-extension"]
+        objc_copts += ["-fapplication-extension"]
+        swift_copts += ["-application-extension"]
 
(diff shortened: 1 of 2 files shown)


rules/library.bzl
```diff
@@ -589,7 +589,9 @@ def apple_library(
 
     # Set extra linkopt for application extension safety
     if extension_safe:
-        linkopts.append("-fapplication-extension")
+        linkopts += ["-fapplication-extension"]
```
luispadron, line 592: Use `+=` here instead of append so that `select` is supported.

thiagohmcruz (approved):

jszumski (approved):

luispadron:
Updated one of the examples, without this change it fails with:

```
tests/ios/frameworks/dynamic/b/lib.swift:11:15: error: 'accessoryCircular' is only available in iOS 16.0 or newer
            [.accessoryCircular]
              ^
```

so should catch regressions in the future too

PR #879, opened 2024-06-27
Update for rules_swift 2.x

luispadron:
This fixes a missing `_swift_vfs.yaml` error related to the changes in: https://github.com/bazelbuild/rules_swift/pull/1191

Fixes #875

PR #880, opened 2024-07-05
Move resource generation below framework vfs

luispadron:
This moves resource generation (and being added to `deps`) below the `framework_vfs` creation which uses the `deps`

With the change in #879 to add the vfs to the `swiftc_inputs`  instead of the `deps` of the `swift_library` the `framework_vfs` was now also forwarding the resources (because it gets added to `deps` after generation) to other rules (like `rules_xcodeproj`) which collect extra files from this the `swiftc_inputs` attr. This led to duplicated resources in the project as they're collected from `resources` and now the transitive `deps` of the `framework_vfs`

AFAIK resources aren't needed for the `framework_vfs` rules and so this change should resolve the new `rules_xcodeproj` incompatibility while making the `deps` to the `framework_vfs` more scoped.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -819,14 +819,6 @@ def apple_library(
     if vendored_deps_arm64_sim:
         deps += vendored_deps_arm64_sim
 
-    resource_bundles = library_tools["resource_bundle_generator"](
-        name = name,
-        library_tools = library_tools,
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):
At first glance I don't see any issues with the change. I'll test it in our project and report issues if I find any.

PR #881, opened 2024-07-05
Remove redundant modulemap in `swiftc_inputs`

luispadron:
In #879 we added the extended modulemap to the `swiftc_inputs` below this line. This PR removes the non extended modulemap from being added as an input as we only need the extended version.

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -951,8 +951,6 @@ def apple_library(
             })
 
         swiftc_inputs = other_inputs + objc_hdrs + objc_private_hdrs
-        if module_map:
-            swiftc_inputs.append(module_map)
         if swift_objc_bridging_header:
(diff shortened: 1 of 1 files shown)


rules/library.bzl
```diff
@@ -951,8 +951,6 @@ def apple_library(
             })
 
         swiftc_inputs = other_inputs + objc_hdrs + objc_private_hdrs
-        if module_map:
-            swiftc_inputs.append(module_map)
```
luispadron, line 955: Already added here: https://github.com/bazel-ios/rules_ios/pull/881/files#diff-23068963449a2bf4d162d813a4dda51236b9d044d1442e8a4bb4ad7680473478R973

mattrobmattrob (approved):

thiagohmcruz (approved):

PR #893, opened 2024-08-12
Revert "Add headers to `headermap` action `inputs`"

luispadron:
Reverts bazel-ios/rules_ios#888

### DIFF
--- a/rules/hmap.bzl
+++ b/rules/hmap.bzl
@@ -32,16 +32,9 @@ def _make_hmap(actions, headermap_builder, output, namespace, hdrs_lists):
     args.set_param_file_format(format = "multiline")
     args.use_param_file("@%s")
 
-    headers = [
-        header
-        for header_list in hdrs_lists
-        for header in header_list
-    ]
-
     actions.run(
         mnemonic = "HmapCreate",
         arguments = [args],
-        inputs = headers,
         executable = headermap_builder,
         outputs = [output],
     )


thiagohmcruz (approved):

mattrobmattrob:
Why the revert of this one?

justinseanmartin (approved):
This seems right to me. After discussing with @congt - we don't think the hmaptool code requires the headers themselves to generate the hmap, just the paths to those files, therefore we don't need to declare them as inputs.

We could potentially make this a bit more explicit by having the input to `_make_hmap` from `_make_headermap_impl` be an array of strings instead of an array of files, but that isn't entirely necessary. It would make things more consistent with what we're passing internally to the `make_hmap` method and what the method generally requires to invoke the hmap generator (and we rely on being able to pass strings in the case of VFS overlays). That would technically be a breaking change though, because `make_hmap` is exposed publicly.

luispadron:
Right I was under the impression the hmap tool would need to access the headers themelves but effectively what it does is create a file that looks like:

```
Foo/MyHeader.h -> MyHeader.h
```

luispadron:
This API should 100% be `string_list` instead of `label_list`: https://github.com/bazel-ios/rules_ios/pull/893/files#diff-c7a846a20e1dd3e0a809cc965ea0d4134555adb5bc82fcc298bfb6a933f5e3c0R110

PR #894, opened 2024-08-16
Fix apple_library rule not working in sandbox when do `SwiftCompile` in mixed sources

gyfelton:
To repro, run `bazel clean; bazel build //tests/ios/frameworks/mixed-source/only-source:SwiftLibrary -s --sandbox_debug --strategy=SwiftCompile=sandboxed` and got 
`<unknown>:0: error: underlying Objective-C module 'SwiftLibrary' not found` error

using`bazel aquery //tests/ios/frameworks/mixed-source/only-source:SwiftLibrary_swift` to look at input list, realized `bazel-out/darwin_arm64-dbg/bin/tests/ios/frameworks/mixed-source/only-source/SwiftLibrary-modulemap/SwiftLibrary.modulemap`  entry is not there even though that's sth generated by the module_map generator

So adding this fixed the issue because sandbox can now pick up this file 

Next step: setup CI job here that run everything in sandbox mode, try to fix anything that is broken in another PR

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -931,6 +931,7 @@ def apple_library(
 
     if has_swift_sources:
         additional_swift_copts += ["-Xcc", "-I."]
+        swiftc_inputs = other_inputs + objc_hdrs + objc_private_hdrs
         if module_map:
             # Frameworks find the modulemap file via the framework vfs overlay
             if not namespace_is_module_name:
@@ -940,8 +941,8 @@ def apple_library(
                 "@build_bazel_rules_ios//:swift_disable_import_underlying_module": [],
                 "//conditions:default": ["-import-underlying-module"] if not feature_names.swift_disable_import_underlying_module in features else [],
             })
+            swiftc_inputs += [module_map]
 
-        swiftc_inputs = other_inputs + objc_hdrs + objc_private_hdrs
         if swift_objc_bridging_header:
             if swift_objc_bridging_header not in objc_hdrs:
                 swiftc_inputs.append(swift_objc_bridging_header)
@@ -952,6 +953,8 @@ def apple_library(
         generated_swift_header_name = module_name + "-Swift.h"
 
         if module_map:
+            # TODO: now that we always add module_map as a swiftc_input,
+            # we should consider removing this one if it's not needed
             extend_modulemap(
                 name = module_map + ".extended." + name,
                 destination = "%s.extended.modulemap" % name,


rules/library.bzl
```diff
@@ -940,8 +941,8 @@ def apple_library(
                 "@build_bazel_rules_ios//:swift_disable_import_underlying_module": [],
                 "//conditions:default": ["-import-underlying-module"] if not feature_names.swift_disable_import_underlying_module in features else [],
             })
+            swiftc_inputs += [module_map]
```
luispadron, line 944: Does this already get added for the extended modulemap below? Or is that overridden 👀

congt: This `module_map` is passed to framework_vfs_overlay_name_swift, while the one in extended modulemap below is different and passed to framework_vfs_overlay_name and objc_libname. I think it's unnecessary to add the extended modulemap to swiftc_inputs.

(BTW, the code is really hard to reason. We need to clean it up)

congt: @gyfelton Could you test deleting line 965 to see if it still works as expected?

congt (approved):

rules/library.bzl
luispadron: Yeah agreed on it being hard to reason about. I'm happy to merge whatever has the least amount of changes, thinking about it more im not sure extended module map should be an input so let's try deleting it. 

It might be best to create a sandbox test ci job and merge these together

gyfelton: So the extended modulemap has two major diff:
1. location: the added module map has umbrella header next to it, but the extended one is at root
Screenshot 2024-08-19 at 9 43 48 AM
2. content:
extended one has the following:
```
framework module SwiftLibrary {
    umbrella header "SwiftLibrary-umbrella.h"

    export *
    module * { export * }
}

module SwiftLibrary.Swift {
    header "SwiftLibrary-Swift.h"
    requires objc
}
```
The SwiftLibrary.Swift module is the net addition to the original module map

deleting the adding of extended module map works for my case but i am not sure about the implication of removing it.
I will add a TODO here for now and if we all agree on the future plan, i can come back to this esp. during the time when i need to have rules_ios to fully pass CI in sandbox.  WDYT?

gyfelton: strike that, that is needed or we got  `-swift.h` file not found error in our repo's compilation

congt: > that is needed or we got -swift.h file not found error in our repo's compilation

Yes, we still need the extended modulemap for framework_vfs_overlay_name and objc_libname, but I don't think we need to add it to swiftc_inputs on line 967 because the swift compilation shouldn't need the umbrella header or -Swift.h.

rules/library.bzl
```diff
@@ -952,6 +953,8 @@ def apple_library(
         generated_swift_header_name = module_name + "-Swift.h"
 
         if module_map:
+            # TODO: now that we always add module_map as a swiftc_input,
```
congt, line 956: We should move the TODO to line 967.

gyfelton: yup removing in my next PR trying to get CI sandbox run

PR #895, opened 2024-08-19
Add sandbox mode to CI testing matrix

gyfelton:
What changed:
1. Add sandbox on/off to testing matrix, so we have 2 bazel version * VFS on/off * sandbox on off = 8 tests in total ATM
2. Fix redefinition of `AppErrorCode` enum error in sandbox mode by removing the extra import in bridging header. The tests still works and can catch renaming of the enum for example. Why it was working in non sandbox mode is unknown
3. Add `sandbox` config to include all settings needed to pass CI. specifically allow a small set of actions to run locally.

Why this change:
To add CI coverage to sandbox mode to ensure no regression on it, now that we have fixed issues with sandbox mode in https://github.com/bazel-ios/rules_ios/pull/894

Discussion:
CI duration might be a concern, we can exclude certain combo in the test matrix if we want, it's possible with `exclude` keyword

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,7 +2,7 @@
 common --enable_bzlmod
 
 # Spawn strategy required for some tests
-build --spawn_strategy=standalone
+build --spawn_strategy=local
 
 # Setup Xcode configuration.
 build --xcode_version_config=//:host_xcodes
@@ -61,6 +61,16 @@ build:remote_cache --nolegacy_important_outputs
 build:remote_cache --noslim_profile
 build:remote_cache --remote_cache=grpcs://remote.buildbuddy.io
 build:remote_cache --remote_timeout=3600
+
+build:sandboxed --spawn_strategy=sandboxed
+# Otherwise it still use the worker strategy
+build:sandboxed --strategy=SwiftCompile=sandboxed
+build:sandboxed --strategy=Genrule=sandboxed,local
+build:sandboxed --strategy=StoryboardCompile=sandboxed,local
+build:sandboxed --strategy=CleaningFramework=sandboxed,local
+build:sandboxed --strategy=StoryboardLink=sandboxed,local
+build:sandboxed --strategy=ProcessAndSign=sandboxed,local
+build:sandboxed --strategy=AssetCatalogCompile=sandboxed,local
 
 # By default don't upload local results to remote cache, only CI does this.
 build --noremote_upload_local_results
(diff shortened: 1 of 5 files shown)


.bazelrc
```diff
@@ -62,6 +62,14 @@ build:remote_cache --noslim_profile
 build:remote_cache --remote_cache=grpcs://remote.buildbuddy.io
 build:remote_cache --remote_timeout=3600
 
+build:sandboxed --strategy=SwiftCompile=sandboxed
```
luispadron, line 65: Is it possible to test sandboxed for all actions here and only allow certain ones to fallback to local instead? This might give us more info

gyfelton: that's what we are doing here no? default setting has swiftCompile using worker and everything else local
This new config is everything sandboxed with a select set being allowed to fallback to local

.bazelrc
```diff
@@ -62,6 +62,14 @@ build:remote_cache --noslim_profile
...
+build:sandboxed --spawn_strategy=sandboxed
+build:sandboxed --strategy=Genrule=sandboxed,local
+build:sandboxed --strategy=StoryboardCompile=sandboxed,local
+build:sandboxed --strategy=CleaningFramework=sandboxed,local
+build:sandboxed --strategy=StoryboardLink=sandboxed,local
+build:sandboxed --strategy=ProcessAndSign=sandboxed,local
```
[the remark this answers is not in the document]
gyfelton: without them CI fails with `No usable spawn strategy found for spawn with mnemonic XXX. Are your --spawn_strategy or --strategy flags too strict?` error.
This list might grow in future, but I don't think they need to be run in sandbox anyway.

.bazelrc
luispadron: I meant, could we do `--strategy=sandboxed`

.bazelrc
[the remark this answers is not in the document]
luispadron: Gotcha, that makes sense

.bazelrc
luispadron: hmm i guess you need it to override the above like:

```
build --strategy=SwiftCompile=worker
```

nvm this all looks good, thank you!

luispadron:
Updated the required check in GitHub this should be ready once CI completes

thiagohmcruz (approved):

PR #900, opened 2024-08-21
Remove extended modulemap from swiftc_inputs

luispadron:
The normal modulemap was added in https://github.com/bazel-ios/rules_ios/pull/894 and the extended modulemap is only for the objective-c part of the library so we shouldn't add it for the swiftc_inputs


congt (approved):

gyfelton:
i tested removing the creation of this extended.modulemap entirely and it failed the build... but yeah i should have tested this version where i still add it to the VFS overlay side

PR #901, opened 2024-08-25
End-to-end support of SDK clang and swift modules

congt:
### What

* Reimplemented sdk_clang_module to support interdependency among sdk_clang_modules
* Adds rule sdk_swiftmodule_import to support SDK swift modules
* Updates xcode_sdk_frameworks to allow a swift_binary to build with explicit SDK modules. Also updates the explicit_modules example to build with SDK modules.

### Test

`bazel run --config=explicit_modules example/explicit_module:hello_world`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,7 +5,7 @@ common --enable_bzlmod
 build --spawn_strategy=standalone
 
 # Setup Xcode configuration.
-build --xcode_version_config=//:host_xcodes
+build --xcode_version=15.4.0.15F31d
 
 build --experimental_strict_conflict_checks
 
@@ -65,6 +65,7 @@ build:remote_cache --remote_timeout=3600
 # Configure for explicit module compilation
 build:explicit_modules --features=swift.use_c_modules
 build:explicit_modules --features=swift.emit_c_module
+build:explicit_modules --features=swift.use_explicit_swift_module_map
 build:explicit_modules --repo_env=EXPLICIT_MODULES=1
 
 # By default don't upload local results to remote cache, only CI does this.
--- a/example/explicit_module/BUILD.bazel
+++ b/example/explicit_module/BUILD.bazel
@@ -1,23 +1,12 @@
-load("@build_bazel_rules_swift//swift:swift.bzl", "swift_binary")
-load("//rules/explicit_module:sdk_clang_module.bzl", "sdk_clang_module")
-
-sdk_clang_module(
-    name = "swift_concurrency_shims",
-    module_map = "__BAZEL_XCODE_SDKROOT__/usr/lib/swift/shims/module.modulemap",
-    module_name = "_SwiftConcurrencyShims",
-)
-
-sdk_clang_module(
-    name = "shims",
-    module_map = "__BAZEL_XCODE_SDKROOT__/usr/lib/swift/shims/module.modulemap",
-    module_name = "SwiftShims",
+load(
+    "@build_bazel_rules_swift//swift:swift.bzl",
+    "swift_binary",
 )
 
 swift_binary(
     name = "hello_world",
     srcs = ["main.swift"],
     deps = [
-        ":shims",
-        ":swift_concurrency_shims",
+        "@xcode_sdk_frameworks//version15_4_0_15F31d/MacOSX:Foundation_swift",
     ],
 )
--- a/example/explicit_module/main.swift
+++ b/example/explicit_module/main.swift
@@ -1,1 +1,3 @@
+import Foundation
+
 print("Hello world")
--- a/rules/explicit_module/sdk_clang_module.bzl
+++ b/rules/explicit_module/sdk_clang_module.bzl
@@ -1,47 +1,103 @@
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load("@build_bazel_rules_swift//swift/internal:compiling.bzl", "precompile_clang_module")
 load("@build_bazel_rules_swift//swift/internal:feature_names.bzl", "SWIFT_FEATURE_SYSTEM_MODULE")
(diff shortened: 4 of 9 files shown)


.bazelrc
```diff
@@ -5,7 +5,7 @@ common --enable_bzlmod
 build --spawn_strategy=standalone
 
 # Setup Xcode configuration.
-build --xcode_version_config=//:host_xcodes
+build --xcode_version=15.4.0.15F31d
```
gyfelton, line 8: maybe override this just for `build:explicit_modules`?

example/explicit_module/BUILD.bazel
```diff
@@ -1,23 +1,12 @@
...
+    "swift_binary",
 )
 
 swift_binary(
     name = "hello_world",
     srcs = ["main.swift"],
     deps = [
-        ":shims",
-        ":swift_concurrency_shims",
+        "@xcode_sdk_frameworks//version15_4_0_15F31d/MacOSX:Foundation_swift",
```
gyfelton, line 10: possible to have this as a separate var that can be specified next to `build:explicit_module --xcode_version=xxx`? this way we at least control the xcode version EMB points to easily

example/explicit_module/main.swift
```diff
@@ -1 +1,3 @@
+import Foundation
```
gyfelton, line 1: 🎉

rules/explicit_module/sdk_clang_module.bzl
```diff
@@ -1,47 +1,103 @@
...
     return [
-        swift_common.create_swift_interop_info(
-            module_map = ctx.attr.module_map,
-            module_name = ctx.attr.module_name,
-            requested_features = [SWIFT_FEATURE_SYSTEM_MODULE],
+        create_swift_info(
+            modules = [
+                create_module(
+                    name = ctx.attr.module_name,
+                    clang = create_clang_module(
```
gyfelton, line 57: why not just surface them up to be on the same level as precompiled_module at line 41? that way it's clear to see that the chain is: `precompile_clang_module -> create_clang_module -> create_module -> create_swift_info`

rules/explicit_module/sdk_clang_module.bzl
```diff
@@ -1,47 +1,103 @@
...
-        ),
-        "module_name": attr.string(
-            doc = """\
-The name of the top-level module in the module map that this target represents.
-""",
-            mandatory = True,
-        ),
-    },
+    implementation = _sdk_clang_module_impl,
+    fragments = ["cpp"],
```
gyfelton, line 73: having this is to access what's listed here? https://bazel.build/rules/lib/fragments/cpp

rules/explicit_module/sdk_clang_module.bzl
```diff
@@ -1,47 +1,103 @@
...
         ),
-        # We need to return CcInfo and its compilation_context. We may also consider to update swift_clang_module_aspect.
-        # See https://github.com/bazelbuild/rules_swift/blob/d68b21471e4e9d922b75e2b0621082b8ce017d11/swift/internal/swift_clang_module_aspect.bzl#L548
-        CcInfo(compilation_context = cc_common.create_compilation_context()),
-        # Required to add sdk_clang_module targets to the deps of swift_module_alias.
-        # TODO(cshi): create the SwiftInfo correctly
-        create_swift_info(),
     ]
 
 sdk_clang_module = rule(
```
gyfelton, line 71: i notice that the refence to sdk_clang_module is removed in the example... that is because those are code gen by the repo rules right?

rules/explicit_module/sdk_swiftmodule_import.bzl
```diff
@@ -0,0 +1,38 @@
...
+    )
+    return [DefaultInfo(files = depset([output_swiftmodule]))]
+
+sdk_swiftmodule_import = rule(
+    implementation = _sdk_swiftmodule_import_impl,
+    attrs = {
+        "swiftmodule_path": attr.string(
+            doc = """\
+The absolute path to the SDK swiftmodule, e.g.,
+/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphoneos/prebuilt-modules/17.5/UIKit.swiftmodule.
```
gyfelton, line 26: this does not do substitution? referring to __BAZEL_XCODE_SDKROOT__  for the module_map attr

rules/xcode_sdk_frameworks.bzl
```diff
@@ -231,13 +244,15 @@ def _create_build_file_for_sdk(
     for i in range(module_count):
         name_idx = 2 * i
         details_idx = 2 * i + 1
-        targets.append(_target_for_module(
+        target = _target_for_module(
```
gyfelton, line 247: `_target_for_module` might return `None`? is that even allowed?  or this rules is still in construction since some code is commented out

gyfelton (approved):
just a bunch of questions and no logic altering suggestions

.bazelrc
gyfelton: also i still cannot build as is with xcode 16, i wonder if we just add this to line 418 of `rules/xcode_sdk_frameworks.bzl` to indicate that we for now just support xcode 15.4
```
        if not xcode_version_name.startswith("version15_4"):
            continue
```

rules/explicit_module/sdk_clang_module.bzl
congt: Yeah, it's required to build the explicit pcm

rules/explicit_module/sdk_swiftmodule_import.bzl
congt: run_shell doesn't do that. We will probably need to use apple_common

PR #910, opened 2024-08-30
Fix import_middleman broken with Bazel 7 + sandbox mode

gyfelton:
What changed and why:
1. added test case that broke with main, under sandbox mode and bazel 7 and `arm64_simulator_use_device_deps` feature turned on :
To repro locally, run `bazel build //tests/ios/unit-test/test-imports-app:TestImports-App --config=ios --features apple.arm64_simulator_use_device_deps` and it fails. But success if change `.bazelversion` to 6.4.0
The error is "unable to find header `basic.h`" which is the same issue with what our own repo has. Also this only break Objc side not swift side (probably because CcInfo is more used by objc_library?)

2. To fix above: use the compilation_context generated originally.
The original fix https://github.com/bazel-ios/rules_ios/pull/873 is missing fields inside `compilation_context` such as `headers`. So might as well use the original CcInfo collected, and only recreate the linking context.
BTW i believe the original PR aims to fix this kind of error in bazel 7:
```
ld: building for 'iOS-simulator', but linking in object file (/path/to/someframework.framework[arm64][2] built for 'iOS'
```
Which is the error we got if trying to just use the original CcInfo.
3. Update the test matrix to have sandbox mode for the tests for `arm64_simulator_use_device_deps` feature
 
Tests done:
Without the change from https://github.com/bazel-ios/rules_ios/pull/903 some checks should still fail but the ones using `arm64_simulator_use_device_deps` should be green

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,7 @@ jobs:
           path: bazel-testlogs
 
   build_arm64_simulator:
-    name: arm64 Simulator (Bazel ${{ matrix.bazel_version }} / Xcode ${{ matrix.xcode_version }})
+    name: arm64 Simulator (Bazel ${{ matrix.bazel_version }} / Xcode ${{ matrix.xcode_version }}) / Sandbox ${{ matrix.sandbox }}
     runs-on: macos-14
     strategy:
       fail-fast: false
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,7 @@ jobs:
           path: bazel-testlogs
 
   build_arm64_simulator:
-    name: arm64 Simulator (Bazel ${{ matrix.bazel_version }} / Xcode ${{ matrix.xcode_version }}) / Sandbox ${{ matrix.sandbox }}
+    name: arm64 Simulator (Bazel ${{ matrix.bazel_version }} / Xcode ${{ matrix.xcode_version }} / Sandbox ${{ matrix.sandbox }})
     runs-on: macos-14
     strategy:
       fail-fast: false
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,6 +86,7 @@ jobs:
       fail-fast: false
       matrix:
         bazel_version: [6.5.0, 7.1.0]
+        sandbox: [true, false]
         xcode_version: [15.2]
     env:
         XCODE_VERSION: ${{ matrix.xcode_version }}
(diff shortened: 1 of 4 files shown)


tests/ios/unit-test/test-imports-app/empty.swift
```diff
@@ -1,9 +1,12 @@
 import Foundation
 import SomeFramework
+import Basic
```
gyfelton, line 3: this still work with bazel 7 probably because swift compile does not depend on CcInfo

rules/import_middleman.bzl
```diff
@@ -297,10 +291,6 @@ def _file_collector_rule_impl(ctx):
                 ]),
             ),
         )
-    else:
-        dep_cc_infos = [dep[CcInfo] for dep in ctx.attr.deps if CcInfo in dep]
-        cc_info = cc_common.merge_cc_infos(cc_infos = dep_cc_infos)
```
luispadron, line 302: I think we should add back the else case here so that we dont `merge_cc_infos` and waste time in the Bazel 7 case

luispadron (approved):

thiagohmcruz (approved):

rules/import_middleman.bzl
thiagohmcruz: You mean Bazel 6 right? Unless I misunderstood something :)

luispadron: merging is the Bazel 6 case, without the if/else we always merge and then throw away the result in Bazel 7

gyfelton: I started with original` if else` clause by having more fields in the create_compilation_context, but then it kept failing on other places. 
the impl behind the merge_cc_info is also merging of the fields of each one, isn't that the same as the the new impl?

luispadron: ah i see `cc_info` is also returned below so this makes sense

.github/workflows/tests.yml
```diff
@@ -86,6 +86,7 @@ jobs:
       fail-fast: false
       matrix:
         bazel_version: [6.5.0, 7.1.0]
+        sandbox: [true, false]
```
luispadron, line 89: Can you add this info to the `name` of the step as well so its possible to tell the difference in github

gyfelton: right

gyfelton: updated, probably need to update the required checks as well so this can be landed

luispadron: i can do that now

PR #912, opened 2024-09-06
Ensures the determinism of vfsoverlay file

congt:
hdrs, private_hdrs and swiftmodules of the same framework may be in different order when building with different build settings in Bazel 7. This change will ensure that the vfsoverlay file will be always the same for the same dependency graph.

### DIFF
--- a/rules/framework/vfs_overlay.bzl
+++ b/rules/framework/vfs_overlay.bzl
@@ -9,10 +9,10 @@ in-memory tree roots, it pre-pends the prefix of the `vfsoverlay` path
 to each of the entries.
 """
 
-load("//rules:providers.bzl", "FrameworkInfo")
-load("//rules:features.bzl", "feature_names")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_common")
+load("//rules:features.bzl", "feature_names")
+load("//rules:providers.bzl", "FrameworkInfo")
 
 FRAMEWORK_SEARCH_PATH = "/build_bazel_rules_ios/frameworks"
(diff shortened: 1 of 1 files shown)


gyfelton (approved):

rules/framework/vfs_overlay.bzl
```diff
@@ -376,6 +376,12 @@ def make_vfsoverlay(ctx, hdrs, module_map, private_hdrs, has_swift, swiftmodules
     vfs_parent_len = len(vfs_parent.split("/")) - 1
     vfs_prefix = _make_relative_prefix(vfs_parent_len)
 
+    # Ensures the determinism of vfsoverlay file
+    # hdrs, private_hdrs and swiftmodules of the same framework may be in different order.
+    hdrs = sorted(hdrs)
```
thiagohmcruz, line 381: I'd suggest to verify that these sort steps don't introduce a significant performance regression in the analysis phase.

thiagohmcruz (approved):
LGTM, just one comment on perf.

rules/framework/vfs_overlay.bzl
congt: The overhead should be fine. [apple_library](https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl#L527) has already sorted all srcs files. The number of hdrs, private_hdrs and swiftmodules should be smaller than that, especially for swift modules.

gyfelton: I wonder what kind of CI test can red green on this... sandbox mode only ensure all inputs that should be present are present, but nothing about ensuring no change on file content when a rebuild happend

gyfelton: maybe start with adding before after in the PR desc? I am sure we can figure some CI test out in future based on the repro

PR #917, opened 2024-09-17
Add swift.experimental_rules_swift_package_manager flag

thiagohmcruz:


rules/framework.bzl
```diff
@@ -267,7 +267,27 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
 
     # We need to map all the deps here - for both swift headers and others
     fw_dep_vfsoverlays = []
+
+    # Wheter or not to collect SwiftInfo providers from rules_swift_package_manager targets
```
jschear, line 271:
```suggestion
    # Whether or not to collect SwiftInfo providers from rules_swift_package_manager targets
```

rules/framework.bzl
```diff
@@ -267,7 +267,27 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
...
+    enable_rules_swift_package_manager = feature_names.experimental_rules_swift_package_manager in ctx.features
+
     for dep in transitive_deps + deps:
+        if enable_rules_swift_package_manager and SwiftInfo in dep:
+            if dep.label.workspace_name.startswith("swiftpkg"):
```
congt, line 276: Could we make it more generic so that it also works when a apple_framework_packaging depends on a swfit_library?

thiagohmcruz: I'm adding this as an experimental flag with the intent of iterating on this one step at a time so we don't unintentionally break other things. I'd rather land a change that addresses the specific use case we're dealing with at the moment and progressively expand support. It's a good suggestion but I'll handle this in follow up changes if that's ok.

gyfelton:
actually one thing i realized is this branch does not contain fixes for latest rules, so any test that fails with `latest rules true` can be ignored

congt (approved):

PR #918, opened 2024-09-17
Fix new PR cancel previous PR's workflow

gyfelton:
see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,6 @@ on:
       - master
 
 # One active job per PR, cancel older ones on push
-# Just add here to verify above is true
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


mattrobmattrob:
Have the specific example where you saw this not working?

thiagohmcruz (approved):

gyfelton:
@mattrobmattrob While i am waiting for my PR to pass, it got cancelled here: https://github.com/bazel-ios/rules_ios/actions/runs/10905270306/attempts/2 
I checked actions list and saw another PR coming in, so I start looking at the github doc and it seems that the new change is what is recommended. 
I also test myself by add a new commit on top of this PR. And previous run is cancelled: https://github.com/bazel-ios/rules_ios/pull/918/checks?sha=9c421cdfc3b13537a707a7a9524d21235dff2aa4

i assume can only test whether another PR job will kill previous run or not after merging this change

gyfelton:
Note that runners we have are super limited, so new PR just cannot have CI job running until another PR finishes. We are considering to remove support for bazel 6.5.0 in test matrix to help with this. But not sure when to do it

PR #916, opened 2024-09-16
Merge clang/swift modules into one `direct_module` in SwiftInfo for a framework

gyfelton:
### What changed:
Rework of `_get_merged_swift_info` by bubble up the creation of clang module to `_apple_framework_packaging_impl` level, where it determine how to populate the clang module based on VFS feature on/off.
Remove the need to pass `virtualize_frameworks` flag to the private method `_get_merged_swift_info`

### Why this change
add `fail(swift_info.direct_modules)` after creation of it at the end of the packging impl method when running `bazel build //tests/ios/frameworks/testonly/... --config=ios --config=vfs`, you get 
```
[
struct(
clang = struct(compilation_context = <unknown object com.google.devtools.build.lib.rules.cpp.CcCompilationContext>, module_map = None, precompiled_module = None), compilation_context = None, is_system = False, name = "SwiftLibrary", 
swift = None),

 struct(
 clang = None, compilation_context = None, is_system = False, name = "SwiftLibrary",
 swift = struct(ast_files = (), defines = (), indexstore = None, plugins = [], private_swiftinterface = None, swiftdoc = <generated file tests/ios/frameworks/testonly/SwiftLibrary.swiftdoc>, swiftinterface = None, swiftmodule = <generated file tests/ios/frameworks/testonly/SwiftLibrary.swiftmodule>, swiftsourceinfo = None, symbol_graph = None))
]
```
with the current impl in `master` branch. Note that for the same module `SwiftLibrary`, there are two structs representing it, one has clang and no swift, and one has swift and nothing inside `clang`.

So this PR effectively merge two, but the clang module generation differ based on VFS turn on or not:
with VFS on:
```
[struct(
    clang = struct(compilation_context = <unknown object com.google.devtools.build.lib.rules.cpp.CcCompilationContext>, module_map = None, precompiled_module = None), compilation_context = None, is_system = False, name = "SwiftLibrary", swift = None), struct(clang = None, compilation_context = None, is_system = False, name = "SwiftLibrary", 
 
   swift = struct(ast_files = (), defines = (), indexstore = None, plugins = [], private_swiftinterface = None, swiftdoc = <generated file tests/ios/frameworks/testonly/SwiftLibrary.swiftdoc>, swiftinterface = None, swiftmodule = <generated file tests/ios/frameworks/testonly/SwiftLibrary.swiftmodule>, swiftsourceinfo = None, symbol_graph = None))
]
```
with VFS off:
```
[struct(

[...]

PR #919, opened 2024-09-17
Remove swift info provider creation inside hmap rule

gyfelton:
Last place that touched this code is on https://github.com/bazel-ios/rules_ios/pull/906where I thought it's important to construct SwiftInfo provider for hmap and add compilation context to pass the failing test. 
However that leads to a new [issue](https://github.com/bazel-ios/rules_ios/pull/916#pullrequestreview-2310552165) with `docc_archive` rule with latest `rules_apple/rules_swift`. We realized it's actually not needed to provide any SwiftInfo provider at all or it will "confuse" rules_swift/apple (since newer version starts to rely more on clang info from Swift info itself instead `apple_common.Objc`). Removing this also have the original failing test passing.

Tests done:
- [x] my own downstream repo still builds fine.
- [x] CI is green

Will squash before merge as it contains already merged commits

### DIFF
--- a/rules/hmap.bzl
+++ b/rules/hmap.bzl
@@ -78,22 +78,12 @@ def _make_headermap_impl(ctx):
     cc_info_provider = CcInfo(
         compilation_context = compilation_context,
     )
-    swift_info_provider = swift_common.create_swift_info(
-        modules = [swift_common.create_module(
-            name = ctx.attr.name,
-            clang = swift_common.create_clang_module(
-                compilation_context = compilation_context,
-                module_map = None,
-            ),
-        )],
-    )
     providers = [
         DefaultInfo(
             files = depset([headermap]),
         ),
         apple_common.new_objc_provider(),
         cc_info_provider,
-        swift_info_provider,
     ]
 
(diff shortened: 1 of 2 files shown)


rules/hmap.bzl
```diff
@@ -78,22 +78,12 @@ def _make_headermap_impl(ctx):
...
-                compilation_context = compilation_context,
-                module_map = None,
-            ),
-        )],
-    )
     providers = [
         DefaultInfo(
             files = depset([headermap]),
         ),
         apple_common.new_objc_provider(),
```
luispadron, line 85: We should try to remove this as well

thiagohmcruz (approved):

rules/hmap.bzl
gyfelton: my understanding is bazel 6 still need it. but i can be wrong. 
I feel that demands a separate PR to just remove all `apple_common.objc` references? Also helps when bump to bazel 7.4 / nightly as the API is just gone now: https://bazel.build/rules/lib/toplevel/apple_common

luispadron (approved):

PR #922, opened 2024-09-18
Relanding of `Add swift.experimental_rules_swift_package_manager flag` back to master branch

gyfelton:
already approved at https://github.com/bazel-ios/rules_ios/pull/917

Now it's about landing it in master and since it is guarded by a flag, it's no op unless you enable this feature

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -267,7 +267,27 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
 
     # We need to map all the deps here - for both swift headers and others
     fw_dep_vfsoverlays = []
+
+    # Whether or not to collect SwiftInfo providers from rules_swift_package_manager targets
+    enable_rules_swift_package_manager = feature_names.experimental_rules_swift_package_manager in ctx.features
+
     for dep in transitive_deps + deps:
+        if enable_rules_swift_package_manager and SwiftInfo in dep:
+            if dep.label.workspace_name.startswith("swiftpkg"):
+                for spm_dep in dep[SwiftInfo].transitive_modules.to_list():
+                    if spm_dep.swift:
+                        spm_dep_vfs = make_vfsoverlay(
+                            ctx,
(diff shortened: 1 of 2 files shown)


rules/framework.bzl
```diff
@@ -267,7 +267,27 @@ def _get_virtual_framework_info(ctx, framework_files, compilation_context_fields
...
     # We need to map all the deps here - for both swift headers and others
     fw_dep_vfsoverlays = []
+
+    # Whether or not to collect SwiftInfo providers from rules_swift_package_manager targets
+    enable_rules_swift_package_manager = feature_names.experimental_rules_swift_package_manager in ctx.features
+
     for dep in transitive_deps + deps:
+        if enable_rules_swift_package_manager and SwiftInfo in dep:
+            if dep.label.workspace_name.startswith("swiftpkg"):
+                for spm_dep in dep[SwiftInfo].transitive_modules.to_list():
```
luispadron, line 277: I dont really understand why we need to specifically check for rules_spm here, can you help me understand?

rules_spm creates `swift_library_group` targets which have `deps` attr, and all the things in `deps` should already have `SwiftInfo` so couldn't you just iterate through the deps here?

thiagohmcruz: We had other internal targets going through this conditional internally, the intent was to explicitly filter for only the SPM ones to minimize the chances of unintentionally breaking things for other consumers due to this being WIP-code under an experimental flag for now.

That said, if there's a more optimal way to collect the same info def open to suggestions.

thiagohmcruz (approved):

rules/framework.bzl
luispadron: I guess I just don't understand what this is doing and why we should do it for rules_spm stuff but not rules_swift

luispadron: Not trying to block, feel free to merge if this fixes something for y'all. Trying to understand because we also use rules_spm and haven't needed this

PR #927, opened 2024-11-18
Fix unit test attr forwarding

luispadron:
The `test.bzl` macros were incorrectly forwarding some of the attrs like `features`. This was leading to silent errors where things passed via `features` were dropped (such as main thread checking). This PR fixes the macro by properly forwarding ALL supported unit test attrs to the unit test rule.

### DIFF
--- a/rules/test.bzl
+++ b/rules/test.bzl
@@ -9,7 +9,6 @@ _IOS_TEST_KWARGS = [
     "bundle_id",
     "bundle_name",
     "env",
-    "features",
     "flaky",
     "frameworks",
     "infoplists",
@@ -34,7 +33,6 @@ _APPLE_BUNDLE_ATTRS = {
         "bundle_id",
         "bundle_name",
         "families",
-        "features",
         "frameworks",
         "infoplists",
         "linkopts",
@@ -44,6 +42,22 @@ _APPLE_BUNDLE_ATTRS = {
         "test_host",
     ]
 }
+
+# See: https://bazel.build/reference/be/common-definitions#common-attributes
+# NOTE: `testonly` arent listed because we have custom logic for them in here.
+_ALWAYS_AVAILABLE_ATTRS = [
+    "compatible_with",
+    "deprecation",
+    "distribs",
+    "exec_compatible_with",
+    "exec_properties",
+    "features",
+    "restricted_to",
+    "tags",
+    "target_compatible_with",
+    "toolchains",
+    "visibility",
+]
 
 _DEFAULT_APPLE_TEST_RUNNER = "@build_bazel_rules_apple//apple/testing/default_runner:ios_default_runner"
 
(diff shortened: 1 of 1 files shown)


rules/test.bzl
```diff
@@ -104,15 +104,12 @@ def _make_test(name, test_rule, **kwargs):
     Helper to create an individual test
     """
     runner = kwargs.pop("runner", None) or _DEFAULT_APPLE_TEST_RUNNER
-    test_attrs = {k: v for (k, v) in kwargs.items() if k not in _APPLE_BUNDLE_ATTRS}
+    test_attrs = {k: v for (k, v) in kwargs.items() if k in _IOS_TEST_KWARGS}
```
luispadron, line 107: This has been broken for a while it seems but afaik the right logic here is to grab the attrs for the unit test rule.. not what it was originally doing. Without this change `features` is dropped because it appars in `_APPLE_BUNDLE_ATTRS` (because every rule supports the `features` attr).

jszumski: My guess is this might have been to support custom rules with attributes unknown to this macro?

luispadron: Updated this maintain the original logic with the addition of including implicit rule attrs as well (like `features)

rules/test.bzl
```diff
@@ -45,6 +43,22 @@ _APPLE_BUNDLE_ATTRS = {
     ]
 }
 
+# See: https://bazel.build/reference/be/common-definitions#common-attributes
```
gyfelton, line 46: so if bazel's newer version contains new common attribute, we need to update this list right?

rules/test.bzl
```diff
@@ -104,7 +118,7 @@ def _make_test(name, test_rule, **kwargs):
     Helper to create an individual test
     """
     runner = kwargs.pop("runner", None) or _DEFAULT_APPLE_TEST_RUNNER
-    test_attrs = {k: v for (k, v) in kwargs.items() if k not in _APPLE_BUNDLE_ATTRS}
+    test_attrs = {k: v for (k, v) in kwargs.items() if (k not in _APPLE_BUNDLE_ATTRS) or (k in _ALWAYS_AVAILABLE_ATTRS)}
```
gyfelton, line 121: when downstream project point to this commit, probably worth printing out what is dropped before VS after, to see if anything other than `features` are dropped also

gyfelton (approved):

rules/test.bzl
luispadron: Ideally yeah, however most of these attrs are unused

rules/test.bzl
luispadron: Do you mean adding a `print` here?

congt (approved):

rules/test.bzl
gyfelton: yeah but do it temp. during the testing

PR #928, opened 2024-11-19
Allow library tool resource handling to be disabled

luispadron:
This allows the wrap resource in filegroups and resource bundle generators to be disabled completely. It allows users to just pass in `data` and have the rule just forward that to the underlying libraries.

This cleans up the code a bit too since we use `data` directly instead of `module_data` which had to be converted into a list in a few places.

I added a test showcasing where this ability would be required. It fails without this change and passes when disabling the library tools

### DIFF
--- a/rules/library.bzl
+++ b/rules/library.bzl
@@ -980,12 +980,18 @@ def apple_library(
         #enable_framework_vfs = enable_framework_vfs
     )
 
-    # Generate resource bundles
-    module_data = library_tools["wrap_resources_in_filegroup"](
-        name = name + "_wrapped_resources_filegroup",
-        srcs = data,
-        testonly = testonly,
-    )
+    # Wrap resources in a filegroup if requested.
+    if library_tools["wrap_resources_in_filegroup"]:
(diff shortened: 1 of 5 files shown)


jszumski (approved):

rules/library.bzl
```diff
@@ -980,19 +980,26 @@ def apple_library(
...
+        data = [
+            library_tools["wrap_resources_in_filegroup"](
+                name = name + "_wrapped_resources_filegroup",
+                srcs = data,
+                testonly = testonly,
+            ),
+        ]
+
+    # Generate resource bundles for any requested resource files.
+    if library_tools["resource_bundle_generator"]:
```
congt, line 995: Is it necessary to set resource_bundle_generator to None? It's no-op if a target doesn't have `resource_bundles`.  If we want to make it explicit, we probably should raise if `resource_bundle_generator` is None && `resource_bundles` is not empty.

luispadron: I see what you mean though about this not being used if the resource_bundles attr isn't set (which is the case I care about) so I'll back this change out and keep just the wrap resources disabling ability

PR #931, opened 2024-12-16
Add objc_library binary to outputs of apple_framework_packaging rule when virtualized frameworks feature is enabled

jschear:
This fixes an issue where building an `apple_framework_packaging` rule with the VFS feature enabled wouldn't trigger objc compilation for the framework.

When the VFS feature is disabled, the binaries from the `objc_library` and `swift_library` targets are merged using libtool, and that binary is included in the outputs of the `apple_framework_packaging` rule ([see packaging action](https://github.com/bazel-ios/rules_ios/blob/f7dcd0c4f90985495bba517dc8cc7d86ce7f7632/rules/framework/framework_packaging.py#L19)). 

When the VFS feature is enabled, all of the packaging actions are skipped, and the binaries from the `objc_library` and `swift_library` targets aren't merged. One of the input binaries is arbitrarily included in the rule's output files -- for a target containing swift and objc sources, this would be the `swift_library` binary.

The PR adds both input binaries to the rule outputs when the VFS feature is enabled, so building an `apple_framework_packaging` rule triggers both objc and swift compilation actions.

### DIFF
--- a/rules/framework.bzl
+++ b/rules/framework.bzl
@@ -1,36 +1,36 @@
 """Framework rules"""
 
-load("//rules/framework:vfs_overlay.bzl", "VFSOverlayInfo", "make_vfsoverlay")
-load("//rules:features.bzl", "feature_names")
-load("//rules:library.bzl", "PrivateHeadersInfo", "apple_library")
-load("//rules:plists.bzl", "process_infoplists")
-load("//rules:providers.bzl", "AvoidDepsInfo", "FrameworkInfo")
-load("//rules:transition_support.bzl", "transition_support")
-load("//rules:utils.bzl", "is_bazel_7")
-load("//rules/internal:objc_provider_utils.bzl", "objc_provider_utils")
 load("@bazel_skylib//lib:partial.bzl", "partial")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 load("@build_bazel_rules_apple//apple/internal:apple_product_type.bzl", "apple_product_type")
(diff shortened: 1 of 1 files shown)


luispadron (approved):
Makes sense thank you!

thiagohmcruz (approved):
LGTM! Thx for digging on this one!

rules/framework.bzl
```diff
@@ -221,11 +221,6 @@ def _framework_packaging_multi(ctx, action, inputs, outputs, manifest = None):
         return []
     if inputs == [None]:
         return []
-
-    virtualize_frameworks = feature_names.virtualize_frameworks in ctx.features
-    if virtualize_frameworks:
-        return inputs
```
gyfelton, line 227: before this change, `swiftmodule_out` equals to `[swiftmodule_in]` when VFS is enabled
with thie PR, `swiftmodule_out` equals to `swiftmodule_in` instead. Am I understanding this diff correctly?

jschear: I think `swiftmodule_out` is equal to `swiftmodule_in` both before and after this change.

Before:
```
swiftmodule_out = _framework_packaging_single(ctx, "swiftmodule", [swiftmodule_in], swiftmodule_out, framework_manifest)

def _framework_packaging_single(ctx, action, inputs, output, manifest = None):
    outputs = _framework_packaging_multi(ctx, action, inputs, [output], manifest = manifest)
    return outputs[0] if outputs else None

def _framework_packaging_multi(ctx, action, inputs, outputs, manifest = None):
   # ...
    if virtualize_frameworks:
        return inputs
   # ...
```

`_framework_packaging_multi` just returns the `inputs` list, and `_framework_packaging_single` returns the first item of that list. Does that make sense?

Issue #862, opened 2024-04-02
Consider removing `force_load_direct_deps_rule` in Bazel 7+


luispadron:
The flag `--incompatible_objc_alwayslink_by_default` should effectively do the same thing. Folks can also use `alwayslink = True` to be more precise. 

More info: https://github.com/bazelbuild/rules_apple/issues/1938

PR #934, opened 2024-12-21
Remove references to objc_provider and remove force_load_direct_deps to support rules_apple > 3.15.0

ssarad:
1) With rules_apple removing legacy objc_providers, rules_apple > 3.15.0 is no longer compatible with rules_ios due to https://github.com/bazelbuild/rules_apple/pull/2611

2) Removing force load direct deps as its no longer needed in Bazel 7 https://github.com/bazel-ios/rules_ios/issues/862

```sh
ERROR: /Users/ssarad/Bazel/BUILD.bazel:3:15: in apple_framework_packaging rule 
	File "/Users/ssarad/.bazel_cache/output/external/rules_ios~/rules/framework.bzl", line 1109, column 48, in _apple_framework_packaging_impl
		bundle_outs = _bundle_dynamic_framework(ctx, is_extension_safe = is_extension_safe, avoid_deps = avoid_deps)
	File "/Users/ssarad/.bazel_cache/output/external/rules_ios~/rules/framework.bzl", line 888, column 44, in _bundle_dynamic_framework
		partials.framework_provider_partial(
	File "/Users/ssarad/.bazel_cache/output/external/rules_apple~/apple/internal/partials/framework_provider.bzl", line 94, column 5, in framework_provider_partial
		def framework_provider_partial(
Error: framework_provider_partial() got unexpected keyword argument: objc_provider

```

This PR is attempt to make rules_ios compatible with rules_apple + remove the legacy code to force load direct deps as with Bazel 7, `--incompatible_objc_alwayslink_by_default` does the same thing.


ssarad:
@luispadron If I can get your review that'd be fantastic 🙏

.github/workflows/tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         XCODE_VERSION: ${{ matrix.xcode_version }}
         USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
         LATEST_RULES_SWIFT_VERSION: 2.1.1
-        LATEST_RULES_APPLE_VERSION: 3.8.0
+        LATEST_RULES_APPLE_VERSION: 3.15.0
```
ssarad, line 32: Added this line because CI was complaining about this (do let me know if this isnt right)

luispadron:
Thanks for getting the work started here! I think theres some left-over usage:

```
Error: <target //tests/ios/in-tree-vendor-prebuilt-deps/Pods/MiSnap:MiSnap-MiSnap.framework-import> (rule 'apple_dynamic_framework_import') doesn't contain declared provider 'objc'
```

Do we need to upgrade rules_apple in `MODULE.bazel`, if so, we'll need to drop Bazel 6 support (probably in it's own PR).

ssarad:
> Thanks for getting the work started here! I think theres some left-over usage:
> 
> ```
> Error: <target //tests/ios/in-tree-vendor-prebuilt-deps/Pods/MiSnap:MiSnap-MiSnap.framework-import> (rule 'apple_dynamic_framework_import') doesn't contain declared provider 'objc'
> ```
> 
> Do we need to upgrade rules_apple in `MODULE.bazel`, if so, we'll need to drop Bazel 6 support (probably in it's own PR).

@luispadron I think when I didnt change rules_apple to 3.15.0, I got build errors that rules_apple was expecting the objc_provider parameter.

I would vote to drop Bazel 6 support mainly because 

1) Your recent survey showed that most of us are on Bazel 7
2) rules_apple has dropped support for Bazel 6 

What do you think? If you’d think its fine to drop support for Bazel 6, I’m happy to start that PR before we let this in.

luispadron:
Yeah I think it makes sense! @thiagohmcruz @gyfelton FYI in case we have some reason for keeping Bazel 6

ssarad:
> Yeah I think it makes sense! @thiagohmcruz @gyfelton FYI in case we have some reason for keeping Bazel 6

Okay perfect. I can start that PR and will tag you along. I’ll likely complete it after New Years when I am back from vacation.

ssarad:
@luispadron @jszumski 

When you have some time, can I get your reviews on this PR please? All the checks should pass so its ready to review.

Thank you and much appreciated! 🙏

mccorkill1 (approved):

rules/force_load_direct_deps.bzl
```diff
@@ -35,142 +35,3 @@ def _cc_info_static_libraries(dep):
             static_cc_libraries.append(library_to_link.static_library)
 
     return static_cc_libraries
-
-# TODO: We should deprecate this rule for Bazel 7+ as `--incompatible_objc_alwayslink_by_default` effectively
```
jszumski, line 39: Should we be setting this flag in CI somewhere or using `alwayslink` on our Obj-C library targets?  I'm surprised nothing broke without it, but maybe an app would only blow up at launch time.

ssarad: Ah good catch! Looks like the flag is off by default, I assumed it is ON by default 

https://bazel.build/reference/command-line-reference#flag--incompatible_objc_alwayslink_by_default

I can set this flag in bazelrc (or somewhere else if you prefer)

ssarad: added in bazelrc! @jszumski

thiagohmcruz:
I don't think this was being exercised in OSS properly, I'm running CI internally pointing to this PR since we rely on the features being touched here. My first test has failed with a bunch of `ld: symbol(s) not found for architecture arm64`. Digging more to see if it's related to this PR or not and why.

rules/import_middleman.bzl
```diff
@@ -95,15 +95,6 @@ def _find_imports_impl(target, ctx):
     if ctx.rule.kind == "objc_import":
         imported_library_file.append(target[apple_common.Objc].imported_library)
```
luispadron, line 96: Theres a bit more usage of `apple_common.Objc` in this file

rules/import_middleman.bzl
```diff
@@ -95,15 +95,6 @@ def _find_imports_impl(target, ctx):
...
 
-    elif AppleFrameworkImportInfo in target:
-        static_framework_file.append(target[apple_common.Objc].imported_library)
```
thiagohmcruz, line 99: @ssarad confirmed testing internally that this PR as is breaks things for us due to linker errors, instead of removing it can't we propagate the same artifacts collecting it from `CcInfo`? E.g. (code is not final, just to illustrate my point):
```
diff --git a/rules/import_middleman.bzl b/rules/import_middleman.bzl
index bdd2250..fde199b 100644
--- a/rules/import_middleman.bzl
+++ b/rules/import_middleman.bzl
@@ -94,8 +94,12 @@ def _find_imports_impl(target, ctx):
 
     if ctx.rule.kind == "objc_import":
         imported_library_file.append(target[apple_common.Objc].imported_library)
-
     elif AppleFrameworkImportInfo in target:
+        if CcInfo in target:
+            for linker_input in target[CcInfo].linking_context.linker_inputs.to_list():
+                for library_to_link in linker_input.libraries:
+                    if library_to_link.static_library:
+                        static_framework_file.append(depset([library_to_link.static_library]))
         target_dynamic_framework_file = target[apple_common.Objc].dynamic_framework_file
         target_dynamic_framework_file_list = target_dynamic_framework_file.to_list()
         if len(target_dynamic_framework_file_list) > 0:
 ```
 
 cc @luispadron thoughts?

Comment (ssarad) [reply rules/import_middleman.bzl:99]: @thiagohmcruz Thank you very much for debugging this! I'm fine with propagating this way. Frankly speaking I wasn't too familiar with this file so I removed it but happy to follow what you think is best.

Comment (ssarad) [reply rules/import_middleman.bzl:99]: @thiagohmcruz Okay so I got your piece of code added into my PR, i tried building everything locally and they are fine for me but could you give this a shot again? 

Let me know if there is a way I can also test this locally, I can totally understand this being painful for you and happy to test things on my side before I push it.

File: rules/import_middleman.bzl (hunk 3 above)
Comment (ssarad) [reply rules/import_middleman.bzl]: Done! Thanks for catching this

File: rules/import_middleman.bzl (hunk 4 above)
Comment (thiagohmcruz) [reply rules/import_middleman.bzl:99]: Thx! Running one more test later today and will share results

File: rules/library.bzl [hunk 5]
```diff
@@ -1069,6 +1069,7 @@ def apple_library(
             deps = deps + private_deps + private_dep_names,
             defines = defines,
             tags = tags_manual,
+            alwayslink = True,
```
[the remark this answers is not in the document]
ssarad: done! thanks @jszumski

jszumski (approved):

rules/import_middleman.bzl
thiagohmcruz: For future reference, our CI has failed pointing to the latest version of this PR but not due to the ObjC provider so it's likely due to us being not up to date on `rules_apple` and `rules_swift`, meaning it's unrelated to your changes. Thx for making the adjustments here!

cc @jschear this likely blocks us from bumping `rules_ios` until we're up to date on the rules. Let me know if there's anything you'd rather solve in this PR before landing

thiagohmcruz (approved):

rules/force_load_direct_deps.bzl
```diff
@@ -35,142 +35,3 @@ def _cc_info_static_libraries(dep):
             static_cc_libraries.append(library_to_link.static_library)
 
     return static_cc_libraries
-
```
jschear, line 38: Can this entire file be deleted now?

ssarad: Now that I think about this, I think yes. @thiagohmcruz @jszumski Do you know if we need `_cc_info_static_libraries` and `_objc_provider_static_libraries`? I don't see its reference anywhere else outside this file

jszumski: Nope, I think the whole thing can go

ssarad:
I think we got all the stuffs covered here. Anything pending to merge this? @jschear @jszumski @thiagohmcruz 

Thank you all again for all the reviews. I appreciate it! 🙏

PR #936, opened 2025-01-13
Remove use of struct.to_json()

nataliejameson:
Bazel 8 does not appear to support struct.to_json() anymore, so remove its usage in vfs_overlay. It appears this API has been around since at least 6.5 (https://bazel.build/versions/6.5.0/rules/lib/core/json), so changing to it shouldn't be an issue.


thiagohmcruz (approved):
Thx for the PR! No major concerns here, not sure how the performance of the two diff APIs compare but not super worried about landing this since worst case it's simple to revert (regardless using the latest API is preferred indeed).

PR #945, opened 2025-06-13
Update for rules_swift 3.0

luispadron:


rules/framework.bzl
```diff
@@ -695,6 +696,7 @@ def _bundle_dynamic_framework(ctx, is_extension_safe, avoid_deps):
     actions = ctx.actions
     apple_mac_toolchain_info = ctx.attr._mac_toolchain[AppleMacToolsToolchainInfo]
     apple_xplat_toolchain_info = ctx.attr._xplat_toolchain[AppleXPlatToolsToolchainInfo]
+    cc_toolchain_forwarder = ctx.split_attr._child_configuration_dummy
```
jschear, line 699: Huh, what's this?

rules/library.bzl
```diff
@@ -1106,15 +1106,23 @@ def apple_library(
     if module_map:
         objc_hdrs.append(module_map)
 
+    swift_interop_hint_name = name + "_swift_interop_hint"
+    swift_interop_hint(
```
jschear, line 1110: AIUI using `swift_interop_hint` is preferred over the `objc_library` `module_map` attr, so this seems good -- but is this necessary for rules_swift 3.x?

jschear (approved):

rules/library.bzl
luispadron: `module_map` is now "deprecated" in `objc_library` so it fails when using rules_swift 3.x: https://github.com/bazelbuild/rules_swift/blob/master/swift/swift_clang_module_aspect.bzl#L220-L225

